PR #8, opened 2018-03-22
1.0.2 release

danac-gs:
Discounts improvements. 
Order creation triggered on live page event - grand total change, when discounts are added / removed on checkout page.


daisy1754:
Sorry for delay here. For some reason I missed this PR.

PR #10, opened 2018-04-03
Update hints data on front end input fields change / Bugsnag: Include TraceID & Track Bolt API error http response / Cart totals mismatch prevent / Prefetch and cache Shipping and Tax 

danac-gs:


Model/Api/ShippingMethods.php
```diff
@@ -251,6 +252,8 @@ private function getShippingOptions($quote)
 
 		foreach ($output as $shippingMethod) {
 
+			if ($shippingMethod->getErrorMessage()) continue;
```
daisy1754: let's bugsnag

view/frontend/templates/js/replacejs.phtml
```diff
@@ -2,18 +2,11 @@
...
+ *
  * Replace js template
  * Puts / replaces the checkout buttons according to defined selectors.
  * Maintains Bolt checkout functionality.
- *
- * @var $block \Bolt\Boltpay\Block\Js
  */
-
 if (!$this->isActive()) return;
 ?>
```
daisy1754: Hmm, looks like there is no corresponding open `<?php`
Is it right?

view/frontend/templates/js/replacejs.phtml
```diff
@@ -377,55 +387,51 @@ window.boltConfig = <?=$this->getSettings(); ?>;
...
+            var prefill = isObject(data.hints.prefill) ? deepMeggeObjects(hints.prefill, data.hints.prefill) : hints.prefill;
+            hints = deepMeggeObjects(hints, data.hints);
+            hints.prefill = prefill;
+
             hints.publishableKey = getCheckoutKey();
-            BoltConnect.process(cart, hints, callbacks);
-        };
 
-        var onError = function () {
+            BoltConnect.process(cart, hints, callbacks);
```
daisy1754: Please use `BoltCheckout.configure` instead of `BoltConnect.process`. These two are basically same but we recently renamed.

view/frontend/templates/js/replacejs.phtml
danac-gs: Yes there is, several lines ahead, on line1.
<?php
/**
 * Copyright © 2013-2017 Bolt, Inc. All rights reserved.
 * See COPYING.txt for license details.
 *
 * Replace js template
 * Puts / replaces the checkout buttons according to defined selectors.
 * Maintains Bolt checkout functionality.
 */
if (!$this->isActive()) return;
?>

Helper/Cart.php
```diff
@@ -365,21 +380,37 @@ public function getCartData($payment_only, $place_order_payload)
...
+					'service'          => $shippingAddress->getShippingDescription(),
+				];
+			}
+
+			$tax_amount         = $shippingAddress->getTaxAmount();
+			$rounded_tax_amount = $this->getRoundAmount($tax_amount);
 
-			$totalAmount = $quote->getGrandTotal();
-			$taxAmount   = $this->getRoundAmount($shippingAddress->getTaxAmount());
+			$diff = $tax_amount * 100 - $rounded_tax_amount;
```
daisy1754: can you bugsnag if diff is huge? Also maybe we should error if diff is larger than certain amount (say more than 10 cents).

danac-gs: When 2 decimal places are used for price, as it is the case in Magento2, $diff is always 0. This code covers the case that is not likely to happen. Suppose for some reason the 2 decimal price is changed to 4 decimal places. Not straightforward / configurable but pragmatically possible.

For example we have a product with the price, say, 10.0049. and add 30 of them to the cart.
Every single item is rounded to 10 (1000 in Bolt units, cents, whatever) but the real total is 300.147, or 30015 rounded to bolt total. It is 15 cent difference but that is no error. We must add it to the total the way both sides are satisfied. 

While doing this I added random values for additional 2 decimals and added tens of different products to cart to make sure the code works in that case and it did.

One important thing that I found out by testing and I am exploiting here is how the order is verified at Bolt side. The item totals sum (price * qty) is checked against cart total. The item unit price is not considered, moreover you can put any price in that field. I tested by subtracting the price by 1 to 10 and it passed. That implies the total for the item can be changed / increased as well as the cart total as long as the sum of item totals equals the cart total. So if there is a $diff I add it to the cart total and also to the first item total. Item totals are populated with rounded values previously.  

On order creation there are also discounts, shipping and tax taken into account. 

Similarly for shipping and tax call. If there is any truncated value, it is added to the shipping tax.

But, as I said, this covers very unlikely case.

danac-gs: bugsnaged

danac-gs:
@daisy1754 can you merge this? We may increase the plugin version afterwards, but maybe we don't need to do it yet, until we have at least one integration complete.

PR #13, opened 2018-04-20
Shipping and Tax prefetch improvements / hints reconfigure optimization

danac-gs:


Model/Api/ShippingMethods.php
```diff
@@ -251,17 +251,32 @@ public function getShippingMethods($cart, $shipping_address)
...
-    public function shippingEstimation($quote, $cart, $shipping_address)
+    public function shippingEstimation($quote, $shipping_address)
     {
-
         ////////////////////////////////////////////////////////////////////////////////////////
         // Check cache storage for estimate. If the quote_id, total_amount, country_code,
         // region and postal_code match then use the cached version.
         ////////////////////////////////////////////////////////////////////////////////////////
-        $cache_identifier = $cart['order_reference'].'_'.$cart['total_amount'].'_'.$shipping_address['country_code'].
+        $cache_identifier = $quote->getId().'_'.round($quote->getSubtotal()*100).'_'.$shipping_address['country_code'].
```
daisy1754: It's ok for now, but I think we need to expand this a bit. I think we should probably include SKUs.
For instance, cart can have same amount but different item weight, that can affect shipping fee.

PR #28, opened 2018-05-15
(Admin changes) Order create page: added an ability to pay in the order create page.

YevhenShBolt:


Helper/Order.php
```diff
@@ -421,9 +420,11 @@ private function createOrder($quote, $transaction, $frontend)
         // Emulate frontend area in order for email
         // template to be loaded from the correct path
         // even if run from the hook.
-        $this->appState->emulateAreaCode('frontend', function () use ($order){
-            $this->emailSender->send($order);
-        });
+        if ($frontend) {
```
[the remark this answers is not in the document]
YevhenShBolt: For this part of the code, it produces the error regarding admin area. I am not yet refactoring this area. I plan to do it soon.

Block/Form.php
```diff
@@ -0,0 +1,186 @@
...
+
+use Bolt\Boltpay\Helper\Config;
+use Magento\Backend\Model\Session\Quote as BackendQuote;
+use Magento\Payment\Block\Form as PaymentForm;
+use Magento\Framework\View\Element\Template\Context;
+
+/**
+ * Class Form
+ */
+class Form extends PaymentForm
```
daisy1754: Is this used by both admin and non-admin? If so, please add comment. If it's admin only, maybe rename to AdminForm?

Block/Form.php
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    /**
+     * Get checkout key. Any of the defined publishable keys for use with track.js.
+     *
+     * @return  string
+     */
+    public function getCheckoutKey()
+    {
+        return $this->configHelper->getPublishableKeyPayment();
```
daisy1754: We need to add a new config named `publishable_key_back_office` in `system.xml`. 
We need separate publishable key for back office payment processing because fee is different.

Block/Form.php
```diff
@@ -0,0 +1,186 @@
...
+     * @return string
+     */
+    public function getSettings()
+    {
+        return json_encode([
+            'connect_url'              => $this->getConnectJsUrl(),
+            'publishable_key_payment'  => $this->getCheckoutKey(),
+            'create_order_url'         => $this->getUrl(Config::CREATE_ORDER_ACTION),
+            'save_order_url'           => $this->getUrl(Config::SAVE_ORDER_ACTION),
+            'selectors'                => $this->getReplaceSelectors(),
```
daisy1754: I doubt if we need this. This is to replace checkout button in cart page.

Controller/Adminhtml/Cart/Data.php
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Get cart data for bolt pay ajax
+     *
+     * @return Json
+     * @throws Exception
+     */
+    public function execute()
+    {
+        try {
+            // additional data collected from the (one page checkout) page,
```
daisy1754: remove stale comment

Controller/Adminhtml/Cart/Data.php
```diff
@@ -0,0 +1,114 @@
...
+            // i.e. billing address to be saved with the order
+            $place_order_payload = $this->getRequest()->getParam('place_order_payload');
+            // call the Bolt API
+            $boltpayOrder = $this->cartHelper->getBoltpayOrder(true, $place_order_payload);
+
+            // format and send the response
+            $cart = [
+                'orderToken'  => $boltpayOrder ? $boltpayOrder->getResponse()->token : '',
+                'authcapture' => $this->configHelper->getAutomaticCaptureMode()
```
daisy1754: can you hard code true here? In case of backend processing I think it'll be auto-captured.

Helper/Order.php
```diff
@@ -430,7 +429,9 @@ private function createOrder($quote, $transaction, $frontend)
         // Emulate frontend area in order for email
         // template to be loaded from the correct path
         // even if run from the hook.
-        $this->appState->emulateAreaCode('frontend', function () use ($order){
+        $areaCode = $frontend ? 'frontend' : \Magento\Backend\App\Area\FrontNameResolver::AREA_CODE;
```
daisy1754: I think this contradicts with comment above:

```
// Send order confirmation email to customer.
// Emulate frontend area in order for email
// template to be loaded from the correct path
// even if run from the hook.
```

view/adminhtml/templates/boltpay/js/replacejs.phtml
```diff
@@ -0,0 +1,662 @@
+<?php
```
daisy1754: Do we need this file at all?

Helper/Order.php
YevhenShBolt: Yes, I will delete this comment.

view/adminhtml/templates/boltpay/js/replacejs.phtml
YevhenShBolt: I plan to change this file to the component for admin panel.

Block/Form.php
daisy1754: I think we can refactor later; not blocking.

Block/Form.php
```diff
@@ -0,0 +1,175 @@
...
+            'region'            => $billingAddress->getRegion(),
+            'customerId'        => $billingAddress->getCustomerId(),
+            'street'            => $streetAddressLines,
+            'telephone'         => $billingAddress->getTelephone(),
+            'postcode'          => $billingAddress->getPostcode(),
+            'city'              => $billingAddress->getCity(),
+            'firstname'     => $billingAddress->getFirstname(),
+            'lastname'      => $billingAddress->getLastname(),
+            'extensionAttributes'   => ['checkoutFields' => []],
+            'saveInAddressBook'     => null
```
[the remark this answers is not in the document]
YevhenShBolt: Nope, this method just using for collect data for Bolt.

Controller/Adminhtml/Cart/Data.php
```diff
@@ -0,0 +1,112 @@
...
+    public function execute()
+    {
+        try {
+            $place_order_payload = $this->getRequest()->getParam('place_order_payload');
+            // call the Bolt API
+            $boltpayOrder = $this->cartHelper->getBoltpayOrder(true, $place_order_payload);
+
+            // format and send the response
+            $cart = [
+                'orderToken'  => $boltpayOrder ? $boltpayOrder->getResponse()->token : '',
```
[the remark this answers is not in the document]
YevhenShBolt: if we have $boltOrder - false - we have an alert popup in js (in check function of BoltCheckout)

```javascript
    if (!cart.orderToken) {
        errors.push('Invalid Order Token');
    }

    ...

    if (errors.length) {
          let errorMsg = errors.join(', ');

         console.error('BoltPay Check: ', errorMsg);
         alert(errorMsg);

         return false;
    }
```

Block/Form.php
YevhenShBolt: As i see this block use for both part admin and frontend.
The calling of this block is defined in payment model and this model calls in both part.

>app/code/Bolt/Boltpay/Model/Payment.php

```php

    /**
     * @var string
     */
    protected $_formBlockType = \Bolt\Boltpay\Block\Form::class;

```

YevhenShBolt:
About Block/Form.php

If you look at `vendor/magento/module-payment/Block/Form/Container.php`
```php
/**
         * Prepare children blocks
         *
         * @return $this
         */
        protected function _prepareLayout()
        {
            /**
             * Create child blocks for payment methods forms
             */
            foreach ($this->getMethods() as $method) {
                $this->setChild(
                    'payment.method.' . $method->getCode(),
                    $this->_paymentHelper->getMethodFormBlock($method, $this->_layout)
                );
            }

            return parent::_prepareLayout();
        }
```

and at  `$this->_paymentHelper->getMethodFormBlock()`

```php
    /**
     * Retrieve payment method form html
     *
     * @param MethodInterface $method
     * @param LayoutInterface $layout
     * @return Form
     */
    public function getMethodFormBlock(MethodInterface $method, LayoutInterface $layout)
    {
        $block = $layout->createBlock($method->getFormBlockType(), $method->getCode());
        $block->setMethod($method);
        return $block;
    }
```

you can see that a Form block create automatically from model variable $_formBlockType.

PR #29, opened 2018-05-15
latest development fixes and improvements

danac-gs:
- create order controller: remove redundant class
- order save controller: fix result format and error code
- bugsnag: remove unhandled exceptions reporting
- cart helper: refactor hints prefill, cart creation from API call, fix product image url
- hook helper: change http error type
- hooks: improve error responses
- shipping and tax hook: fix checkCartItems SKU bug, save quote address, clear selected shiping method
- replace.js: do not prefetch shipping with payment only checkout

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -8,8 +8,7 @@ namespace Bolt\Boltpay\Helper;
 use Bolt\Boltpay\Model\Response;
 use Exception;
 use Magento\Checkout\Model\Session as CheckoutSession;
-use Magento\Catalog\Helper\Image as ImageHelper;
-use Magento\Catalog\Model\Product as ProductModel;
+use Magento\Catalog\Model\ProductFactory;
 use Bolt\Boltpay\Helper\Api as ApiHelper;
 use Bolt\Boltpay\Helper\Config as ConfigHelper;
 use Magento\Customer\Model\Session as CustomerSession;
@@ -20,6 +19,9 @@ use Magento\Quote\Model\Quote;
 use Zend_Http_Client_Exception;
 use Bolt\Boltpay\Helper\Log as LogHelper;
 use Magento\Framework\DataObjectFactory;
+use Magento\Framework\View\Element\BlockFactory;
+use Magento\Store\Model\App\Emulation;
+use Magento\Customer\Model\Address;
 
 /**
  * Boltpay Cart helper
@@ -36,19 +38,14 @@ class Cart extends AbstractHelper
     private $customerSession;
 
     /**
-     * @var ImageHelper
-     */
-    private $imageHelper;
-
-    /**
      * @var ApiHelper
      */
     private $apiHelper;
 
     /**
-     * @var ProductModel
+     * @var ProductFactory
      */
-    private $productModel;
+    private $productFactory;
 
     /**
      * @var ConfigHelper
@@ -70,6 +67,16 @@ class Cart extends AbstractHelper
      */
     private $dataObjectFactory;
 
+    /**
+     * @var BlockFactory
+     */
+    private  $blockFactory;
+
+    /**
+     * @var Emulation
+     */
+    private  $appEmulation;
+
     // Billing / shipping address fields that are required when the address data is sent to Bolt.
     private $required_address_fields = [
         'first_name',
@@ -79,6 +86,9 @@ class Cart extends AbstractHelper
         'region',
         'postal_code',
         'country_code',
+    ];
+
+    private $required_billing_address_fields  = [
         'email',
     ];
 
@@ -87,7 +97,6 @@ class Cart extends AbstractHelper
     // Can appear as keys in Quote::getTotals result array.
     ///////////////////////////////////////////////////////
     private $discount_types = [
-//        'reward',
         'giftvoucheraftertax',
     ];
     ///////////////////////////////////////////////////////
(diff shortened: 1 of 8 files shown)


Helper/Cart.php
```diff
@@ -203,16 +215,26 @@ private function getSignResponse($signRequest)
...
 
         if ($place_order_payload) {
             $place_order_payload = @json_decode($place_order_payload);
             $email = @$place_order_payload->email;
         }
 
-        $shippingStreetAddress = $shippingAddress->getStreet();
-        $hints = [
-            'prefill' => [
+        $hints = ['prefill' => []];
```
daisy1754: can you make sure when this is converted to JavaScript, empty prefill is modeled as empty object "{}" not empty array "[]"? We recently introduced some typechecking and passing empty array cause noisy warning in our monitoring.

Helper/Cart.php
```diff
@@ -321,27 +368,41 @@ public function getCartData($payment_only, $place_order_payload)
...
+            $product['sku']          = trim($item->getSku());
+
+            ////////////////////////////////////
+            // Get product description and image
+            ////////////////////////////////////
+            /**
+             * @var \Magento\Catalog\Model\Product
+             */
+            $_product = $this->productFactory->create()->load($productId);
+            // ignore description, reduce the payload until its use is confirmed
```
daisy1754: It is used in Bolt backend. Please include.

Helper/Cart.php
```diff
@@ -321,27 +368,41 @@ public function getCartData($payment_only, $place_order_payload)
...
+            // $product['description'] = strip_tags($_product->getDescription());
+            $productImage = $imageBlock->getImage($_product, 'product_small_image');
```
daisy1754: wondering if it's always exist. Should we fallback to different image when small image does not exist?

Helper/Cart.php
```diff
@@ -399,7 +460,8 @@ public function getCartData($payment_only, $place_order_payload)
                 'country_code'    => $shippingAddress->getCountryId(),
             ];
 
-            if (@$email) {
+            $email = $shippingAddress->getEmail() ?: $email;
+            if ($email) {
```
daisy1754: In case of payment only, current code overrides shipping email with billing email even if user provides different emails in magento checkout page.

Model/Api/OrderManagement.php
```diff
@@ -131,7 +131,10 @@ public function manage(
                 });
             }
 
-            $this->response->setHeader('User-Agent', 'BoltPay/Magento-'.$this->configHelper->getStoreVersion());
+            $this->response->setHeader(
+                'User-Agent',
+                'BoltPay/Magento-' . $this->configHelper->getStoreVersion()
+            );
             $this->response->setHeader('X-Bolt-Plugin-Version', $this->configHelper->getModuleVersion());
```
daisy1754: nit: let's include this in above line

view/frontend/templates/js/replacejs.phtml
```diff
@@ -157,7 +157,7 @@ window.boltConfig = <?php echo $block->getSettings(); ?>;
     // Merge objects
     ////////////////////////////////
     var isObject = function (item) {
-        return (item && typeof item === 'object' && !Array.isArray(item));
+        return (item && typeof item == 'object' && !Array.isArray(item));
```
daisy1754: I think we should use "===" by default

view/frontend/templates/js/replacejs.phtml
```diff
@@ -436,8 +434,8 @@ window.boltConfig = <?php echo $block->getSettings(); ?>;
 
             if (window.BoltCheckout) BoltCheckout.configure(cart, hints, callbacks);
 
-            // prefetch Shipping and Tax
-            prefetchShipping();
+            // prefetch Shipping and Tax for multi-step checkout
```
daisy1754: +1

Helper/Cart.php
danac-gs: It is taken care of in replace.js:

var onSuccess = function(data){
            cart = data.cart;

            var prefill = isObject(data.hints.prefill) ? deepMergeObjects(hints.prefill, data.hints.prefill) : hints.prefill;
            hints = deepMergeObjects(hints, data.hints);
            hints.prefill = prefill;

            hints.publishableKey = getCheckoutKey();

            if (window.BoltCheckout) BoltCheckout.configure(cart, hints, callbacks);

            // prefetch Shipping and Tax for multi-step checkout
            if (getCheckoutType() == 'checkout') prefetchShipping();};

Helper/Cart.php
danac-gs: done

Helper/Cart.php
danac-gs: It exists. Part of the core. Added a fallback anyway.

Helper/Cart.php
danac-gs: There is only one place to enter email in magento checkout. It is at some point added to shipping and billing addresses. If the shipping email exists that one will be used, otherwise the email that user entered is collected from the page and sent with the request ($place_order_payload) and that one will be used. If that can not be found and there is an email stored with billing, that one is used. But in practice it is all one email.

Model/Api/OrderManagement.php
danac-gs: setHeader sets one header at a time

danac-gs:
Requested changes made

PR #30, opened 2018-05-16
[WIP] prototype API to support discount code

danac-gs:


daisy1754:
Merging this for easier testing (but will not document until we actually verify this works :))

PR #32, opened 2018-05-17
associate email to current quote controller

danac-gs:


Controller/Cart/Email.php
```diff
@@ -0,0 +1,80 @@
...
+
+use Magento\Framework\App\Action\Action;
+use Magento\Framework\App\Action\Context;
+use Magento\Checkout\Model\Session as CheckoutSession;
+use Magento\Quote\Model\Quote;
+use Bolt\Boltpay\Helper\Bugsnag;
+use Magento\Framework\Exception\LocalizedException;
+
+/**
+ * Class Prefetch.
```
daisy1754: update comment

Controller/Cart/Email.php
```diff
@@ -0,0 +1,80 @@
...
+                throw new LocalizedException(__('Quote does not exist.'));
+            }
+
+            $email = $this->getRequest()->getParam('email');
+
+            if (!$email) {
+                throw new LocalizedException(__('No email received.'));
+            }
+
+            $quote->setCustomerEmail($email)->save();
```
daisy1754: maybe check if user is logged in? Not sure if we want to override user email in magento

Controller/Cart/Email.php
danac-gs: done

Controller/Cart/Email.php
danac-gs: taken care of

PR #33, opened 2018-05-22
save payment credit card / minimize order creation requests

danac-gs:


view/frontend/templates/js/replacejs.phtml
```diff
@@ -451,6 +451,25 @@ window.boltConfig = <?php echo $block->getSettings(); ?>;
...
+    var create_request_no = 0;
+    var createOrderDelayed = function() {
+        create_request_no++;
+        ! function(request_no) {
+            setTimeout(
+                function() {
+                    if (create_request_no > request_no) return;
+                    createOrder();
+                },
+                500
```
daisy1754: I'm not sure if we should do this. This would slow down loading for all users by 500ms. 

Do we know why we were making multiple requests now?

danac-gs: There are many cases when the cart is updated via ajax. Bolt order is kept in sync with the changes by monitoring specific elements on the page and triggering order creation on their change. The elements are order total, minicart quantity, shipping, etc. The order creation is called asynchronously on every change. By the time the user clicks on bolt checkout it is already updated. Often multiple events occur simultaneously or in a short interval resulting in multiple calls to bolt create order. In those cases I used to just abort previous call on every subsequent call because I am interested only in the latest. RestSupp guys noticed aborted calls were shown in red in dev console as canceled. Moreover those calls although discarded from frontend do reach Bolt and create redundant orders. With this timeout solution the number of requests is reduced and I did not notice it reflects on the usability. The delay is in async call start in the background.

danac-gs: It is on dev.restaurantsupply.com and bolt-magento2.guaranteed.site, it can be tested / chacked. I will remove it or shorten the interval if you say so.

daisy1754: Let's update it to 100ms

danac-gs: done here. do you think I should do it on restaurantsupply branch too? they do many live ajax updates that are pretty slow in response.

PR #36, opened 2018-06-05
Clone quote on bolt order

danac-gs:


Controller/Cart/Email.php
```diff
@@ -74,6 +88,11 @@ public function execute()
                 throw new LocalizedException(__('No email received.'));
             }
 
+            if (!$this->cartHelper->validateEmail($email)) {
+                $this->bugsnag->notifyError('Invalid email address', "quote_id: {$quote->getId()} email: $email");
```
daisy1754: what is the purpose of this logging?

Controller/Cart/Email.php
```diff
@@ -74,6 +88,11 @@ public function execute()
...
+            }
+
+            //$quote->setCheckoutMethod($quote->getCheckoutMethod() ?: 'guest');
```
daisy1754: nit: remove commented out code

Controller/Cart/Email.php
daisy1754: note bolt currently allows you to enter invalid email

Helper/Cart.php
```diff
@@ -315,8 +334,39 @@ public function getCartData($payment_only, $place_order_payload, $quote = null)
...
+        $quote->getReservedOrderId() ?: $this->setReservedOrderId($quote);
+
+        ////////////////////////////////////////////////////////
+        // CLONE THE QUOTE and quote billing / shipping  address
+        ////////////////////////////////////////////////////////
+        $clone = $this->quoteFactory->create();
+
+        $clone->merge($quote);
+
+        foreach ($quote->getData() as $key => $value) {
```
daisy1754: for magento1 we explicitly set a few extra fields. can you double check if these are not needed?

https://github.com/BoltApp/bolt-magento1/blob/094072a4037c28dac49070fc27a48cea0f7036ec/app/code/community/Bolt/Boltpay/Block/Checkout/Boltpay.php#L257

Controller/Cart/Email.php
danac-gs: You can save invalid mail to Magento quote entity as well (It is assigned to the quote itself, not quote shipping / billing address). The validation is done later, on order creation, towards email assigned to the address. This address email may or may not be the one we have previously explicitly assigned to the quote. It depends on the user interaction flow with the system. There is a validation step before the point where our plugin assigns the address email and the address email is left unchanged (empty) if invalid. Empty address email does not break the order creation. Moreover, the validation is not consistent on all Magento versions, which implies there can be some false negatives. The purpose of the log is to save to the quote whatever email the user had entered, and have a notification of Magento marked it invalid if we by any chance need it, like some Exception was raised at a later point. This could not be a bugsnag breadcrumb or additional metadata set because it is done asynchronously from the main flow.

Controller/Cart/Email.php
danac-gs: done

Helper/Cart.php
danac-gs: I've run the tests for anonymous and logged in user, regular and orphaned transactions, Capillus reward points too. Seems like all data is there.

The key difference between M2 and M1 implementation is that, as M2 works in a much more async fashion, not only one clone (immutable quote in M1) exists but every cart change (possibly on different tabs) triggers a new clone (bolt order) creation as a snapshot of the parent quote at the particular point in time. 

Every clone has its own shipping and billing address entities cloned (all the data copied) from the parent quote addresses. On the other hand, M1 immutable quote inherits (gets assigned) the parent address instances. I think that implies the necessity of skipping some fields and adding others explicitly to keep the totals in order. The M2 way every clone has its own totals as well as everything else.

All being said I will double check again.

danac-gs: Double checked.

Controller/Cart/Email.php
YevhenShBolt: I agree with Deniel, magento validate the email address.

danac-gs:
@daisy1754 do not merge this yet. I just need to add one more fix.

Helper/Cart.php
danac-gs: @YevhenShBolt All previously saved data that may have been added by other plugins

Controller/Order/Save.php
```diff
@@ -98,9 +98,9 @@ public function execute()
             // call order save and update
             list($quote, $order) = $this->orderHelper->saveUpdateOrder($reference);
 
-            $orderId = $order->getId();
             // clear the session data
-            if ($orderId) {
+            if ($order->getId()) {
+                $this->checkoutSession->replaceQuote($quote);
```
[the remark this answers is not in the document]
danac-gs: @YevhenShBolt done

Helper/Order.php
```diff
@@ -433,51 +445,73 @@ private function createOrder($quote, $transaction, $frontend, $bolt_trace_id = n
...
 
         $this->setPaymentMethod($quote);
         $quote->collectTotals();
 
         $this->quoteRepository->save($quote);
 
         // assign credit card info to the payment instance
-        $quote->getPayment()->getMethodInstance()->getInfoInstance()->setData('cc_last_4', @$transaction->from_credit_card->last4);
-        $quote->getPayment()->getMethodInstance()->getInfoInstance()->setData('cc_type', @$transaction->from_credit_card->network);
+        $quote->getPayment()->getMethodInstance()->getInfoInstance()->setData(
```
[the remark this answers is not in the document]
danac-gs: @YevhenShBolt done

Model/Api/ShippingMethods.php
```diff
@@ -266,22 +283,26 @@ public function getShippingMethods($cart, $shipping_address)
...
                 );
             }
 
-            $this->checkoutSession->replaceQuote($quote);
-
             $this->checkCartItems($cart, $quote);
 
+            if ($customerId = $quote->getCustomerId()) {
+                $this->customerSession->setCustomer(
+                    $this->customerFactory->create()->load($customerId)
```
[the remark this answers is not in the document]
danac-gs: @YevhenShBolt 
Using QuoteRepository instead of QuoteFacctory now.
CustomerRepository instead of CustomerFactory an this place throws an error. Leaving that part the way it was.

PR #37, opened 2018-06-09
Bump version to 1.0.4

danac-gs:
Version numbers are incremented in composer.json and module.xml. It is mandatory for DB schema to be upgraded. New release tag needed.


daisy1754:
Good stuff. Can we also include fix for https://app.asana.com/0/436430054069245/704352657054200/f ?

PR #42, opened 2018-06-15
Use explicit flag to check if there is on-going request to server

daisy1754:
Before this PR, we use `cart.orderToken` as flag and set empty value to indicate "there is ongoing request" but it was problematic because race condition can lead bolt to read empty order token.

### DIFF
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -378,7 +378,7 @@ window.boltConfig = <?php echo $block->getSettings(); ?>;
 
         check: function () {
             if (createRequest || !cart.orderToken) {
-                console.log("Retrieving order token from server... not ready to checkout yet."):
+                console.log("Retrieving order token from server... not ready to checkout yet.");
                 return false;
             }
             return true;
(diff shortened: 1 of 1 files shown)


view/frontend/templates/js/replacejs.phtml
```diff
@@ -376,11 +376,12 @@ window.boltConfig = <?php echo $block->getSettings(); ?>;
             save_request = ajaxPost(settings.save_order_url, parameters, onSuccess);
         },
 
-        // the Bolt order is created right after the checkout button is clicked
-        // and the checkout modal popup is opened only if order creation was successfull
         check: function () {
-            // if orderToken is set the checkout window will open
-            return !!cart.orderToken;
+            if (createRequest || !cart.orderToken) {
+                console.log("Retrieving order token from server... not ready to checkout yet."):
```
danac-gs: change : to ; at the end of the line

danac-gs: change : to ; at the end of the line

daisy1754:
@danac-gs feel free to merge once you confirm typo fix

PR #53, opened 2018-06-26
add Js Block unit tests

danac-gs:
I merged develop into this branch. We could keep it after merging back.

### DIFF
--- a/Test/Unit/Block/JsTest.php
+++ b/Test/Unit/Block/JsTest.php
@@ -50,6 +50,11 @@ class JsTest extends \PHPUnit\Framework\TestCase
     protected $block;
 
     /**
+     * @var \Magento\Quote\Model\Quote
+     */
+    protected $quoteMock;
+
+    /**
      * @inheritdoc
      */
     protected function setUp()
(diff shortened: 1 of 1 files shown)


YevhenShBolt:
I somewhere saw the article where we can use getMockBuilder instead of invokeMethod to call the private method... but now i cannot find it, if i find it i will share with team.

Test/Unit/Block/JsTest.php
```diff
@@ -241,4 +247,55 @@ public function setSandboxMode($value = true)
             ->method('isSandboxModeSet')
             ->will($this->returnValue($value));
     }
+
+    /**
+     * @inheritdoc
+     */
+    public function testGetJavascriptSuccess()
```
daisy1754: You aren't testing anything meaningful here :)
This test is testing mock framework, not our code.
I suggest removing this.

danac-gs: removed

Test/Unit/Block/JsTest.php
```diff
@@ -241,4 +247,39 @@ public function setSandboxMode($value = true)
...
+        $method = $reflection->getMethod($methodName);
+        $method->setAccessible(true);
+
+        return $method->invokeArgs($object, $parameters);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function testGetQuoteIsVirtual()
```
daisy1754: again, I think this test doesn't test anything interesting. tested method is basically one-line getter so having test doesn't provide much value here.

PR #55, opened 2018-07-02
Cart data cloned quote id placeholder

danac-gs:
Set the order_reference to the parent quote id, preventing order duplication on Bolt side, thus preventing it on Magento too.
Use cloned quote id (immutable quote) on Magento side to prevent cart data mismatches.
Transmit cloned quote id in display_id field alongside reserved order id, concatenated by ' / '.

### DIFF
--- a/Controller/Cart/Data.php
+++ b/Controller/Cart/Data.php
@@ -100,15 +100,15 @@ class Data extends Action
             $response = $boltpayOrder ? $boltpayOrder->getResponse() : null;
 
             // get immutable quote id stored with cart data
-            list(, $orderReference) = $response ? explode(' / ', $response->cart->display_id) : [null, ''];
+            list(, $cartReference) = $response ? explode(' / ', $response->cart->display_id) : [null, ''];
 
             $cart = [
                 'orderToken'  => $response ? $response->token : '',
                 'authcapture' => $this->configHelper->getAutomaticCaptureMode(),
-                'cartReference' => $orderReference,
+                'cartReference' => $cartReference,
             ];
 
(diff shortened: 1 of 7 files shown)


Controller/Cart/Data.php
```diff
@@ -99,15 +99,16 @@ public function execute()
             // format and send the response
             $response = $boltpayOrder ? $boltpayOrder->getResponse() : null;
 
-            $orderReference = $response ? $response->cart->order_reference : '';
+            // get immutable quote id stored with cart data
+            list(, $cartReference) = $response ? explode(' / ', $response->cart->display_id) : [null, ''];
```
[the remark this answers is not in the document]
danac-gs: No. Sometimes we need first, sometimes second, sometimes both of them.
Here only the second is used.

Helper/Order.php
```diff
@@ -498,21 +498,23 @@ private function deleteRedundantQuotes($quote) {
...
+        // get order id and immutable quote id stored with transaction
+        list($incrementId, $quoteId) = explode(' / ', $transaction->order->cart->display_id);
 
         // check if the order exists
         $order = $this->cartHelper->getOrderByIncrementId($incrementId);
 
+        // Load (immutable) quote from entity id
+        $quote = $this->cartHelper->getQuoteById($quoteId);
+
         // if not create the order
```
daisy1754: nit: update this comment or re-arrange statements - I think this is continued sentence from "// check if the order exists" above

danac-gs: re-arranged

PR #58, opened 2018-07-04
replace freegeoip code to ipstack

danac-gs:


Controller/Shipping/Prefetch.php
```diff
@@ -65,8 +66,14 @@ class Prefetch extends Action
...
 
-    // GeoLocation API endpoint
-    private $locationURL = "http://freegeoip.net/json/%s";
+    /**
+     * @var CacheInterface
+     */
+    private $cache;
+
+    // GeoLocation API
+    const   API_KEY = '18beed19bdef35bf26f49f2e53c63842';
```
daisy1754: I'm little uncomfortable putting this in public github repo, since we're charged based on number of request limit. can you add config in admin for this?

YevhenShBolt:
I advise moving the Geoip code to the separate helper or model. In the future, we can use it for our own needs.

Controller/Shipping/Prefetch.php
danac-gs: Adding

danac-gs:
@YevhenShBolt weather we use it or not for other purposes in the future your proposal makes sense. Keeping 3rd party stuff separate is a good design. Will do that way.

etc/adminhtml/system.xml
```diff
@@ -57,6 +57,10 @@
                     <label>Automatic Capture Mode</label>
                     <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
                 </field>
+                <field id="geolocation_api_key" translate="label" type="obscure" sortOrder="85" showInDefault="1" showInWebsite="1" showInStore="0">
+                    <label>Geolocation API Key</label>
```
daisy1754: Add comment:
> If specified, Bolt plugin performs geolocation lookup to optimize the performance. Get key from https://ipstack.com/ or contact Bolt team

Helper/Geolocation.php
```diff
@@ -0,0 +1,184 @@
...
+            return null;
+        }
+    }
+
+    /**
+     * @param bool $useCache    flag that represents if Magento cache is used for reading / storing the result
+     * @return null|string      JSON formated response
+     * @throws \Zend_Http_Client_Exception
+     */
+    public function getLocation($useCache = true) {
```
daisy1754: nit: currently we always set $useCache as true. Let's remove this param until we need it.

danac-gs: done

etc/adminhtml/system.xml
danac-gs: done

PR #61, opened 2018-07-08
add index on quote is_active column

danac-gs:


YevhenShBolt:
For what we need quote index by field "is_active"?

danac-gs:
After examining the query plan I can confirm that this index is not used. Closing PR.

PR #62, opened 2018-07-10
Deactivate parent quote on order create

danac-gs:


daisy1754:
@danac-gs can you add comment to this pull request why you want to do this?

danac-gs:
@daisy1754  If by any chance the quotes do not get deleted and the parent one remains in active state it can cause the "already paid for quote id on bolt side" issue in the current session or future sessions for logged in customer. Deactivating it covers that case.

PR #63, opened 2018-07-10
move cleaning quotes up one level

danac-gs:
Try running deactivateParentQuote and deleteRedundantQuotes on every hook call not just on order create. In case something went wrong and the quotes were not cleared first time, parent quote can remain attached to the session and that might cause issues with the logged in user.

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -427,12 +427,6 @@ class Order extends AbstractHelper
 
         $order = $this->quoteManagement->submit($quote);
 
-        // Set parent quote as inactive
-        $this->deactivateParentQuote($quote);
-
-        // Delete redundant clones and parent quote
-        $this->deleteRedundantQuotes($quote);
-
         if ($frontend) {
             // Send order confirmation email to customer.
             $this->emailSender->send($order);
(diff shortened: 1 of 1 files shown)


Helper/Order.php
```diff
@@ -501,9 +495,20 @@ private function deleteRedundantQuotes($quote) {
...
     public function deactivateParentQuote($quote) {
-        $parentQuote = $this->cartHelper->getQuoteById($quote->getBoltParentQuoteId());
-        $parentQuote->setIsActive(false);
-        $this->cartHelper->saveQuote($parentQuote);
+        try {
+            $parentQuote = $this->cartHelper->getQuoteById($quote->getBoltParentQuoteId());
+            $parentQuote->setIsActive(false);
+            $this->cartHelper->saveQuote($parentQuote);
+        } catch (NoSuchEntityException $e) {
+            $this->bugsnag->registerCallback(function ($report) use ($quote) {
```
daisy1754: here you're swallowing error. shouldn't we notifyExcepgtion here?

danac-gs: @daisy1754 This method, deactivateParentQuote, as well as deleteRedundantQuotes, runs on initial order save (from frontend or hook) and every subsequent hook to ensure the status of quote/order in magento are as expected. Subsequent calls are there in case, from any unforeseen reason, the quotes are not cleared. For example, deleteRedundantQuotes, which is raw query, does not succeed due to some custom, store specific constraints. If error is thrown here in finding the quote means all is good and quote already deleted so we do not have / need anything to deactivate. After all, there is a bugsnag callback registered so we will have this info if exception is raised at a later point. I might add order increment id to it.

PR #64, opened 2018-07-11
Shipping discount

danac-gs:
Take discount on shipping into account when calculating shipping and tax

### DIFF
--- a/Helper/Config.php
+++ b/Helper/Config.php
@@ -96,6 +96,11 @@ class Config extends AbstractHelper
     const XML_PATH_PREFETCH_SHIPPING = 'payment/boltpay/prefetch_shipping';
 
     /**
+     * Reset Shipping Calculation
+     */
+    const XML_PATH_RESET_SHIPPING_CALCULATION = 'payment/boltpay/reset_shipping_calculation';
+
+    /**
      * Enabled
      */
     const XML_PATH_ACTIVE = 'payment/boltpay/active';
@@ -420,7 +425,6 @@ class Config extends AbstractHelper
      */
     public function getAutomaticCaptureMode($store = null)
     {
-        //Automatic capture mode
         return $this->getScopeConfig()->isSetFlag(
             self::XML_PATH_AUTOMATIC_CAPTURE_MODE,
             ScopeInterface::SCOPE_STORE,
@@ -437,9 +441,24 @@ class Config extends AbstractHelper
      */
     public function getPrefetchShipping($store = null)
     {
-        //Automatic capture mode
         return $this->getScopeConfig()->isSetFlag(
             self::XML_PATH_PREFETCH_SHIPPING,
+            ScopeInterface::SCOPE_STORE,
+            $store
+        );
+    }
+
+    /**
+     * Get Reset Shipping Calculation config
+     *
+     * @param int|string|Store $store
+     *
+     * @return  boolean
+     */
+    public function getResetShippingCalculation($store = null)
+    {
+        return $this->getScopeConfig()->isSetFlag(
+            self::XML_PATH_RESET_SHIPPING_CALCULATION,
             ScopeInterface::SCOPE_STORE,
             $store
         );
--- a/Model/Api/ShippingMethods.php
+++ b/Model/Api/ShippingMethods.php
@@ -39,6 +39,7 @@ use Magento\Framework\Webapi\Rest\Request;
 use Magento\Framework\App\CacheInterface;
 use Magento\Customer\Model\Session as CustomerSession;
 use Magento\Customer\Model\CustomerFactory;
+use Magento\Framework\Pricing\Helper\Data as PriceHelper;
 
 /**
  * Class ShippingMethods
@@ -136,12 +137,18 @@ class ShippingMethods implements ShippingMethodsInterface
      */
     private $customerFactory;
 
+    /**
+     * @var PriceHelper
+     */
+    private $priceHelper;
+
     // Totals adjustment threshold
-    private $threshold = 0.01;
+    private $threshold = 1;
 
     private $taxAdjusted = false;
 
     /**
+     * Assigns local references to global resources
      *
      * @param HookHelper $hookHelper
      * @param RegionModel $regionModel
@@ -160,6 +167,7 @@ class ShippingMethods implements ShippingMethodsInterface
      * @param CacheInterface $cache
      * @param CustomerSession $customerSession
      * @param CustomerFactory $customerFactory
+     * @param PriceHelper $priceHelper
      */
     public function __construct(
         HookHelper $hookHelper,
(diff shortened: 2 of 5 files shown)


Model/Api/ShippingMethods.php
```diff
@@ -526,15 +535,22 @@ private function getShippingOptions($quote, $addressData)
             ////////////////////////////////////////////////////////////////
 
             $shippingAddress->setShippingMethod($method);
+            // in order to get correct shipping discounts
+            // the following method must be called twice
+            $this->totalsCollector->collectAddressTotals($quote, $shippingAddress);
```
johnmccombs1: I'm not sure how similar this is to running collectTotals() on Magento 1, but if they are the same, this workaround may cause unintended consequences later on since it's based on calling collectTotals an even number of times (the second time rounds it back down). If so, you may want to look at what was done here https://github.com/BoltApp/bolt-magento1/commit/59fa6c3f377c62b7c51a149d4edbb3710b49789c

Model/Api/ShippingMethods.php
```diff
@@ -136,8 +137,13 @@ class ShippingMethods implements ShippingMethodsInterface
...
     private $customerFactory;
 
+    /**
+     * @var PriceHelper
+     */
+    private $priceHelper;
+
     // Totals adjustment threshold
-    private $threshold = 0.01;
+    private $threshold = 0.5;
```
dumega: I am curious about this threshold, why do we update it to 0.5? For me, I think it is bigger than expect. Why not 0.1? is there any principle on such a threshold?

Model/Api/ShippingMethods.php
```diff
@@ -526,15 +535,22 @@ private function getShippingOptions($quote, $addressData)
...
+            $cost        = $shippingAddress->getShippingAmount() - $shippingAddress->getShippingDiscountAmount();
             $roundedCost = $this->cartHelper->getRoundAmount($cost);
 
             $diff = $cost * 100 - $roundedCost;
 
             $taxAmount = $this->cartHelper->getRoundAmount($shippingAddress->getTaxAmount() + $diff / 100);
 
+            $discount = $this->priceHelper->currency($shippingAddress->getShippingDiscountAmount(), true, false);
+
+            if ($shippingAddress->getShippingDiscountAmount()) $service .= " [$discount" . "_discount]";
```
daisy1754: 1. Can you remove underscore? eg., "[$5 discount]"
2. If shipping is free after discount, can you instead say " [free shipping discount]"?

Model/Api/ShippingMethods.php
```diff
@@ -526,15 +535,22 @@ private function getShippingOptions($quote, $addressData)
...
 
-            $cost         = $shippingAddress->getShippingAmount();
+            $cost        = $shippingAddress->getShippingAmount() - $shippingAddress->getShippingDiscountAmount();
             $roundedCost = $this->cartHelper->getRoundAmount($cost);
 
             $diff = $cost * 100 - $roundedCost;
 
             $taxAmount = $this->cartHelper->getRoundAmount($shippingAddress->getTaxAmount() + $diff / 100);
 
+            $discount = $this->priceHelper->currency($shippingAddress->getShippingDiscountAmount(), true, false);
```
daisy1754: put this in if statement below

Model/Api/ShippingMethods.php
danac-gs: Actually the intended purpose for this threshold was dealing with 1+ cent totals mismatch. Since there was nothing here calculated by the plugin but all the values taken from Magento we did not have the mismatch case so far. Anyway, what was eventually rounded / cut off from shipping amount was added to the tax amount to re-balance the total. Now that eventual discount is subtracted I started getting bugsnag logs and realized that this threshold specifies cents, so it was too low.

Model/Api/ShippingMethods.php
danac-gs: done

Model/Api/ShippingMethods.php
danac-gs: @daisy1754 Will do as you say. The reasons I did it this way:
1. underscore was put to prevent splitting into two lines
2. free shipping discount is another discount type, this one simply subtracts the discount amount from the shipping, and if the shipping is <= discount it becomes 0, free.
But as I said, will do as requested.

mccottry:
For the  `$resetShippingCalculation` method, possibly as a different task, consider adding this as an admin option to be turned on or off by merchants, or placed in a custom config file.

Model/Api/ShippingMethods.php
mccottry: I don't believe that number makes sense because you are comparing what theoretically should be the difference between two integers.  I understand that it may have initially been comparing two floats.

The logical value for this in the context of a check like
```
if (abs($diff) >= $this->threshold) { ...
```
would be `private $threshold = 1;`  Yes, 0.5 would accomplish a similar thing, but the integer 1 is more clear.

danac-gs: @mccottry Correct. Fixed.

Model/Api/ShippingMethods.php
```diff
@@ -526,15 +539,30 @@ private function getShippingOptions($quote, $addressData)
             ////////////////////////////////////////////////////////////////
 
             $shippingAddress->setShippingMethod($method);
+            // in order to get correct shipping discounts
+            // the following method must be called twice
```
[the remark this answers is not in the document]
danac-gs: Done.

Helper/Config.php
```diff
@@ -445,6 +450,23 @@ public function getPrefetchShipping($store = null)
...
+    /**
+     * Get Reset Shipping Calculation config
+     *
+     * @param int|string|Store $store
+     *
+     * @return  boolean
+     */
+    public function getResetShippingCalculation($store = null)
+    {
+        //Automatic capture mode
```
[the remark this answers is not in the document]
danac-gs: Done. Removed redundant comments.

etc/adminhtml/system.xml
```diff
@@ -78,11 +78,15 @@
                 <field id="prefetch_address_fields" translate="label" type="textarea" sortOrder="110" showInDefault="1" showInWebsite="1" showInStore="0">
                     <label>Prefetch Shipping and Tax: Cache Identifier Custom Address Fields</label>
                 </field>
-                <field id="javascript_success" translate="label" type="textarea" sortOrder="110" showInDefault="1" showInWebsite="1" showInStore="0">
+                <field id="reset_shipping_calculation" translate="label" type="select" sortOrder="115" showInDefault="1" showInWebsite="1" showInStore="0">
+                    <label>Reset Shipping Calculation</label>
+                    <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
```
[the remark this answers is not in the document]
danac-gs: Done.

Model/Api/ShippingMethods.php
```diff
@@ -445,6 +459,25 @@ function ($el) {
...
+     * some method in the Magento shipping flow, then that method
+     * may be (indirectly) called from our Shipping And Tax flow more
+     * than once, resulting in wrong prices. This function resets
+     * address shipping calculation but can seriously slow down the
+     * process (on a system with many shipping options available).
+     * Use it carefully only when necesarry.
+     *
+     * @param \Magento\Quote\Model\Quote\Address $shippingAddress
+     */
+    private function resetShippingCalculation ($shippingAddress) {
```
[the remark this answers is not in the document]
danac-gs: Done

PR #67, opened 2018-07-18
add cart item type

danac-gs:


view/frontend/templates/js/replacejs.phtml
```diff
@@ -437,10 +437,44 @@ if (!$block->isEnabled()) return;
...
+                ///////////////////////////////////////////////////////////
+                if (getCheckoutType() === 'payment') {
+
+                    // trigger click on boltpay radio
+                    if ($('#boltpay').prop('checked') === false) $('#boltpay').click();
+
+                    // stop if customer email field exists and is not valid on paymtnt only page
+                    var customer_email = $(customer_email_selector);
+                    var form = customer_email.closest('form');
+                    if (form.validation() && form.validation('isValid') === false) {
```
daisy1754: can you scroll to bad input field? 
my concern is if email input is outside of window it is not obvious for user

view/frontend/templates/js/replacejs.phtml
```diff
@@ -437,10 +437,44 @@ if (!$block->isEnabled()) return;
...
 
             check: function () {
+
                 if (createRequest || !cart.orderToken) {
                     console.log("Retrieving order token from server... not ready to checkout yet.");
                     return false;
                 }
+
+                ///////////////////////////////////////////////////////////
+                // On Bolt button click check if guest checkout is allowed.
```
daisy1754: maybe this can be php comment so it won't included in JavaScript?

PR #72, opened 2018-07-19

danac-gs:
It's not finished yet but is pretty much isolated and does not affect other parts of the plugin. Merging might be necessary for testing on http://magento2-sandbox.bolt.me/ if the develop branch is automatically deployed.


Model/Api/DiscountCodeValidation.php
```diff
@@ -205,111 +206,187 @@ public function validate() {
...
-            // Load the and validate the quote.
-            $quote_id = $request->cart->order_reference;
-            if (empty($quote_id)) {
-                return $this->sendErrorResponse(self::ERR_INSUFFICIENT_INFORMATION, 'The order reference is invalid.', 422);
+            // check if the order has already been created
+            if ($this->cartHelper->getOrderByIncrementId($incrementId)) {
+                return $this->sendErrorResponse(
+                    self::ERR_INSUFFICIENT_INFORMATION,
+                    sprintf('The order #%s has already been created.', $parentQuoteId),
+                    404
```
gow: This should be 422.

Helper/Cart.php
```diff
@@ -871,7 +877,7 @@ public function getCartData($paymentOnly, $placeOrderPayload, $immutableQuote =
             $this->bugsnag->notifyError('Cart Totals Mismatch', "Totals adjusted by $diff.");
         }
 
-        //$this->logHelper->addInfoLog(json_encode($cart, JSON_PRETTY_PRINT));
+        // $this->logHelper->addInfoLog(json_encode($cart, JSON_PRETTY_PRINT|JSON_UNESCAPED_SLASHES));
```
gow: nit: if it's unused, let's just remove it.

PR #73, opened 2018-07-19

danac-gs:


Model/Api/DiscountCodeValidation.php
```diff
@@ -202,7 +202,7 @@ public function validate() {
                 'X-Bolt-Plugin-Version' => $this->configHelper->getModuleVersion()
             ]);
 
-            $this->hookHelper->verifyWebhook();
+            //$this->hookHelper->verifyWebhook();
```
gow: remove if not needed.

danac-gs: it's needed but left commented out while I was doing test from a REST client.

PR #75, opened 2018-07-20
Return proper discount_type

gow:


Model/Api/DiscountCodeValidation.php
```diff
@@ -421,4 +416,17 @@ private function sendSuccessResponse($result, $quote) {
         $this->response->sendResponse();
         return;
     }
+
+    private function convertToBoltDiscountType($type): string {
```
danac-gs: @gow Function return type declarations are supported since PHP7. Some of the Magento store versions that we support can be run on PHP5.6. I will update this, approve and merge into my branch so I can move on. You can check it and merge to develop later.

PR #76, opened 2018-07-20

danac-gs:
Error might or might not nullify the previously applied discount. it depends when it happens.

The cart data is returned in error response whenever it is available so the checkout cart could be updated accordingly.

Removed payment_only cart creation option. will add it later when we start to support it.


danac-gs:
@daisy1754 @gow 

Cart data response is reduced to total_amount, tax_amount and discounts array. As the shipping is called after discount validation it can probably be ignored and removed in the future if it is not needed.

The quote is not cloned again, the discount is applied to the parent quote and existing clone used in Bolt order, and updated clone data sent back.

PR #81, opened 2018-08-08
Updated the DiscountCodeValidation model.

YevhenShBolt:
- Moved a coupon applying process to the separate method
- Added the gift card applying process, also placed in a separate method.
- Fixed the returning statements for the errors.
- Fixed a message returning an error to the Bugsnag in exception

### DIFF
--- a/Model/Api/DiscountCodeValidation.php
+++ b/Model/Api/DiscountCodeValidation.php
@@ -1,26 +1,28 @@
 <?php
 /**
-* Bolt magento2 plugin
-*
-* NOTICE OF LICENSE
-*
-* This source file is subject to the Open Software License (OSL 3.0)
-* that is bundled with this package in the file LICENSE.txt.
-* It is also available through the world-wide-web at this URL:
-* http://opensource.org/licenses/osl-3.0.php
-*
-* @category   Bolt
-* @package    Bolt_Boltpay
-* @copyright  Copyright (c) 2018 Bolt Financial, Inc (https://www.bolt.com)
-* @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
-*/
+ * Bolt magento2 plugin
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ *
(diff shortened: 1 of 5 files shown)


Model/Api/DiscountCodeValidation.php
```diff
@@ -444,21 +538,27 @@ private function sendErrorResponse($errCode, $message, $httpStatusCode, $quote =
...
      * @param array $result
      * @param Quote $quote
+     * @return array
      * @throws \Exception
      */
-    private function sendSuccessResponse($result, $quote) {
+    private function sendSuccessResponse($result, $quote)
+    {
         $result['cart'] = $this->getCartTotals($quote);
+        $this->bugsnag->notifyError('Discount Validation Response', json_encode($result));
```
daisy1754: I'm sure it's for debugging; please make sure you remove it

Model/Api/DiscountCodeValidation.php
```diff
@@ -288,130 +295,213 @@ public function validate() {
...
-                        \IntlDateFormatter::MEDIUM
-                    );
-                return $this->sendErrorResponse(
-                    self::ERR_CODE_NOT_AVAILABLE,
-                    $desc,
-                    422,
-                    $immutableQuote
-                );
-            }
+            return $this->sendSuccessResponse($result, $immutableQuote);
```
daisy1754: applyingCouponCode method may call sendErrorResponse method internally. I guess in such case we shouldn't call sendSuccessResponse?

Model/Api/DiscountCodeValidation.php
YevhenShBolt: I was sure that I removed it.

Model/Api/DiscountCodeValidation.php
YevhenShBolt: Yes sure, i will think and refactor it.

YevhenShBolt: I've added check above

etc/module.xml
```diff
@@ -24,6 +24,7 @@
             <module name="Magento_Sales"/>
             <module name="Magento_Checkout"/>
             <module name="Magento_Tax"/>
+            <module name="Magento_GiftCardAccount"/>
```
danac-gs: What happens here if the module does not exist? Assuming it is ignored gracefully.

YevhenShBolt: Nothing, i checked. It just determines (or like) the order of module load.

PR #84, opened 2018-08-23
poplate hints from response before cart promise is resolved

danac-gs:


YevhenShBolt:
The problem is in:
1) Bolt\Boltpay\Test\Unit\Block\JsTest::testGetSettings
Error: Call to a member function getValue() on null
/home/travis/build/magento/app/code/Bolt/Boltpay/Helper/Config.php:384
/home/travis/build/magento/app/code/Bolt/Boltpay/Block/Js.php:109
/home/travis/build/magento/app/code/Bolt/Boltpay/Block/Js.php:160
/home/travis/build/magento/app/code/Bolt/Boltpay/Test/Unit/Block/JsTest.php:200

PR #92, opened 2018-09-05
Add additional button class option

YevhenShBolt:
https://app.asana.com/0/436430054069245/806908348117542/f


view/frontend/templates/js/replacejs.phtml
```diff
@@ -416,15 +420,6 @@ if (!$block->isEnabled()) return;
                 if (callbacks.success_url) location.href = callbacks.success_url;
             },
 
-            onCheckoutStart: function() {
```
daisy1754: why remove?

YevhenShBolt: This was for debugging. It shows when the pop-up window opens.

PR #98, opened 2018-09-12
Api emulate session

danac-gs:


Helper/Session.php
```diff
@@ -0,0 +1,168 @@
...
+
+            if ($sessionData["sessionType"] == "frontend") {
+                // shipping and tax, orphaned transaction
+                // cart belongs to logged in customer?
+                if ($customerId) {
+                    $this->setSession($this->checkoutSession, $sessionID);
+                    $this->setSession($this->customerSession, $sessionID);
+                }
+            } else {
+                // orphaned transaction
```
daisy1754: orphan? admin?

Helper/Session.php
```diff
@@ -0,0 +1,168 @@
...
+        if ($serialized = $this->cache->load($cacheIdentifier)) {
+
+            $sessionData = unserialize($serialized);
+            $sessionID = $sessionData["sessionID"];
+
+            if ($sessionData["sessionType"] == "frontend") {
+                // shipping and tax, orphaned transaction
+                // cart belongs to logged in customer?
+                if ($customerId) {
+                    $this->setSession($this->checkoutSession, $sessionID);
```
daisy1754: even if you're not logged in you still want to set this session, no?

Helper/Session.php
danac-gs: @daisy1754 why not? connection could be lost in the middle of communication the same as on frontend. tested. works.

Helper/Session.php
danac-gs: @daisy1754 In general you do, but in this case you don't. When called from shipping and tax and the session in the API call is replaced this way for anonymous user the basic session gets lost. In other words after you collect the shipping options you lose the cart data on the main page (reload). 
```
$this->replaceQuote($immutableQuote);
```
several lines below is sufficient for non logged in user case since all the data is in the quote.

PR #108, opened 2018-09-16
Call callback even when order creation AJAX fails

daisy1754:
Currently our custom dialog (`showBoltErrorMessage`) is hidden under Bolt checkout modal with infinite spinner.


YevhenShBolt:
Can we increase a z-index of an error popup?

daisy1754:
@YevhenShBolt  we can increase z-index but after user close this error dialog, Bolt would show infinite spinner unless we call `callback` here... I think assuming success is the best thing we can do with given options.

PR #110, opened 2018-09-18
unnested order note assignment

davidwebber10:
We need the ability to store an order comment in a different manner other than as a status history comment. In order to push to our ERP, we need to integrate this with an already defined order field. Unnesting the code that assigns the order comment to it's own function allows us to internally rewrite just that function and change where the comment is saved.

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -579,10 +579,9 @@ class Order extends AbstractHelper
 
             // Add the user_note to the order comments and make it visible for customer.
             if (isset($transaction->order->user_note)) {
-                $order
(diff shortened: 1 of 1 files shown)


Helper/Order.php
```diff
@@ -602,6 +601,25 @@ public function saveUpdateOrder($reference, $frontend = true, $bolt_trace_id = n
         }
     }
 
+    /**
+     * Add user note as status history comment
+     *
+     * @param OrderModel $order
+     * @param \stdClass $transaction
```
daisy1754: this should be string $userNote

daisy1754: I'll merge this anyway

PR #115, opened 2018-09-20
MiniCart fix for m2.1.9

YevhenShBolt:
Fix the problem when Mini cart not cleared after user place orders.

It was affected Magento till version 2.2.0


daisy1754:
@YevhenShBolt if you can add more comments (+maybe link to bug on magento2 github) in PR comments and in source code comment regarding what is issue and how we workaround, that'd be great.

YevhenShBolt:
Magento Issue: https://github.com/magento/magento2/issues/12504

PR #117, opened 2018-09-24
Initiate checkout on log in

danac-gs:
There are 2 ways of login, through 2 separate controllers, Account and Ajax. The code redirects to shopping cart page if the cart exists after user logs in through any of them. The logic is appended to the default controller action in form of a plugin, overriding the default result by adding a redirect.


Plugin/LoginPlugin.php
```diff
@@ -0,0 +1,102 @@
...
+            $response['redirectUrl'] = self::$shoppingCartPath;
+
+            // Set the flag in session to auto-open Bolt checkout on redirected (shopping cart) page
+            $this->setBoltOpenCheckoutFlag();
+
+            $resultJson = $this->resultJsonFactory->create();
+            return $resultJson->setData($response);
+
+        } catch (\Exception $e) {
+            // On any exception pass the original result through
```
daisy1754: let's bugsnag

Plugin/LoginPluginTrait.php
```diff
@@ -0,0 +1,58 @@
...
+     * @return bool
+     */
+    private function hasCart() {
+        return $this->checkoutSession->hasQuote() && count($this->checkoutSession->getQuote()->getAllVisibleItems()) > 0;
+    }
+
+    /**
+     * @return bool
+     */
+    private function allowRedirect() {
```
daisy1754: shouldRecirectToCartPage

YevhenShBolt:
Sorry @danac-gs, i posted the wrong review above.

danac-gs:
@daisy1754 @mccottry 
The code is refactored in a way that a parent abstract class is used instead of a trait.
Checkout auto-opens after login only if click on a bolt button triggered user authentication.

PR #118, opened 2018-09-27
Add Unirgy GiftCert support

YevhenShBolt:


YevhenShBolt:
Yes, it does not change any behavior if Unirgy does not present.

PR #123, opened 2018-10-01
removed redundant collect totals calls

danac-gs:


Helper/Order.php
```diff
@@ -501,7 +500,6 @@ public function deactivateParentQuote($quote) {
         try {
             $parentQuote = $this->cartHelper->getQuoteById($quote->getBoltParentQuoteId());
             $parentQuote->setIsActive(false);
-            $this->cartHelper->saveQuote($parentQuote);
```
daisy1754: intentional?

danac-gs: @daisy1754 yes. It is unnecessary, the quote is deleted in next step. And collectTotals, called in saveQuote flow, breaks things. It should be called once only in the session, for immutable quote (from saveQuote flow). The giftcard points are subtracted from the balance and added to quote in collectTotals (plugin or observer) and those two, now deleted, lines were the reason of a really nasty bug in saving the order, and, later, customer checkout / quote session data getting invalidated every time with the result in error in adding to cart and showing the shopping cart page.

PR #125, opened 2018-10-09
Webhook rewrite

danac-gs:
Drop dependence on transaction timeline field, rely on last transaction state.


Helper/Order.php
```diff
@@ -619,329 +663,331 @@ public function setOrderUserNote($order, $userNote){
...
-            ];
-
-            // there is a new transaction to be recorded
-            if ($transaction_type) {
-                // create the invoice for payment / capture transaction
-                if ($invoice_transaction) {
-                    $this->createOrderInvoice($order, $invoice_transaction, $amount->amount / 100);
+                if ($fromHook) {
+                    // Refunds need to be initiated from ste store admin (Invoice -> Credit Memo)
+                    // If called from Bolt merchant dashboard there is no enough info to sync the totals
```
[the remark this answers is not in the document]

To do a refund from Magento2 you need to specify the items you refund, as well as the shipping if you also refund the shipping. If done from the store admin it works perfect, even partial refunds. The data is collected internally and only the total sent to Bolt. On successful response it gets saved correctly. Refunds are triggered from the Invoice page (Credit Memo button)

There is another option, offline refund from the order page, which records the refund of the total amount, no partial refund, no options. Maybe we could use that for a specific case when only a complete refund is called from the merchant dashboard, but I'd rather not. Merchant should know to do refunds from the store. If they eventually call if from the dashboard they can sync by doing offline refund from the invoice page (not the same option as aforementioned order page offline refund). There they can specify the items and also have some options to tweak the total if needed.

Model/Payment.php
```diff
@@ -213,16 +213,16 @@ public function cancel(InfoInterface $payment)
     public function void(InfoInterface $payment)
     {
         try {
-            $hookTransactionId = $payment->getAdditionalInformation('real_transaction_id');
+            $transactionId = $payment->getAdditionalInformation('real_transaction_id');
 
-            if (empty($hookTransactionId)) {
+            if (empty($transactionId)) {
                 throw new LocalizedException(
                     __('Please wait while transaction get updated from Bolt.')
```
mccottry: nit: *gets

Model/Payment.php
```diff
@@ -277,6 +277,7 @@ public function fetchTransactionInfo(InfoInterface $payment, $transactionId)
                 $order = $payment->getOrder();
                 $this->orderHelper->updateOrderPayment($order, null, $transactionReference);
             }
+
```
mccottry: @danac-gs Is my understanding correct in that every time you fetch the transaction data from Bolt, you check whether an update is needed and update the payment according?  (I would imagine this could or does serve the role of implementing an merchant-side "Update Payment" button where the merchant syncs his actions with Bolt.)

danac-gs: @mccottry Correct. In M2 there is a "Fetch" button on the transaction page that does exactly that.

Helper/Order.php
```diff
@@ -619,329 +663,331 @@ public function setOrderUserNote($order, $userNote){
...
+     * @param string $reference
+     * @return string
+     */
     public function formatReferenceUrl($reference) {
         $url = $this->configHelper->getMerchantDashboardUrl().'/transaction/'.$reference;
         return '<a target="_blank" href="'.$url.'">'.$reference.'</a>';
     }
 
+    /**
+     * @param mixed $transaction
```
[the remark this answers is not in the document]
danac-gs: @mccottry json_decode in general returns mixed, that's why it's there, but in our particular case it is always stdClass, so I am updating it, thanks.

PR #131, opened 2018-10-15
add store url to bugsnag metadata, DRY - move adding API metadata and…

danac-gs:
… headers to Hook Helper

### DIFF
--- a/Helper/Bugsnag.php
+++ b/Helper/Bugsnag.php
@@ -89,7 +89,7 @@ class Bugsnag extends AbstractHelper
         $this->bugsnag->getConfig()->setReleaseStage($release_stage);
         $this->bugsnag->getConfig()->setAppVersion($this->configHelper->getModuleVersion());
 
-        $this->addCommonMetadata();
+        $this->addCommonMetaData();
 
         ////////////////////////////////////////////////////////////////////////
         // Reporting unhandled exceptions. This option is turned off by default.
@@ -141,7 +141,7 @@ class Bugsnag extends AbstractHelper
     /**
      * Add metadata to every bugsnag log
      */
-    private function addCommonMetadata()
+    private function addCommonMetaData()
     {
         $this->bugsnag->registerCallback(function ($report) {
             $report->addMetaData([


YevhenShBolt:
Just want to clarify.

For me it is in vendor folder, it looks like:
- `vendor/bugsnag`
- `vendor/guzzle`

How did you get it in `lib/internal` folder?

danac-gs:
@YevhenShBolt It is in vendor as it should always be if installed via composer. This is added to support composerless install (requested by Capillus in the past). It was commented out but I uncommented it in order to not have to edit/change the code if we would have the similar request in the future. It does not break anything, can go like that. For composerless installs, on demand, we will provide the additional contents to put in the lib folder. It is not a part of the plugin core.

Helper/Hook.php
```diff
@@ -154,4 +168,31 @@ public function verifyWebhook()
...
     }
+
+    /**
+     * Set bugsnag metadata bolt_trace_id
+     */
+    public function setMetadata()
+    {
+        if ($bolt_trace_id = $this->request->getHeader(ConfigHelper::BOLT_TRACE_ID_HEADER)) {
+            $this->bugsnag->registerCallback(function ($report) use ($bolt_trace_id) {
+                $report->setMetaData([
```
daisy1754: shouldn't this addMetaData so it won't override existing metadata with store URL?

Helper/Hook.php
```diff
@@ -154,4 +168,31 @@ public function verifyWebhook()
             throw new WebapiException(__('Precondition Failed'), 6001, 412);
         }
     }
+
+    /**
+     * Set bugsnag metadata bolt_trace_id
+     */
+    public function setMetadata()
```
daisy1754: maybe setCommonMetaData?

Also nit: inconsistent capitalization about Metadata vs MetaData

Helper/Hook.php
danac-gs: @daisy1754 In fact it is the opposite. setMetaData by default uses array_merge_recursive which keeps multiple values in an array.

```php
    public function setMetaData(array $metaData, $merge = true)
    {
        $this->metaData = $merge ? array_merge_recursive($this->metaData, $metaData) : $metaData;

        return $this;
    }
    public function addMetaData(array $metadata)
    {
        $this->metaData = array_replace_recursive($this->metaData, $metadata);
        $this->metaData = $this->removeNullElements($this->metaData);

        return $this;
    }
```

Helper/Hook.php
danac-gs: @daisy1754 fixed. Helper::Bugsnag::addCommonMetadata() as well.

PR #132, opened 2018-10-15
Field for additional / voluntary / hidden configuration options

danac-gs:


Helper/Config.php
```diff
@@ -623,4 +628,36 @@ public function getGoogleTrackOnCheckoutStart()
...
+
+    /**
+     * Get Additional Config property
+     *
+     * @param $name
+     * @return mixed
+     */
+    private function getAdditionalConfigProperty($name) {
+        $config = $this->getAdditionalConfigObject();
+        return @$config->$name;
```
YevhenShBolt: Maybe better way to return array instead of the StdClass from $this->getAdditionalConfigObject(), is it?

I suggest make like `json_decode($this->getAdditionalConfigString(), true)` after that we have assoc array in `getAdditionalConfigObject` method.

Then here we can check is set this key and not use the `@`.

YevhenShBolt:
@danac-gs , Can you check the methods with this tools?

- https://devdocs.magento.com/guides/v2.2/coding-standards/code-standard-sniffers.html
- https://www.atwix.com/magento-2/configure-code-sniffer-for-phpstorm/

It seems that we should have bracket `{` on a new line after methods. These tools should help us to ensure that we meet Magento code standards.

Helper/Config.php
danac-gs: @YevhenShBolt I prefer using objects later when we actually use this method and the configuration options stored like this. They could be nested objects themselves. It is less verbose and somehow more natural. On the other hand, it may be just a personal preference.

🡣🡣🡣 fixed the brackets

daisy1754:
Where do we use this config? Will it be used by override?

danac-gs:
@daisy1754 It is the one additional config option that should have multiple purposes. It was mentioned on recent meetings. Can be used  for not exposed options, like bolt checkout button attributes. I am currently using it to define some parameters regarding toggle between Bolt and default checkout buttons for subscription products on Capillus. For the purpose the config is like this 
```
{
  "toggleCheckout": {
    "active": true,
    "magentoButtons": [
      "#top-cart-btn-checkout",
      "button[data-role=proceed-to-checkout]"
    ],
    "productRestrictionMethods": [
      "getSubscriptionActive"
    ],
    "itemRestrictionMethods": [
      "getIsSubscription"
    ]
  }
}
```

PR #133, opened 2018-10-15
Toggle checkout

danac-gs:
Bolt buttons are replaced with Magento default buttons (links) and vice versa
according to Bolt checkout restriction state. Bolt checkout may be restricted if there are
restricted items in cart, e.g. subscription products.


YevhenShBolt:
Please check with phpcs some classed. As an example, regarding a PSR2, some methods do not have a bracket in a new line.

danac-gs:
[WIP] flag put back, waiting for finishing the tests on Capillus before the merge.

PR #137, opened 2018-10-22
fix for required login bypass when checkout button is clicked too early

danac-gs:


mccottry:
@danac-gs 
code seems good.  unit test are failing, though it seems unrelated to the code.

PR #147, opened 2018-11-01
Implemented multiple capture

johnmccombs1:


Helper/Order.php
```diff
@@ -95,6 +97,7 @@ class Order extends AbstractHelper
             self::TS_ZERO_AMOUNT
         ],
         self::TS_AUTHORIZED => [
+            self::TS_AUTHORIZED,
```
danac-gs: Does this mean that partially captured transaction when fetched still has the type of **cc_payment** and status **authorized**?

Helper/Order.php
```diff
@@ -732,9 +743,6 @@ private function checkTotalsMismatch($order, $transaction)
             case self::TS_ZERO_AMOUNT:
                 $boltTotal = 0;
                 break;
-            case self::TS_COMPLETED:
```
danac-gs: Why is total mismatch check removed for completed transactions?

Helper/Order.php
```diff
@@ -886,17 +898,21 @@ public function updateOrderPayment($order, $transaction = null, $reference = nul
...
-                $transactionId = $transaction->id.'-capture';
                 $amount = $transaction->capture->amount->amount;
 
                 if ($prevTransactionState == self::TS_AUTHORIZED) {
                     $transactionType = Transaction::TYPE_CAPTURE;
-                    $transactionId = $transaction->id.'-capture';
+                    $transactionId = $transaction->id.'-capture-'.time();
                     $parentTransactionId = $transaction->id.'-auth';
                 } else {
                     $transactionType = Transaction::TYPE_PAYMENT;
```
danac-gs: Isn't setting of $transactionId ($transaction->id.'-capture') needed here?
Assuming it is not necessary, can stay at default,  $transaction->id, but maybe more precise name shown in admin can have its benefits.

Helper/Order.php
```diff
@@ -1015,46 +1032,181 @@ public function updateOrderPayment($order, $transaction = null, $reference = nul
...
+        /** @var OrderPaymentInterface $payment */
+        $payment = $order->getPayment();
+        $reference = $transaction->reference;
+
+        $boltCaptures = $this->getNewBoltCaptures($payment, $transaction);
+        // Create invoices for items from $boltCaptures that are not exists on Magento
+        $identifier = count($boltCaptures) > 1 ? 0 : null;
+        foreach ($boltCaptures as $captureAmount) {
+            $invoice = $this->createInvoice($order, $captureAmount / 100);
+            $invoice->setTransactionId($reference);
```
danac-gs: Is real transaction id as stored in Magento needed here?

Helper/Hook.php
```diff
@@ -35,6 +35,8 @@
  */
 class Hook extends AbstractHelper
 {
+    public static $fromBolt = false;
```
daisy1754: I'm not very familiar with how php/magento works. Is it safe to use static variables for request-scoped variable? ie., if magento get two requests at the same time, is this variable scoped to one request?

Model/Service/InvoiceService.php
```diff
@@ -0,0 +1,66 @@
+<?php
+namespace Bolt\Boltpay\Model\Service;
+
+use Bolt\Boltpay\Helper\Bugsnag;
+use Magento\Sales\Api\Data\OrderInterface;
+
+class InvoiceService extends \Magento\Sales\Model\Service\InvoiceService
```
daisy1754: can you add comment explaining why you need to override this?

Helper/Hook.php
johnmccombs1: No, this variable is scoped to each request.

Model/Service/InvoiceService.php
johnmccombs1: Sorry for the confusion. It's overriding the Magento core class. It is extending the Magento core class to reuse the Magento core method and property for our new method named prepareInvoiceWithoutItems()

Helper/Order.php
johnmccombs1: Yes, after a partial capture, the transaction still has an authorized status. When I implemented multiple capture on M1, it had a completed status. I think I need to recheck it on M1 and put in a fix for it as there is new logic now.

Helper/Order.php
johnmccombs1: I didn't touch this part, other than making it unique since there will now be more than one. Is there a reason you think it should only be "$transaction->id.'-capture'"?

danac-gs: @johnmccombs1 what I meant is putting $transactionId = $transaction->id.'-capture'; in `else {` section

johnmccombs1: Currently it's set to:
$transactionId = $transaction->id.'-payment';

Does it need to change to:
$transactionId = $transaction->id.'-capture';

Just let me know and I'll update it.

PR #149, opened 2018-11-12
tracking callback enabled for every checkout state

danac-gs:


etc/adminhtml/system.xml
```diff
@@ -101,16 +101,31 @@
                 <field id="additional_js" translate="label" type="textarea" sortOrder="140" showInDefault="1" showInWebsite="1" showInStore="0">
                     <label>Additional Javascript</label>
                 </field>
-                <field id="g_track_on_success" translate="label" type="textarea" sortOrder="160" showInDefault="1" showInWebsite="1" showInStore="0">
-                    <label>Tracking: Success</label>
+                <field id="g_track_on_checkout_start" translate="label" type="textarea" sortOrder="150" showInDefault="1" showInWebsite="1" showInStore="0">
+                    <label>Tracking: Shipping</label>
```
daisy1754: as discussed, let's name this as onCheckoutStart. same for others

view/frontend/templates/js/replacejs.phtml
```diff
@@ -510,6 +530,9 @@ if (!$block->isEnabled()) return;
 
                     <?php echo $block->getjavascriptSuccess(); ?>
                     callbacks.success_url = data.success_url;
+
+                    gBoltTrack.gSuccess();
```
daisy1754: try-catch so we can always call `callback`

danac-gs: @daisy1754 done

etc/adminhtml/system.xml
danac-gs: @daisy1754 done

PR #152, opened 2018-11-26
update partial capture, fix auto capture bug, update total mismatch c…

danac-gs:
…heck

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -110,7 +110,6 @@ class Order extends AbstractHelper
             self::TS_AUTHORIZED,
             self::TS_COMPLETED,
             self::TS_REJECTED_IRREVERSIBLE,
-            self::TS_CANCELED,
             self::TS_CANCELED
         ],
         self::TS_REJECTED_IRREVERSIBLE => [],
@@ -688,7 +687,7 @@ class Order extends AbstractHelper
     }
 
     /**
-     * Format the (class internal) unambiguous transaction state out of the type and status.
+     * Format the (class internal) unambiguous transaction state out of the type, status and previously recordered status.
      * Infer eventually missing states. Represent Bolt partial capture AUTHORIZED state as CAPTURED.
      *
      * @param \stdClass $transaction
@@ -701,6 +700,12 @@ class Order extends AbstractHelper
         $prevTransactionState = $payment->getAdditionalInformation('transaction_state');
         $prevCaptures = $payment->getAdditionalInformation('captures') ?: [];
 
+        // No previous state recorded.
+        // Unless the state is TS_ZERO_AMOUNT (valid start transaction state, as well as TS_PENDING)
+        // or TS_CREDIT_COMPLETED (for historical reasons, old orders refund,
+        // legacy code when order was created, no state recorded) put it in TS_PENDING state.
+        // It can corelate with the $transactionState or not in case the hook is late due connection problems and
+        // the status has changed in the meanwhile.
         if (!$prevTransactionState) {
             if (in_array($transactionState, [self::TS_ZERO_AMOUNT, self::TS_CREDIT_COMPLETED])) {
                 return $transactionState;
@@ -708,6 +713,9 @@ class Order extends AbstractHelper
             return self::TS_PENDING;
         }
 
+        // The previously recorded state is either TS_PENDING or TS_REJECTED_REVERSIBLE. Authorization occured but also
+        // some of the funds were captured before the TS_AUTHORIZED state is recorded in Magento. Mark the transaction
+        // as TS_AUTHORIZED and wait for the next hook to record the CAPTURE.
         if (in_array($prevTransactionState, [self::TS_PENDING, self::TS_REJECTED_REVERSIBLE]) &&
             $transactionState == self::TS_AUTHORIZED &&
             $transaction->captures
@@ -715,6 +723,9 @@ class Order extends AbstractHelper
             return self::TS_AUTHORIZED;
         }
 
+        // The previously recorded state is either TS_PENDING or TS_REJECTED_REVERSIBLE and the transaction state is
+        // TS_COMPLETED. If there is more than one capture in $transaction->captures array then the TS_AUTHORIZED state
+        // is missing. Set the state to TS_AUTHORIZED and process captures on next hook requests.
         if (in_array($prevTransactionState, [self::TS_PENDING, self::TS_REJECTED_REVERSIBLE]) &&
             $transactionState == self::TS_COMPLETED &&
             count($transaction->captures) > 1
@@ -722,6 +733,8 @@ class Order extends AbstractHelper
             return self::TS_AUTHORIZED;
         }
 
+        // The treansaction was in TS_AUTHORIZED state, now it's TS_COMPLETED but not all partial captures are
+        // processed. Mark it TS_CAPTURED.
         if ($prevTransactionState == self::TS_AUTHORIZED &&
             $transactionState == self::TS_COMPLETED &&
             count($transaction->captures) - count($prevCaptures) > 1
@@ -729,18 +742,22 @@ class Order extends AbstractHelper
             return self::TS_CAPTURED;
         }
 
+        // The treansaction was TS_AUTHORIZED, now it has captures, put it in TS_CAPTURED state.
         if ($transactionState == self::TS_AUTHORIZED &&
             $transaction->captures
         ) {
             return self::TS_CAPTURED;
         }
 
+        // Previous partial capture was partially or fully refunded. Transaction is still TS_AUTHORIZED on Bolt side.
+        // Set it to TS_CAPTURED.
         if ($prevTransactionState == self::TS_CREDIT_COMPLETED &&
             $transactionState == self::TS_AUTHORIZED
         ) {
             return self::TS_CAPTURED;
         }
 
+        // return transaction state as it is in fetched transaction info. No need to change it.
         return $transactionState;
     }
 
@@ -763,9 +780,9 @@ class Order extends AbstractHelper
      *
      * @param OrderModel $order
      * @param \stdClass $transaction
-     * @return void
+     * @return bool true if the order was placed on hold, otherwise false
      */
-    private function isTotalsMismatch($order, $transaction)
+    private function holdOnTotalsMismatch($order, $transaction)
     {
         $boltTotal = $transaction->order->cart->total_amount->amount;
         $storeTotal = round($order->getGrandTotal() * 100);
@@ -790,7 +807,7 @@ class Order extends AbstractHelper
         $order->addStatusHistoryComment($comment);
 
         // Get the quote id
-        list(, $quoteId) = array_pad(
+        list($incrementId, $quoteId) = array_pad(
             explode(' / ', $transaction->order->cart->display_id),
             2,
             null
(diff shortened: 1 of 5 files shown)


Helper/Order.php
```diff
@@ -97,20 +90,36 @@ class Order extends AbstractHelper
...
+            self::TS_COMPLETED,
+            self::TS_CREDIT_COMPLETED
+        ],
         self::TS_ZERO_AMOUNT => [],
         self::TS_REJECTED_REVERSIBLE => [
             self::TS_AUTHORIZED,
             self::TS_COMPLETED,
-            self::TS_REJECTED_IRREVERSIBLE
+            self::TS_REJECTED_IRREVERSIBLE,
+            self::TS_CANCELED,
```
dumega: duplicate `self::TS_CANCELED`

Helper/Order.php
```diff
@@ -660,10 +645,12 @@ public function saveUpdateOrder($reference, $frontend = true, $bolt_trace_id = n
...
 
-        // update order payment transactions
-        $this->updateOrderPayment($order, $transaction);
-
-        if ($frontend) {
+        if (Hook::$fromBolt) {
+            // if called from hook update order payment transactions
+            $this->updateOrderPayment($order, $transaction);
+        } else {
+            // if called from the store controller return quote and order
```
daisy1754: Just out of curiosity, why is that? shouldn't we attach payment even if it's still pending?

Helper/Order.php
```diff
@@ -701,14 +688,60 @@ public function formatReferenceUrl($reference)
...
-        return $transaction->type.":".$transaction->status;
+        $transactionState = $transaction->type.":".$transaction->status;
+        $prevTransactionState = $payment->getAdditionalInformation('transaction_state');
+        $prevCaptures = $payment->getAdditionalInformation('captures') ?: [];
+
+        if (!$prevTransactionState) {
+            if (in_array($transactionState, [self::TS_ZERO_AMOUNT, self::TS_CREDIT_COMPLETED])) {
+                return $transactionState;
+            }
+            return self::TS_PENDING;
```
daisy1754: how will that work for old transaction that is completed and now merchant want to refund?

Helper/Order.php
```diff
@@ -701,14 +688,60 @@ public function formatReferenceUrl($reference)
...
+        }
+
+        if (in_array($prevTransactionState, [self::TS_PENDING, self::TS_REJECTED_REVERSIBLE]) &&
+            $transactionState == self::TS_AUTHORIZED &&
+            $transaction->captures
+        ) {
+            return self::TS_AUTHORIZED;
```
daisy1754: comment above says `Represent Bolt partial capture AUTHORIZED state as CAPTURED` but here you use `authorized`?

Helper/Order.php
```diff
@@ -701,14 +688,60 @@ public function formatReferenceUrl($reference)
...
+            return self::TS_CAPTURED;
+        }
+
+        if ($transactionState == self::TS_AUTHORIZED &&
+            $transaction->captures
+        ) {
+            return self::TS_CAPTURED;
+        }
+
+        if ($prevTransactionState == self::TS_CREDIT_COMPLETED &&
```
daisy1754: how is it possible?

Helper/Order.php
```diff
@@ -701,14 +688,60 @@ public function formatReferenceUrl($reference)
...
+        }
+
+        if (in_array($prevTransactionState, [self::TS_PENDING, self::TS_REJECTED_REVERSIBLE]) &&
+            $transactionState == self::TS_AUTHORIZED &&
+            $transaction->captures
+        ) {
+            return self::TS_AUTHORIZED;
+        }
+
+        if (in_array($prevTransactionState, [self::TS_PENDING, self::TS_REJECTED_REVERSIBLE]) &&
```
daisy1754: can you add short comment what is expected situation here?

Helper/Order.php
```diff
@@ -726,37 +759,20 @@ private function validateTransition($prevTransactionState, $newTransactionState)
     /**
      * Record total amount mismatch between magento and bolt order.
      * Log the error in order comments and report via bugsnag.
+     * Put the order ON HOLD if it's a mismatch.
      *
      * @param OrderModel $order
      * @param \stdClass $transaction
      * @return void
```
daisy1754: update doc

Helper/Order.php
```diff
@@ -726,37 +759,20 @@ private function validateTransition($prevTransactionState, $newTransactionState)
...
      */
-    private function checkTotalsMismatch($order, $transaction)
+    private function isTotalsMismatch($order, $transaction)
```
daisy1754: given this method has side effect, I prefer calling it `checkTotalsMismatch`

Helper/Order.php
```diff
@@ -796,77 +812,132 @@ private function checkTotalsMismatch($order, $transaction)
...
+    {
+        return [
+            'Time' => $this->timezone->formatDateTime(
+                date('Y-m-d H:i:s', $transaction->date / 1000),
+                2,
+                2
+            ),
+            'Reference' => $transaction->reference,
+            'Amount'    => $order->getBaseCurrency()->formatTxt($amount / 100),
+            'Real ID'   => $transaction->id,
```
daisy1754: Let's call it `ID` or `Transaction ID`

Helper/Order.php
```diff
@@ -796,77 +812,132 @@ private function checkTotalsMismatch($order, $transaction)
...
+    }
+
+    /**
+     * Return the first unprocessed capture from the captures array (or null)
+     *
+     * @param OrderPaymentInterface $payment
+     * @param \stdClass $transaction
+     * @return mixed
+     */
+    private function getNewCapture($payment, $transaction)
```
daisy1754: `getUnprocessedCapture`?  I don't think it's not necessarily chronological order

Helper/Order.php
```diff
@@ -796,77 +812,132 @@ private function checkTotalsMismatch($order, $transaction)
         $this->bugsnag->registerCallback(function ($report) use ($transaction, $cart) {
```
daisy1754: do we already track Bolt transaction reference and magento order ID here? If not can you add?

If hook failed due to on-hold order, first thing oncall would do is to search bugsnag with reference

Helper/Order.php
```diff
@@ -1170,44 +1180,7 @@ protected function validateCaptureAmount(OrderInterface $order, $captureAmount)
         $isInvalidAmountRange = $order->getTotalInvoiced() + $captureAmount > $order->getGrandTotal();
 
         if($isInvalidAmount || $isInvalidAmountRange) {
-            throw new Exception( __('Capture amount is invalid'));
+            throw new \Exception( __('Capture amount is invalid'));
```
daisy1754: is slash needed?

Model/Payment.php
```diff
@@ -335,8 +351,7 @@ public function capture(InfoInterface $payment, $amount)
             $capturedData = [
                 'transaction_id' => $realTransactionId,
                 'amount'         => $captureAmount,
-                'currency'       => $order->getOrderCurrencyCode(),
-                'skip_hook_notification' => true
```
daisy1754: why removing?

Helper/Order.php
```diff
@@ -881,80 +952,83 @@ public function updateOrderPayment($order, $transaction = null, $reference = nul
 
         // preset default payment / transaction values
         // before more specific changes below
-        $amount = $transaction->amount->amount;
+        $amount = @$newCapture->amount->amount ?: $transaction->amount->amount;
```
mccottry: It seems like to me that this code at times will be trying to reference data properties from null.  I believe `$newCapture` can be null.  Therefore, you have `null->amount->amount`.  While `@` will silence this error, I feel like it quietly halts the script and is a big error.  

What am I missing here?

Helper/Order.php
mccottry: In this name space `namespace Bolt\Boltpay\Helper;` and the removal of the previous `use` on line 46, the slash is needed to denote top level namespace.  This style is the accepted norm in OO PHP.

Helper/Order.php
If, for some reason, event handler would fail from the frontend call, e.g. totals mismatch, the error would propagate to the end user, breaking the checkout flow, showing a message, etc. It is not necessary, and affects both user and merchant experience with Bolt.
This way, If the payment was successful, the order is saved with the default, pending, status, it is shown in customer dashboard if opened, browser gets redirected to the success page as expected and  the payment update is postponed until the first hook is received (30 seconds?).
If the hook fails due to mismatch, it will fail early, on the first pending hook request, and we will be the first to know. The order will be put ON HOLD with a total mismatch message recorded.

Helper/Order.php
I have just refunded some transactions completed before this code update. And a few even older ones, from about 2 moths ago. Probably it would fail on some real old ones, or created with old plugin version. We should deal with them manually if / when it happens.

Helper/Order.php
This is the case when, for some reason, the base AUTHORIZED state was not recorded, and there has already been a partial capture. We first record the AUTHORIZED event and then the CAPTURE on the next hook.

Helper/Order.php
This is where the transaction states transition logic gets complicated. It is not as linear as it was. Consider this flow:
1. PENDING
2. AUTHORIZED
3. partially CAPTURED (<- Magento state), invoice created,  Bolt transaction status is still AUTHORIZED
4. partially or fully refunded previous invoice, CREDIT_COMPLETED state saved to payment in magento, Bolt transaction state is still AUTHORIZED
5. partially CAPTURED again (THIS CASE, which sends us back to #3) or fully CAPTURED (different case, COMPLETED status on both Magento and Bolt), new invoice gets created...

Helper/Order.php
danac-gs: done

Helper/Order.php
Renamed to getUnprocessedCapture. It seems the captures are stored in chronological order, but it does not matter since the record is kept of previously processed ones to avoid duplicates.

Model/Payment.php
danac-gs: @daisy1754 
The purpose of this was to avoid duplicate processing, I guess. And it's a sort of a hack. It can lead to an error if the connection breaks after the capture is done on Bolt and before the response is received in Magento. Now the record of each capture (its ID) is kept within the payment instance after it was processed. No duplicates possible.

Helper/Order.php
danac-gs: @mccottry 
The only case when the amount to process (used at several places after this point in the code ) is not the one stored in $transaction->amount->amount is when the transaction is of type (partial) CAPTURE. $newCapture is an object (processing capture case) or null (other cases) previously determined. So if there is no capture to process the error is silenced and $transaction->amount->amount used.
The code is too abbreviated, thus confusing, changing it to:
`
if ($newCapture) {
            $amount = $newCapture->amount->amount;
        } else {
            $amount = $transaction->amount->amount;
        }
`

mccottry: @danac-gs I've learned that `@null->amount->amount ?:` does not throw a run time error and will evaluate to false.  Interesting  Thanks!

Helper/Order.php
```diff
@@ -701,14 +687,78 @@ public function formatReferenceUrl($reference)
...
+        // TS_COMPLETED. If there is more than one capture in $transaction->captures array then the TS_AUTHORIZED state
+        // is missing. Set the state to TS_AUTHORIZED and process captures on next hook requests.
+        if (in_array($prevTransactionState, [self::TS_PENDING, self::TS_REJECTED_REVERSIBLE]) &&
+            $transactionState == self::TS_COMPLETED &&
+            count($transaction->captures) > 1
+        ) {
+            return self::TS_AUTHORIZED;
+        }
+
+        // The treansaction was in TS_AUTHORIZED state, now it's TS_COMPLETED but not all partial captures are
```
mccottry: "transaction" not "treansaction"

Helper/Order.php
```diff
@@ -701,14 +687,78 @@ public function formatReferenceUrl($reference)
...
+        // processed. Mark it TS_CAPTURED.
+        if ($prevTransactionState == self::TS_AUTHORIZED &&
+            $transactionState == self::TS_COMPLETED &&
+            count($transaction->captures) - count($prevCaptures) > 1
+        ) {
+            return self::TS_CAPTURED;
+        }
+
+        // The treansaction was TS_AUTHORIZED, now it has captures, put it in TS_CAPTURED state.
```
mccottry: spelling: transaction

PR #169, opened 2018-12-13
force approve/reject failed hook fix

danac-gs:


Helper/Order.php
```diff
@@ -955,12 +963,6 @@ public function updateOrderPayment($order, $transaction = null, $reference = nul
...
         }
 
-        if ($order->getState() == OrderModel::STATE_HOLDED) {
-            throw new LocalizedException(__(
-                'Order is in ON HOLD state.'
-            ));
-        }
-
         // Check for total amount mismatch between magento and bolt order.
         if ($this->holdOnTotalsMismatch($order, $transaction)) {
```
daisy1754: hmm not sure if I understand the fix. Doesn't this method still throws if order is on-hold?

It throws but only if there is a mismatch and order has already been put on hold. The mismatch check is the first thing that is done and if there is a mismatch the (first, pending) hook fails, the comment is added to order and logged in bugsnag. Nn subsequent hooks, still the mismatch, we just throw en error (bugsnag recorded), nothing else. 

If there's no mismatch order can be put on hold on reversible reject, but then the flow does not go this way, so no error thrown and execution broken. On forced approval or rejection the hooks pass and the order status gets updated accordingly.

I should update 'Order is in ON HOLD state.' to something more meaningful.

Helper/Order.php
```diff
@@ -817,6 +817,14 @@ private function holdOnTotalsMismatch($order, $transaction)
             return false;
         }
 
+        // Order is already put ON HOLD, this is a subsequent hook call
+        // No need to add mismatch comment again
+        if ($order->getState() == OrderModel::STATE_HOLDED) {
+            throw new LocalizedException(__(
+                'Order is in ON HOLD state.'
```
daisy1754: can you use comment below as error message here?

```
        $comment = __(
            'BOLTPAY INFO :: THERE IS A MISMATCH IN THE ORDER PAID AND ORDER RECORDED.<br>
             Paid amount: %1 Recorded amount: %2<br>Bolt transaction: %3',
            $boltTotal / 100,
            $order->getGrandTotal(),
            $this->formatReferenceUrl($transaction->reference)
        );
```

PR #173, opened 2018-12-19
Amasty gift card

danac-gs:


Helper/Config.php
```diff
@@ -735,4 +735,20 @@ public function getToggleCheckout()
     {
         return $this->getAdditionalConfigProperty('toggleCheckout');
     }
+
+    /**
+     * Get Bolt additional configuration for Amasty Gift Card support, stored in the following format:
+     *
+     * {
+     *   "amastyGiftCard": {
+           "payForEverything": true|false   // true, the default,
```
matt-thomason: It looks like you're missing a `*` here to make this line part of the block comment.

Helper/Discount.php
```diff
@@ -0,0 +1,320 @@
...
+     * @param int|string $immutableQuoteId
+     */
+    public function cloneAmastyGiftCards($parentQuoteId, $immutableQuoteId)
+    {
+        try {
+            $connection = $this->resource->getConnection();
+            $giftCardTable = $this->resource->getTableName('amasty_amgiftcard_quote');
+
+            // Clear previously applied gift cart codes from the immutable quote
+            $sql = "DELETE FROM {$giftCardTable} WHERE quote_id = :quote_id";
```
matt-thomason: These should be in some sort of transaction, so we can roll back in case of disaster. What if we do a successful `DELETE` but then our connection is interrupted before we do our `INSERT`? That applies to most of the database interactions throughout.

Helper/Discount.php
```diff
@@ -0,0 +1,320 @@
...
+
+    /**
+     * Get Amasty Gift Card codes, stored comma separated in total title field, return as array.
+     *
+     * @param array $totals totals array collected from quote
+     * @return array
+     */
+    public function getAmastyGiftCardCodesFromTotals($totals) {
+        $giftCardCodes = explode(',', $totals[self::AMASTY_GIFTCARD]->getTitle());
+        return array_filter(array_map('trim', $giftCardCodes));
```
matt-thomason: The filter doesn't seem necessary here? Do we typically get whitespace in our gift card code csv?

Helper/Discount.php
```diff
@@ -0,0 +1,320 @@
...
+     * Load Amasty Gift Card account object.
+     * @param string $code Gift Card coupon code
+     * @return \Amasty\GiftCard\Model\Account|null
+     */
+    public function loadAmastyGiftCard($code)
+    {
+        try {
+            $accountModel = $this->amastyAccountFactory->getInstance()
+                ->create()
+                ->loadByCode($code);
```
YevhenShBolt: Does Amasty have something like Repository instead of getInstance()->createe()->loadByCode()?

Helper/Discount.php
```diff
@@ -0,0 +1,320 @@
...
+
+        $isValid = $this->amastyGiftCardManagement->getInstance()->validateCode($quote, $code);
+
+        if (!$isValid) {
+            throw new LocalizedException(__('Coupon with specified code "%1" is not valid.', $code));
+        }
+
+        if ($accountModel->canApplyCardForQuote($quote)) {
+            $quoteGiftCard = $this->amastyQuoteFactory->getInstance()->create();
+            $this->amastyQuoteResource->getInstance()->load($quoteGiftCard, $quoteId, 'quote_id');
```
YevhenShBolt: Can you clarify what exactly load here?
Maybe I'm wrong (i do not have current module underhand) but it seems to me that it is extra load/prepare for/from Amasty module.

Helper/Discount.php
```diff
@@ -0,0 +1,320 @@
...
+            $sql = "DELETE FROM {$giftCardTable} WHERE quote_id = :quote_id";
+            $connection->query($sql, ['quote_id' => $immutableQuoteId]);
+
+            // Copy all gift cart codes applied to the parent quote to the immutable quote
+            $sql = "INSERT INTO {$giftCardTable} (quote_id, code_id, account_id, base_gift_amount, code) 
+                    SELECT :immutable_quote_id, code_id, account_id, base_gift_amount, code 
+                    FROM {$giftCardTable} WHERE quote_id = :quote_id";
+            $connection->query($sql, ['immutable_quote_id' => $immutableQuoteId, 'quote_id' => $parentQuoteId]);
+        } catch (\Zend_Db_Statement_Exception $e) {
+            // The amasty_amgiftcard_quote table is missing, Amasty Gift Cart is not installed,
```
daisy1754: can we check if amasty extension is there beginning of this function + early return, and here we track exception with bugsnag?

Helper/Discount.php
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        if ($accountModel->canApplyCardForQuote($quote)) {
+            $quoteGiftCard = $this->amastyQuoteFactory->getInstance()->create();
+            $this->amastyQuoteResource->getInstance()->load($quoteGiftCard, $quoteId, 'quote_id');
+            $subtotal = $quoteGiftCard->getSubtotal($quote);
+
+            if ($quoteGiftCard->getCodeId() && $accountModel->getCodeId() == $quoteGiftCard->getCodeId()) {
+
+                throw new LocalizedException(__('This gift card account is already in the quote.'));
```
daisy1754: if same code is applied, can we return the same discount information? It's possible bolt send the same request multiple times

danac-gs: @daisy1754 sending the same request multiple times is taken care of in DiscountCodeValidation class by simply removing the discount (deleting a record from the db table) and applying it again. So the same data is returned.

Helper/Discount.php
danac-gs: @YevhenShBolt the loaded object is of Amasty\GiftCard\Model\Quote class. It implements QuoteInterface and also stores applied Gift Card related data.

Helper/Discount.php
danac-gs: @YevhenShBolt It uses Factories and Resource Models. getInstance() is our ThirdPartyModuleFactory call for (trying) instantiating third party module objects that might not be present (installed). I think it's your (great) solution. The rest is more or less copied from the module, apply controller in this particular case. That is how it's done in the Amasty module.

Helper/Discount.php
danac-gs: @daisy1754  Sure.

Helper/Config.php
danac-gs: @matt-thomason Fixed

Helper/Discount.php
danac-gs: @matt-thomason removed

Helper/Discount.php
matt-thomason: @danac-gs Any plans for wrapping this in a transaction? Or is this going to be called somewhere else where we can roll it back?

Helper/Discount.php
```diff
@@ -0,0 +1,352 @@
...
+    /**
+     * Get accumulated unused balance of all Amasty Gift Cards corresponding to passed gift card coupons array
+     *
+     * @param array $giftCardCodes
+     * @return float|int|mixed
+     */
+    public function getAmastyGiftCardCodesCurrentValue($giftCardCodes) {
+        $amount = 0;
+        foreach ($giftCardCodes as $giftCardCode) {
+            $giftCard = $this->loadAmastyGiftCard($giftCardCode);
```
daisy1754: loadAmastyGiftCard can return null (per type definition). need null check?

Helper/Discount.php
danac-gs: @matt-thomason It is wrapped. Not sure why you don't see it. Here is the link to the file:
https://github.com/BoltApp/bolt-magento2/blob/f7949f38f77cc3c9a5160cd7b3ec50363aa3b896/Helper/Discount.php#L232

matt-thomason: huh I'm crazy

Helper/Discount.php
johnmccombs1: This seems to be running a lot of queries to get each gift card value. Would something like this work better?
`$giftCardCollection->addFieldToFilter('current_value', ['in'=>$giftCardCodes])->getData();`
And then summarize the current_value on the returned array.

Helper/Discount.php
YevhenShBolt: Thank you, now i see.

Helper/Discount.php
danac-gs: Thanks @johnmccombs1. Updated.

PR #177, opened 2018-12-21
set order status refactor, fix for no status after unhold

danac-gs:


danac-gs:
@mccottry 
In M2 you can hold and unhold order (without facing no state issue) if the order was in either pending (new) or processing state. The order is always created with pending (new) state and the first thing checked on a first hook is totals mismatch. If thgere's a mismatch the hook fails and (pending) order is put on hold. Can be unholded. In other case (reversibly rejected) the previous state is forced to processing so unhold can also work.

PR #186, opened 2019-01-11
hook context overriding oauth context fix

danac-gs:
This needs to be deployed to Taos as a Returnly (and others) oauth authorization fix.

### DIFF
--- a/etc/webapi_rest/di.xml
+++ b/etc/webapi_rest/di.xml
@@ -22,7 +22,7 @@
             <argument name="userContexts" xsi:type="array">
                 <item name="hookContext" xsi:type="array">
                     <item name="type" xsi:type="object">Bolt\Boltpay\Model\Authorization\HookContext</item>
-                    <item name="sortOrder" xsi:type="string">40</item>
+                    <item name="sortOrder" xsi:type="string">90</item>
                 </item>
             </argument>
         </arguments>


danac-gs:
@daisy1754 By investigating the code I found out that multiple user contexts exists and are passed as an array to the CompositeUserContext constructor. Then, walking through that array by sortOrder, the first context that match is chosen. The default context array for REST API is defined as follows:
```xml
    <type name="Magento\Authorization\Model\CompositeUserContext">
        <arguments>
            <argument name="userContexts" xsi:type="array">
                <item name="tokenUserContext" xsi:type="array">
                    <item name="type" xsi:type="object">Magento\Webapi\Model\Authorization\TokenUserContext</item>
                    <item name="sortOrder" xsi:type="string">10</item>
                </item>
                <item name="oauthUserContext" xsi:type="array">
                    <item name="type" xsi:type="object">Magento\Webapi\Model\Authorization\OauthUserContext</item>
                    <item name="sortOrder" xsi:type="string">40</item>
                </item>
                <item name="guestUserContext" xsi:type="array">
                    <item name="type" xsi:type="object">Magento\Webapi\Model\Authorization\GuestUserContext</item>
                    <item name="sortOrder" xsi:type="string">100</item>
                </item>
            </argument>
        </arguments>
    </type>
``` 
Coincidently, our hookContext was defined with sortOrder 40 and had overridden or put in place before OauthUserContext. Since I did not find any other context definition in the code base with sortOrder between 40 and 100 I picked a random number between those two. This might need an adjustment on some systems, though, if custom contexts like ours exist. I was not sure should I move it more towards 40 or 100.

PR #185, opened 2019-01-10
Allow empty email in api (ApplePay)

YevhenShBolt:
Task: https://app.asana.com/0/941920570700290/962234572439827/f

PR #188, opened 2019-01-14
Unit test for getShippingMethods() method

YevhenShBolt:
Task - "Unit test for Apple Pay shipping&tax"
https://app.asana.com/0/941920570700290/962234572439827/f

### DIFF
--- a/Model/Api/ShippingMethods.php
+++ b/Model/Api/ShippingMethods.php
@@ -495,17 +495,30 @@ class ShippingMethods implements ShippingMethodsInterface
 
         $shippingMethods = $this->getShippingOptions($quote, $addressData);
 
-        $shippingOptionsModel = $this->shippingOptionsInterfaceFactory->create();
-        $shippingOptionsModel->setShippingOptions($shippingMethods);
-
-        $shippingTaxModel = $this->shippingTaxInterfaceFactory->create();
-        $shippingTaxModel->setAmount(0);
-        $shippingOptionsModel->setTaxResult($shippingTaxModel);
+        $shippingOptionsModel = $this->getShippingOptionsData($shippingMethods);
 
         // Cache the calculated result
         if ($prefetchShipping) {
             $this->cache->save(serialize($shippingOptionsModel), $cacheIdentifier, [], 3600);
         }
+
+        return $shippingOptionsModel;
+    }
+
+    /**
+     * Set shipping methods to the ShippingOptions object
+     *
+     * @param $shippingMethods
+     */
+    protected function getShippingOptionsData($shippingMethods)
+    {
+        $shippingOptionsModel = $this->shippingOptionsInterfaceFactory->create();
+
+        $shippingTaxModel = $this->shippingTaxInterfaceFactory->create();
+        $shippingTaxModel->setAmount(0);
+
+        $shippingOptionsModel->setShippingOptions($shippingMethods);
+        $shippingOptionsModel->setTaxResult($shippingTaxModel);
 
         return $shippingOptionsModel;
     }
@@ -540,7 +553,7 @@ class ShippingMethods implements ShippingMethodsInterface
      *
      * @return ShippingOptionInterface[]
      */
-    private function getShippingOptions($quote, $addressData)
+    public function getShippingOptions($quote, $addressData)
     {
         if ($quote->isVirtual()) {
             $billingAddress = $quote->getBillingAddress();
--- a/Model/Api/ShippingMethods.php
+++ b/Model/Api/ShippingMethods.php
@@ -341,7 +341,8 @@ class ShippingMethods implements ShippingMethodsInterface
         } catch (BoltException $e) {
             $this->catchExceptionAndSendError($e, $e->getMessage(), $e->getCode());
         } catch (\Exception $e) {
-            $this->catchExceptionAndSendError($e, '', 6009, 422);
+            $msg = __('Unprocessable Entity') . ': ' . $e->getMessage();
+            $this->catchExceptionAndSendError($e, $msg, 6009, 422);
         }
     }
 
@@ -354,9 +355,6 @@ class ShippingMethods implements ShippingMethodsInterface
     protected function catchExceptionAndSendError($e, $msg = '', $code = 6009, $httpStatusCode = 422)
     {
         $this->bugsnag->notifyException($e);
-        if (!$msg) {
-            $msg = __('Unprocessable Entity') . ': ' . $e->getMessage();
-        }
 
         $this->sendErrorResponse($code, $msg, $httpStatusCode);
     }
--- a/Model/Api/ShippingMethods.php
+++ b/Model/Api/ShippingMethods.php
@@ -347,14 +347,14 @@ class ShippingMethods implements ShippingMethodsInterface
     }
 
     /**
-     * @param        $e
+     * @param        $exception
      * @param string $msg
      * @param int    $code
      * @param int    $httpStatusCode
      */
-    protected function catchExceptionAndSendError($e, $msg = '', $code = 6009, $httpStatusCode = 422)
+    protected function catchExceptionAndSendError($exception, $msg = '', $code = 6009, $httpStatusCode = 422)
     {
-        $this->bugsnag->notifyException($e);
+        $this->bugsnag->notifyException($exception);
 
         $this->sendErrorResponse($code, $msg, $httpStatusCode);
     }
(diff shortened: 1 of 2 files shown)


Model/Api/ShippingMethods.php
```diff
@@ -342,16 +337,70 @@ public function getShippingMethods($cart, $shipping_address)
...
-            $this->bugsnag->notifyException($e);
+            $this->catchExceptionAndSendError($e, '', 6009, 422);
+        }
+    }
+
+    /**
+     * @param        $e
+     * @param string $msg
+     * @param int    $code
+     * @param int    $httpStatusCode
```
daisy1754: no need for this param?

Model/Api/ShippingMethods.php
```diff
@@ -342,16 +337,70 @@ public function getShippingMethods($cart, $shipping_address)
...
+     */
+    protected function catchExceptionAndSendError($e, $msg = '', $code = 6009, $httpStatusCode = 422)
+    {
+        $this->bugsnag->notifyException($e);
+        if (!$msg) {
             $msg = __('Unprocessable Entity') . ': ' . $e->getMessage();
```
daisy1754: move this outside of this method.

Model/Api/ShippingMethods.php
```diff
@@ -342,16 +337,70 @@ public function getShippingMethods($cart, $shipping_address)
...
+    public function getQuoteById($quoteId)
+    {
+        return $this->cartHelper->getQuoteById($quoteId);
+    }
+
+    /**
+     * @throws LocalizedException
+     * @throws \Magento\Framework\Webapi\Exception
+     */
+    protected function proceedWithHook()
```
daisy1754: how about preprocessHook ?

Test/Unit/Model/Api/ShippingMethodsTest.php
```diff
@@ -0,0 +1,322 @@
...
+ *
+ * @package Bolt\Boltpay\Test\Unit\Model\Api
+ */
+class ShippingMethodsTest extends TestCase
+{
+    /**
+     * @var BoltShippingMethods
+     */
+    private $currentMock;
+
```
daisy1754: nit - Is having newline between each variable PHP standard thing? I personally find it easier to read if there is no line breaks in between

Test/Unit/Model/Api/ShippingMethodsTest.php
```diff
@@ -0,0 +1,322 @@
...
+        $this->bugsnag = $this->getMockBuilder(Bugsnag::class)
+            ->setMethods(['notifyException'])
+            ->disableOriginalConstructor()
+            ->getMock();
+        $this->bugsnag->method('notifyException')
+            ->willReturnSelf();
+    }
+
+    public function testGetShippingMethodsIfQuoteEmpty()
```
daisy1754: it's better if test can also say expected outcome - eg., `testGetShippingMethodsRetrunsErrorIfQuoteEmpty`

Test/Unit/Model/Api/ShippingMethodsTest.php
YevhenShBolt: Please, looking on core code style, as an example:
vendor/magento/module-catalog/Test/Unit/Model/LayerTest.php

In a current PSR, i could not find any info about spaces between such variables.

Model/Api/ShippingMethods.php
YevhenShBolt: renamed

Model/Api/ShippingMethods.php
YevhenShBolt: it is needed. The `httpStatusCode` parameter is passed here, a bit above for `Webapi\Exception`

Test/Unit/Model/Api/ShippingMethodsTest.php
matt-thomason: PSR-2 does not offer a recommendation on this. If this is the current style across the rest of the codebase, then it's probably worth keeping it to maintain consistency.

Test/Unit/Model/Api/ShippingMethodsTest.php
```diff
@@ -0,0 +1,494 @@
...
+
+        $this->bugsnag = $this->getMockBuilder(Bugsnag::class)
+            ->setMethods(['notifyException'])
+            ->disableOriginalConstructor()
+            ->getMock();
+        $this->bugsnag->method('notifyException')
+            ->willReturnSelf();
+    }
+
+    public function testGetShippingMethodsIfQuoteEmptyReturnException()
```
matt-thomason: This is a style choice, but I'm a big fan of using PHPUnit's annotations. You can tag methods with `@test` so you don't need to prefix each test function, and here you can use `@expectedException` and `@expectedExceptionMessage` for assertions, making the purpose of this test really clear without having to scan the body.

Test/Unit/Model/Api/ShippingMethodsTest.php
```diff
@@ -0,0 +1,494 @@
...
+            ->getMock();
+
+        $message = new Phrase('Unprocessable Entity: Unknown quote id: ' . $quoteId);
+        $this->currentMock->method('throwUnknownQuoteIdException')
+            ->willThrowException(new LocalizedException($message));
+        $this->currentMock->method('catchExceptionAndSendError')
+            ->withAnyParameters()
+            ->willThrowException(new LocalizedException($message));
+
+        try {
```
matt-thomason: This is not the recommended way of testing exceptions in PHPUnit. You shouldn't have a `try-catch` block - what if your mock somehow doesn't throw an exception?
Let the exception be thrown and assert with either an annotation or `$this->expectException`/`$this->expectExceptionMessage()`.

Test/Unit/Model/Api/ShippingMethodsTest.php
YevhenShBolt: I think so, we should keep it.

Test/Unit/Model/Api/ShippingMethodsTest.php
YevhenShBolt: gotcha, the code was refactored.

Test/Unit/Model/Api/ShippingMethodsTest.php
YevhenShBolt: uppated

Test/Unit/Model/Api/ShippingMethodsTest.php
```diff
@@ -173,7 +173,12 @@ public function setUp()
             ->willReturnSelf();
     }
 
-    public function testGetShippingMethodsIfQuoteEmptyReturnException()
+    /**
+     * @test
+     * @expectedException
```
matt-thomason: You don't need both `@expectedException` and `$this->expectException`.
Perhaps better to just go with the `$this->expectException` format, since it's implemented correctly here (the annotations need to expected values)

Test/Unit/Model/Api/ShippingMethodsTest.php
```diff
@@ -0,0 +1,504 @@
...
+        $option
+            ->setService('Flat Rate - Fixed')
+            ->setCost(5600)
+            ->setReference('flatrate_flatrate')
+            ->setTaxAmount(0)
+        ;
+
+        $shippingOptionData = [$option];
+
+        $this->currentMock->method('shippingEstimation')
```
daisy1754: can you mock one-level deeper? Idea of adding unit test is to make sure our logic is working for apple pay. If someone revert https://github.com/BoltApp/bolt-magento2/pull/185 unit test should fail. However current test would pass because we mock shippingEstimation

Test/Unit/Model/Api/ShippingMethodsTest.php
```diff
@@ -0,0 +1,658 @@
...
+            $this->request,
+            $this->cache,
+            $this->priceHelper,
+            $this->sessionHelper
+        );
+
+        $result = $testClass->shippingEstimation($quote, $shippingAddressData);
+
+        $this->assertEquals($this->factoryShippingOptionsMock, $result);
+        $this->assertEquals($this->shippingTaxInterfaceFactory, $result->getTaxResult());
```
daisy1754: why you need to assert `$result->getTaxResult()` when you already assert `$result` above?

YevhenShBolt: removed.

PR #203, opened 2019-01-29
fix login prompt for restricted guest checkout on non cart pages

danac-gs:


danac-gs:
@daisy1754 can you rerun circleci jobs?

matt-thomason:
Rerunning circleci

PR #206, opened 2019-01-30
Unirgy_Giftcert plugin support

YevhenShBolt:
The Unirgy_Giftcert add a discount which covers only product price (subtotal).
The Unirgy_Giftcert accumulate correct balance only after address collectTotals, but if i am trying to get  Unirgy_Giftcert balance at shipping_methods step i have no correct price value - i have only product price discount.
We should get the whole balance at first of the Unirgy_Giftcert.

P.S.: Unirgy_Giftcert - also want to mention that some part of the module is encoded (the part like "add the Giftcert to the quote/cart" and some save "after update quote" event)

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -954,6 +954,20 @@ class Cart extends AbstractHelper
                 }
                 ///////////////////////////////////////////////////////////////////////////
 
+                ///////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


YevhenShBolt:
- Added description
- Removed an extra code because Bolt automatically calculates a discount.

YevhenShBolt:
I created another task for PHPUnit Test. I plan do separate PR for it.

PR #211, opened 2019-02-04
Reserve Order ID for the child quote, copy it to parent just before the quote to order submit

danac-gs:
Some 3rd party plugins read Quote::getReservedOrderId() value and if set consider the quote to be a complete order. It is set by calling Quote::reserveOrderId(). We used to call it on the parent quote before creating the immutable quote. Now it is called on the first created immutable quote and stored in the separate field added to the quote model to be set on the parent quote just before order creation.

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -533,7 +533,7 @@ class Cart extends AbstractHelper
      * Defer setting ReservedOrderId on the parent quote
      * until the quote to order submission.
      * Reason: Some 3rd party plugins read this value and if set
-     * consider the quote to bo a complete order.
+     * consider the quote to be a complete order.
      *
      * @param Quote $immutableQuote
      * @param Quote $quote
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -480,7 +480,7 @@ class Cart extends AbstractHelper
         $child,
         $save = true,
         $emailFields = ['customer_email', 'email'],
-        $excludeFields = ['entity_id', 'address_id']
+        $excludeFields = ['entity_id', 'address_id', 'reserved_order_id']
     ) {
         foreach ($parent->getData() as $key => $value) {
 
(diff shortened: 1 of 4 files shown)


Setup/UpgradeSchema.php
```diff
@@ -46,6 +46,17 @@ public function upgrade(SchemaSetupInterface $setup, ModuleContextInterface $con
             ]
         );
 
+        $setup->getConnection()->addColumn(
```
daisy1754: is reservedOrderId field indexed? if so maybe we should index this too

Helper/Cart.php
```diff
@@ -526,6 +526,56 @@ public function replicateQuoteData($source, $destination)
...
+    protected function reserveOrderId($immutableQuote, $quote)
+    {
+        $reservedOrderId = $immutableQuote->getBoltReservedOrderId();
+        if (!$reservedOrderId) {
+            $reservedOrderId = $immutableQuote->reserveOrderId()->getReservedOrderId();
+            $immutableQuote->setBoltReservedOrderId($reservedOrderId);
+            $quote->setBoltReservedOrderId($reservedOrderId);
+            $this->quoteResourceSave($quote);
+        } else {
+            $immutableQuote->setReservedOrderId($reservedOrderId);
```
daisy1754: I'm not sure if I follow here. what is this else block for?

Setup/UpgradeSchema.php
danac-gs: Index is not required we just do set and get on a loaded model, no searches or queries of any kind.

Helper/Cart.php
danac-gs: reserveOrderId() is called on the first immutable quote and then propagated to parent and to subsequent immutables in BoltReservedOrderId field and in the end copied from parent BoltReservedOrderId to ReservedOrderId. This else is for consistency, to keep every subsequent immutable having all the fields populated. Actually, one more important reason, some lines below $cart['display_id'] = $immutableQuote->getReservedOrderId() . ' / ' . $immutableQuote->getId(); It is expected to be there. So every call to getReservedOrderId remain unchanged.

PR #239, opened 2019-03-08
Allow Admin to update order manually

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/1113207840437480

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -1127,6 +1127,9 @@ class Order extends AbstractHelper
             $reference == $prevTransactionReference &&
             !$newCapture
         ) {
+            if ($this->isAnAllowedUpdateFromAdminPanel($order, $transactionState)){
+                $payment->setIsTransactionApproved(true);
+            }
             return;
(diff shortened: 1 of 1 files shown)


Helper/Order.php
```diff
@@ -1388,4 +1391,16 @@ protected function validateCaptureAmount(OrderInterface $order, $captureAmount)
             throw new \Exception( __('Capture amount is invalid'));
         }
     }
+
+    /**
+     * @param OrderInterface $order
+     * @param $transactionState
+     * @return bool
+     */
+    protected function getPaymentUpdateFromAdminPanel(OrderInterface $order, $transactionState)
```
mccottry: While this method is doing a little more than what the name is advertising by checking order state and transaction state, and generally, I would suggest that a method returning a boolean by convention would have the verb "to be" usually as "is" somewhere in it (e.g. `isAnAllowedUpdateFromAdminPanel`), the code looks functionally good to me.

PR #241, opened 2019-03-13
Add multistore support

YevhenShBolt:
Task - https://app.asana.com/0/0/1113210174697856/f

### DIFF
--- a/Block/Checkout/Cart/ComponentSwitcherProcessor.php
+++ b/Block/Checkout/Cart/ComponentSwitcherProcessor.php
@@ -45,7 +45,7 @@ class ComponentSwitcherProcessor implements LayoutProcessorInterface
     public function process($jsLayout) {
         // Store Credit
         $jsLayout['components']['block-totals']['children']['storeCredit']['componentDisabled'] =
-            ! $this->configHelper->getStoreCreditConfig();
+            ! $this->configHelper->useStoreCreditConfig();
 
         return $jsLayout;
     }
--- a/Block/Form.php
+++ b/Block/Form.php
@@ -3,7 +3,7 @@
 namespace Bolt\Boltpay\Block;
 
 use Bolt\Boltpay\Helper\Config;
-use Magento\Backend\Model\Session\Quote as BackendQuote;
+use Magento\Framework\Session\SessionManagerInterface as MagentoQuote;
 use Magento\Payment\Block\Form as PaymentForm;
 use Magento\Framework\View\Element\Template\Context;
 
@@ -12,6 +12,9 @@ use Magento\Framework\View\Element\Template\Context;
  */
 class Form extends PaymentForm
 {
+    /**
+     * @var \Magento\Backend\Model\Session\Quote
+     */
     private $_quote;
 
     /**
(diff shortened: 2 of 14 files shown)


Block/Checkout/Cart/ComponentSwitcherProcessor.php
```diff
@@ -45,7 +45,7 @@ public function __construct(
     public function process($jsLayout) {
         // Store Credit
         $jsLayout['components']['block-totals']['children']['storeCredit']['componentDisabled'] =
-            ! $this->configHelper->getStoreCreditConfig();
+            ! $this->configHelper->useStoreCreditConfig();
```
danac-gs: Is $storeId parameter needed here as well?

YevhenShBolt: @danac-gs i am planning to create separate PR and ticket for frontend part.

YevhenShBolt:
@danac-gs In system.xml not needs option showInStore="1".

Look at class \Magento\Framework\App\Config::getValue. There is exist a resolver which checks the website also and if the website option is available it gets the correct value from website/store field.

Ok i will wait for  #242 PR.

I tested it on local environment only. The default functionality of our plugin and frontend part did not change. 
The current fix for one merchant only (for now) which have special requirements for admin side.
This PR fixed it.

I am planning to continue work on it. I will create 
another ticket for all parts: frontend, admin side, unit tests and of course readiness tests.

P.S. @matt-thomason  @daisy1754  what do you think?

danac-gs:
@YevhenShBolt As I understand, but I may be wrong, showInStore="1" must be set for the merchant to have options to enter different values for different stores. But I guess that you will cover in the separate PRs, admin and frontend.

Block/Checkout/Cart/ComponentSwitcherProcessor.php
YevhenShBolt: @danac-gs I checked. it works now for non multiwebsite

danac-gs:
> @YevhenShBolt As I understand, but I may be wrong, showInStore="1" must be set for the merchant to have options to enter different values for different stores. But I guess that you will cover in the separate PRs, admin and frontend.

@YevhenShBolt can you clarify this? ^

YevhenShBolt:
> @YevhenShBolt As I understand, but I may be wrong, showInStore="1" must be set for the merchant to have options to enter different values for different stores. But I guess that you will cover in the separate PRs, admin and frontend.

@danac-gs . For the first step, i already solved the separate multi-website admin configuration.
For the second step i plan to update Frontend part for whole support.

As for showInStore="1", i am not sure is it need for 100%. 
I refer to the current ticket description. Initially, it was necessary to add support for websites only, as i understand.
Now the code supports both the change for websites and for stores, we only should change to "showInStore = 1" in "adminhtml/system.xml".

@matt-thomason What do you think? Should we add this option showInStore=1 for the merchant "Trex/Polywood".

Maybe we can discuss and solve it in another PR and with the full frontend part.

matt-thomason:
@YevhenShBolt When merging this branch please don't delete this from the repo right away - I will ask Polywood/Trex to pull this branch for testing on their staging server.

PR #242, opened 2019-03-13
one step checkout support / disable bolt on payment only checkout pages

danac-gs:


etc/adminhtml/system.xml
```diff
@@ -141,6 +141,10 @@
                     <label>Use Store Credit on Shopping Cart</label>
                     <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
                 </field>
+                <field id="enable_payment_only_checkout" translate="label" type="select" sortOrder="250" showInDefault="1" showInWebsite="1" showInStore="1">
```
daisy1754: shouldn't this be inferred via presence of paymentonly publishable key?

Block/Checkout/ComponentSwitcherProcessor.php
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * Process the layout
+     *
+     * @param array $jsLayout
+     * @return array
+     */
+    public function process($jsLayout) {
+        // Payment only checkout based on Magento default one-page checkpout
+        // (one-page checkout, Mageplaza one-step checkout etc.)
+        $jsLayout['components']['checkout']['children']['steps']['children']['billing-step']['children']['payment']
```
daisy1754: is it possible any of it is undefined, like `$jsLayout['components']['checkout']['children']['steps']` is undefined?

etc/adminhtml/system.xml
danac-gs: @daisy1754 Yes but the solution was front-end based and had an error as well so we had issues with checkout that is not the default Magento one-page checkout but 3rd party (Magewrox one-step checkout on Coastal Business). This way it is removed from layout completely.

Block/Checkout/ComponentSwitcherProcessor.php
danac-gs: No, it is defined in our layout extension 
https://github.com/BoltApp/bolt-magento2/blob/one-step-checkout-support/view/frontend/layout/checkout_index_index.xml

PR #250, opened 2019-04-02
Set hints.virtual_terminal_mode=true for admin order

dumega:
asana task https://app.asana.com/0/941920570700290/1115277704594367/f

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -375,11 +375,12 @@ class Cart extends AbstractHelper
      * Get the hints data for checkout
      *
      * @param string $cartReference            (immutable) quote id
+     * @param string $checkoutType             'cart' | 'admin' Default to `admin`
      *
      * @return array
      * @throws NoSuchEntityException
      */
-    public function getHints($cartReference = null)
+    public function getHints($cartReference = null, $checkoutType = 'admin')
     {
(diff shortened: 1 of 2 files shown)


Helper/Cart.php
```diff
@@ -375,11 +375,12 @@ private function getSignResponse($signRequest)
      * Get the hints data for checkout
      *
      * @param string $cartReference            (immutable) quote id
+     * @param string $checkoutType             'cart' | 'admin' Default to `admin`
      *
      * @return array
      * @throws NoSuchEntityException
      */
-    public function getHints($cartReference = null)
+    public function getHints($cartReference = null, $checkoutType = 'admin')
```
danac-gs: Perhaps the default $checkoutType should be 'cart'

dumega: @danac-gs if the default value is `cart`, then https://github.com/BoltApp/bolt-magento2/blob/develop/Controller/Adminhtml/Cart/Data.php#L109 would be $hints = $this->cartHelper->getHints(null, 'admin');  I think there is no need to send default value for the first variable, so I choose to set the second variable default to `admin`

danac-gs: @dumega makes sense

PR #251, opened 2019-04-04
Multistore support.

YevhenShBolt:


Block/Js.php
```diff
@@ -450,7 +442,13 @@ public function shouldDisableBoltCheckout()
...
     {
-        return (int) ($this->magentoQuote && $this->magentoQuote->getStoreId()) ?
-            $this->magentoQuote->getStoreId() : 0;
+        if ($this->checkoutSession instanceof BackendSessionQuote) {
+            $quote = $this->checkoutSession;
+        } else {
+            $quote = $this->checkoutSession->getQuote();
+        }
+
+        return (int) ($quote && $quote->getStoreId()) ?
```
danac-gs: Why casting to int?

YevhenShBolt: I mistyped, I wanted to be sure that we return int type but i was mistyped and convert to int only a condition.

PR #255, opened 2019-04-10
Bump version to 1.1.12.1

matt-thomason:


etc/module.xml
```diff
@@ -18,7 +18,7 @@
 -->
 <!-- Bolt Payment Module Config. Plugin version and dependency loading sequence. -->
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Module/etc/module.xsd">
-    <module name="Bolt_Boltpay" setup_version="1.1.12">
+    <module name="Bolt_Boltpay" setup_version="1.1.12.1">
```
danac-gs: Not sire if 4-part setup_version is supported. Never tried. Could not find any resources about that. We'll have to try it. In a similar case with plugin version 1.1.8.x I just updated the composer.json (currently 1.1.8.3) and left module.xml at 1.1.8. Let's see what happens. Approving.

PR #258, opened 2019-04-12
Partial void implementation

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/1115713860206741

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -67,6 +67,7 @@ class Order extends AbstractHelper
     const TS_PENDING               = 'cc_payment:pending';
     const TS_AUTHORIZED            = 'cc_payment:authorized';
     const TS_CAPTURED              = 'cc_payment:captured';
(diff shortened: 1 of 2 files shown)


Helper/Order.php
```diff
@@ -1403,4 +1425,23 @@ protected function isAnAllowedUpdateFromAdminPanel(OrderInterface $order, $trans
...
     }
+
+    /**
+     * @param OrderPaymentInterface $payment
+     *
+     * @return \Magento\Framework\Phrase|string
+     */
+    protected function getVoidMessage(OrderPaymentInterface $payment)
+    {
+        $authorizationTransaction = $payment->getAuthorizationTransaction();
```
matt-thomason: We already got `$authorizationTransaction` in the switch statement - can we pass it into this function instead of getting it again?

mccottry: @johnmccombs1 Is the cost of `$payment->getAuthorizationTransaction();` another database lookup?

PR #262, opened 2019-04-24
Add frontend code for calling configureProductCheckout

vitaliyreznikov:
Reference https://app.asana.com/0/0/1103912883581240/f

### DIFF
--- a/Block/Js.php
+++ b/Block/Js.php
@@ -388,7 +388,7 @@ class Js extends Template
     /**
      * Return true if we are on cart page or checkout page
      */
-    public function IsCartOrCheckoutPage() {
+    public function IsOnPageFromWhiteList() {
         $currentPage = $this->getRequest()->getFullActionName();
         return in_array($currentPage, $this->getPageWhitelist());
     }
--- a/Block/Js.php
+++ b/Block/Js.php
@@ -33,7 +33,7 @@ class Js extends Template
     /**
      * @var Config
      */
-    private $configHelper;
+    protected $configHelper;
 
     /** @var CheckoutSession */
     private $checkoutSession;
@@ -320,7 +320,7 @@ class Js extends Template
      *
      * @return array
      */
-    private function getPageBlacklist()
+    protected function getPageBlacklist()
     {
         return $this->configHelper->getPageBlacklist();
     }
@@ -332,7 +332,7 @@ class Js extends Template
      *
      * @return array
      */
-    private function getPageWhitelist()
+    protected function getPageWhitelist()
     {
         return array_unique(array_merge(Config::$defaultPageWhitelist, $this->configHelper->getPageWhitelist()));
     }
@@ -345,7 +345,7 @@ class Js extends Template
      *
      * @return bool
      */
-    private function isPageRestricted()
+    private function isPageRestricted($type)
     {
         $currentPage = $this->getRequest()->getFullActionName();
 
@@ -354,13 +354,19 @@ class Js extends Template
             return true;
         }
 
-        // If minicart is supported (allowing Bolt on every page)
+        // If minicart/product page checkout is supported (allowing Bolt on every page)
         // and no IP whitelist is defined there are no additional restrictions.
-        if ($this->configHelper->getMinicartSupport() && !$this->configHelper->getIPWhitelistArray()) {
-            return false;
+        if (!$this->configHelper->getIPWhitelistArray()) {
+            if ($this->configHelper->getMinicartSupport()) {
+                return false;
+            }
+            if ($type=="withproductpage" && ($currentPage=="catalog_product_view") && $this->configHelper->getProductPageCheckoutFlag()) {
+                return false;
+            }
+
         }
 
-        // No minicart support or there is IP whitelist defined. Check if the page is whitelisted.
+        // No minicart/product page checkout support or there is IP whitelist defined. Check if the page is whitelisted.
         // If IP whitelist is defined, the Bolt checkout functionality
         // must be limited to the non cached pages, shopping cart and checkout (internal or 3rd party).
         return ! in_array($currentPage, $this->getPageWhitelist());
@@ -384,8 +390,8 @@ class Js extends Template
      *
      * @return bool
      */
-    public function shouldDisableBoltCheckout()
+    public function shouldDisableBoltCheckout($type='')
     {
-        return !$this->isEnabled() || $this->isPageRestricted() || $this->isIPRestricted();
+        return !$this->isEnabled() || $this->isPageRestricted($type) || $this->isIPRestricted();
     }
 }
--- a/Block/Js.php
+++ b/Block/Js.php
@@ -345,7 +345,7 @@ class Js extends Template
      *
      * @return bool
      */
-    private function isPageRestricted($type)
+    private function isPageRestricted()
     {
         $currentPage = $this->getRequest()->getFullActionName();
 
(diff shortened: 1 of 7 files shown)


Block/Js.php
```diff
@@ -354,13 +354,19 @@ private function isPageRestricted()
...
-        // If minicart is supported (allowing Bolt on every page)
+        // If minicart/product page checkout is supported (allowing Bolt on every page)
         // and no IP whitelist is defined there are no additional restrictions.
-        if ($this->configHelper->getMinicartSupport() && !$this->configHelper->getIPWhitelistArray()) {
-            return false;
+        if (!$this->configHelper->getIPWhitelistArray()) {
+            if ($this->configHelper->getMinicartSupport()) {
+                return false;
+            }
+            if ($type=="withproductpage" && ($currentPage=="catalog_product_view") && $this->configHelper->getProductPageCheckoutFlag()) {
```
danac-gs: I don't think $type is necessary. "catalog_product_view" should be enabled by default and added to the default whitelist. Should the JS load can be determined at the begining of the phtml file (getProductPageCheckoutFlag), as done with other files.

Block/Js.php
```diff
@@ -384,8 +390,8 @@ protected function isIPRestricted()
      *
      * @return bool
      */
-    public function shouldDisableBoltCheckout()
+    public function shouldDisableBoltCheckout($type='')
```
danac-gs: consider removing $type

Block/JsProductPage.php
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /**
+     * Check if we support product page checkout for type of current product
+     *
+     * @return boolean
+     */
+    public function IsSupportableType()
+    {
+        if (in_array($this->_product->getTypeId(),array('simple','downloadable'))) {
```
danac-gs: You should probably add **virtual** product type and add this array to Helper\Config as done with $defaultPageWhitelist.
https://www.mageworx.com/wiki/cat/magento-2-product/

Block/JsProductPage.php
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Check if Bolt checkout is restricted on the current loading page.
+     * Takes into account whitelisted pages configuration
+     * as well as the IP restriction.
+     * "Full Action Name", <router_controller_action>, is used to identify the page.
+     *
+     * @return bool
+     */
+    protected function isPageRestricted($type)
```
danac-gs: $type is not used

Block/JsProductPage.php
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Determines if Bolt javascript should be loaded on the current page
+     * and Bolt checkout button displayed. Checks whether the module is active,
+     * bolt button on product is enabled, we support this product type and
+     * the page is Bolt checkout restricted and if there is an IP restriction.
+     *
+     * @return bool
+     */
+    public function shouldDisableBoltCheckout($type='withproductpage')
```
danac-gs: Consider removing the $type if possible.

view/frontend/templates/js/boltjs.phtml
```diff
@@ -20,8 +20,7 @@
  *
  * @var $block \Bolt\Boltpay\Block\Js
  */
-if ($block->shouldDisableBoltCheckout()) return;
-
+if ($block->shouldDisableBoltCheckout("withproductpage")) return;
```
danac-gs: This file is loaded on every page. The "withproductpage" parameter does not seem right.

vitaliyreznikov: For now this file loaded on evert page IF minicart is enabled.
So if minicart is disabled we need to load this file only on product page, cart and checkout page.
But we cound instead parameter use new method and call it shouldDisableBoltScript(). Do you agree?

danac-gs: @vitaliyreznikov my point was it already had worked for the two page types, cart and checkout, without the additional parameter. I guess it should not be much more complicated than adding product page to the list of default pages.

Block/JsProductPage.php
YevhenShBolt: And please use the `const` for product type like:
- `Magento\Catalog\Model\Product\Type::TYPE_SIMPLE`
- `Magento\ConfigurableProduct\Model\Product\Type\Configurable::TYPE_CODE`
- `Magento\Downloadable\Model\Product\Type::TYPE_DOWNLOADABLE`

vitaliyreznikov:
@danac-gs Thank you for your very detailed answers to my private messages.
After our conversation, I think I understand what can be improved.
Function isPageRestricted() performed two roles: it checks if the page is forbidden by white/blacklist, but also it checks if mini cart is enabled.
If I also add to isPageRestricted() condition about product page, then the code becomes too hard.
I propose to remove mini cart check from isPageRestricted()

Also, I think we should not add product page into the white list because, as I see from comments, we wanted to include into it
only noncached pages.

Could you look into code again?

Helper/Config.php
```diff
@@ -250,6 +250,11 @@ class Config extends AbstractHelper
         self::SUCCESS_PAGE_ACTION
     ];
 
+    public static $supportableProductTypes = [
+        \Magento\Catalog\Model\Product\Type::TYPE_SIMPLE,
+        \Magento\Catalog\Model\Product\Type::TYPE_VIRTUAL,
+        \Magento\Downloadable\Model\Product\Type::TYPE_DOWNLOADABLE,
```
daisy1754: downloadable product requires login to purchase. I'm ok not supporting it if it makes logic simpler.

view/frontend/templates/button_product_page.phtml
```diff
@@ -0,0 +1,230 @@
...
+            'Magento_Customer/js/customer-data',
+            'mage/validation/validation',
+            'mage/cookies'
+        ], function ($, authenticationPopup, customerData) {
+            var customer = customerData.get('customer');
+            var shoppingCart = customerData.get('cart');
+            var $qty = jQuery('#qty');
+            var max_qty = <?= $block->getProductQty(); ?>;
+
+            var trackCallbacks = {
```
daisy1754: seems lots of shared code with other file. Would it make sense to combine?

view/frontend/templates/button_product_page.phtml
```diff
@@ -0,0 +1,230 @@
...
+                    items: [{
+                        reference: '<?= $block->getProduct()->getId() ?>',
+                        price: <?= (int) $block->getProduct()->getPrice() ?>,
+                        name: '<?= $block->getProduct()->getName() ?>',
+                        quantity: quantity,
+                    }]
+                };
+
+                var buttonClass = "bolt-product-checkout-button";
+                var hints = {prefill:{}};
```
daisy1754: maybe add TODO fo adding prefill based on logged in user info?

vitaliyreznikov: I created a separate task for it

view/frontend/templates/button_product_page.phtml
vitaliyreznikov: I don't mind, but we need to decide together. I don't know much about magento2, but it seems to me that we could move JS code from *.phtml files into php class.
We can create method into php class that will generate JS code according to parameters.
We did it into woocommerce, but it's a big change. It makes it easier to change JS code, but of course, it affects readability.
@daisy1754 @matt-thomason @danac-gs What's your opinion?

vitaliyreznikov: After the conversation with @matt-thomason I would prefer to put it off and do in future in separate PR

Block/Js.php
```diff
@@ -388,4 +384,30 @@ public function shouldDisableBoltCheckout()
     {
         return !$this->isEnabled() || $this->isPageRestricted() || $this->isIPRestricted();
     }
+
+    /**
+     * Return true if we are on cart page or checkout page
+     */
+    public function IsCartOrCheckoutPage() {
```
danac-gs: Regarding of $this->getPageWhitelist() this can include other pages, not only cart or checkout. Perhaps changing the method name into something more generic, isPageAllowed or something similar.

Helper/Config.php
danac-gs: After thinking it through I agree with @daisy1754, removing it makes more sense.

Block/Js.php
vitaliyreznikov: isOnPageFromWhiteList

PR #263, opened 2019-04-24
Fix bug - fail to apply coupon if it is only applicable to specific shipping method

dumega:
asana task https://app.asana.com/0/941920570700290/1118193263476543/f

### DIFF
--- a/Model/Api/DiscountCodeValidation.php
+++ b/Model/Api/DiscountCodeValidation.php
@@ -573,7 +573,7 @@ class DiscountCodeValidation implements DiscountCodeValidationInterface
             'status'          => 'success',
             'discount_code'   => $couponCode,
             'discount_amount' => abs($this->cartHelper->getRoundAmount($address->getDiscountAmount())),
-            'description'     =>  __('Discount ') . $address->getDiscountDescription(),
+            'description'     => trim( __('Discount ') . $rule->getDescription() ),
             'discount_type'   => $this->convertToBoltDiscountType($rule->getSimpleAction()),
         ];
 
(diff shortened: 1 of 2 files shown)


Model/Api/DiscountCodeValidation.php
```diff
@@ -708,8 +758,8 @@ private function convertToBoltDiscountType($type)
     private function setCouponCode($quote, $couponCode)
     {
         $quote->getShippingAddress()->setCollectShippingRates(true);
-        $quote->setCouponCode($couponCode);
-        $this->cartHelper->saveQuote($quote->collectTotals());
+        $quote->setCouponCode($couponCode)->collectTotals()->save();
+        $this->cartHelper->saveQuote($quote);
```
YevhenShBolt: I think we should not save twice the same object!

YevhenShBolt: What is the reason to add here
`$quote->setCouponCode($couponCode)->collectTotals()->save();`
additional save ?

danac-gs:
@dumega What I meant by testing was not only unit tests but running manual production readiness alike tests on a store with both simple coupons / shipping options and more complicated rules specific to the task and various shipping carriers. My concern is we don't break something that used to work. But I expect you already did that.

dumega:
@danac-gs  @YevhenShBolt @matt-thomason I have run some more tests for this PR today, no issue found for now.

and @YevhenShBolt , for your question,
1. I remove the duplicate save `$this->cartHelper->saveQuote($quote);` and only keep `$quote->setCouponCode($couponCode)->collectTotals()->save();`

2. Then I found that `$this->cartHelper->saveQuote($quote);` does not detect the conditions for coupons if the conditions are applied to shipping method only.

Then I will add unit tests

PR #264, opened 2019-04-24
eliminate Magento_Customer/js/section-config error

danac-gs:


view/frontend/templates/js/replacejs.phtml
```diff
@@ -202,7 +202,8 @@ $trackCallbacks = $block->getTrackCallbacks();
         'Magento_Customer/js/model/authentication-popup',
         'Magento_Customer/js/customer-data',
         'mage/validation/validation',
-        'mage/cookies'
+        'mage/cookies',
+        'domReady!'
```
matt-thomason: Is the `!` necessary here? From what I've read, it's not necessary unless you're adding a resource after it, but I'm not familiar enough with the syntax to know for sure.

danac-gs: @matt-thomason This is what I found and acted upon:

It is both a plugin and a module. If you include it in the the requirements array w/ a trailing ! your module won't execute until it's "safe" to interact w/ the DOM:
```javascript
define(['domReady!'], function () {
    console.info('The DOM is ready before I happen');
});
```
Note that loading and executing are different; you want all your files to load as soon as possible, it's the execution of the contents that is time sensitive.

If you omit the !, then it's just a normal module that happens to be a function, which can take a callback that won't execute before the DOM is safe to interact with:
```javascript
define(['domReady'], function (domReady) {
    domReady(function () {
        console.info('The DOM is ready before I happen');
    });
    console.info('The DOM might not be ready before I happen');        
});
```
So I used it as a plugin (!) to wait for everything to "initialize" before our javascript executes instead of using it as a module and triage the script looking for a particular call that caused issues (found that one too, but this solution seems better).

danac-gs:
@matt-thomason @YevhenShBolt please read my comment before I merge. This saved my day actually. I was aware of the issue for quite a time, unsuccessfully trying to solve it in different ways. It did not cause problems until now so it wan not among my priorities.

PR #269, opened 2019-05-02
Pre-auth back-office create order fix

danac-gs:


Helper/Cart.php
```diff
@@ -610,6 +615,16 @@ private function logAddressData($addressData) {
...
     }
 
+    /**
+     * Check the session type.
+     *
+     * @return bool
+     */
+    private function isBackendSession()
+    {
+        return $this->checkoutSession instanceof \Magento\Backend\Model\Session\Quote;
```
YevhenShBolt: It is better to use `\Magento\Backend\Model\Session\Quote::class` instead of `\Magento\Backend\Model\Session\Quote`

danac-gs: @YevhenShBolt I have checked the whole magento code base, it is used as `\Magento\Backend\Model\Session\Quote`, without the `::class` reference. Everywhere. So I am leaving it as is.

PR #271, opened 2019-05-03
Move holdOnTotalsMismatch after order creation

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/1119769573602581

This PR allows the order to create an online credit memo even if a mismatch occurs

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -722,6 +722,8 @@ class Order extends AbstractHelper
         if (Hook::$fromBolt) {
             // if called from hook update order payment transactions
             $this->updateOrderPayment($order, $transaction, null, $hookType);
+            // Check for total amount mismatch between magento and bolt order.
+            $this->holdOnTotalsMismatch($order, $transaction);
         } else {
             // if called from the store controller return quote and order
             // wait for the hook call to update the payment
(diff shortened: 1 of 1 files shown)


Helper/Order.php
```diff
@@ -722,11 +722,15 @@ public function saveUpdateOrder($reference, $boltTraceId = null, $hookType = nul
...
-        } else {
-            // if called from the store controller return quote and order
-            // wait for the hook call to update the payment
-            return [$quote, $order];
         }
+
+        // Check for total amount mismatch between magento and bolt order.
+        $this->holdOnTotalsMismatch($order, $transaction);
+
+        // if called from the store controller return quote and order
```
daisy1754: comment suggests this shouldn't be called if `Hook::$fromBolt` is true?

Helper/Order.php
```diff
@@ -970,11 +974,6 @@ private function holdOnTotalsMismatch($order, $transaction)
                 ]
             ]);
         });
-
-        throw new LocalizedException(__(
```
daisy1754: I think we still want to return error here so merchant is notified via failed hook? WDYT?

matt-thomason: I think we still want this as well. Merchants are constantly asking for more notifications on failed events during customer calls, so keeping it in is valuable I think.

danac-gs:
LGTM. The change is small and seems OK. My concern is would there be cases when the order state goes back and forth from holded to something else, cluttering the order history. I can not tell it just by the code. Testing this is not strait forward, I usually run different scenarios manually.

PR #276, opened 2019-05-07
received url refactor, add bolt reference to order history, validateS…

danac-gs:
…hippingCost fix for virtual products and backent orders (payment only)

### DIFF
--- a/Controller/Adminhtml/Order/ReceivedUrl.php
+++ b/Controller/Adminhtml/Order/ReceivedUrl.php
@@ -17,47 +17,22 @@
 
 namespace Bolt\Boltpay\Controller\Adminhtml\Order;
 
-use Magento\Framework\App\Action\Action;
 use Magento\Framework\App\Action\Context;
 use Bolt\Boltpay\Helper\Log as LogHelper;
 use Bolt\Boltpay\Helper\Cart as CartHelper;
 use Bolt\Boltpay\Helper\Config as ConfigHelper;
 use Bolt\Boltpay\Helper\Bugsnag;
-use Magento\Framework\Exception\LocalizedException;
-use Magento\Framework\Exception\NoSuchEntityException;
 use Magento\Checkout\Model\Session as CheckoutSession;
-use Magento\Quote\Model\Quote;
 use Magento\Sales\Model\Order;
+use Bolt\Boltpay\Helper\Order as OrderHelper;
 
 /**
  * Class ReceivedUrl
  *
  * @package Bolt\Boltpay\Controller\Adminhtml\Order
  */
-class ReceivedUrl extends Action
+class ReceivedUrl extends \Bolt\Boltpay\Controller\Order\ReceivedUrl
 {
-    /**
-     * @var LogHelper
-     */
-    private $logHelper;
-
-    /**
-     * @var ConfigHelper
-     */
-    private $configHelper;
-    /**
-     * @var Bugsnag
-     */
-    private $bugsnag;
-    /**
-     * @var CartHelper
-     */
-    private $cartHelper;
-    /**
-     * @var CheckoutSession
-     */
-    private $checkoutSession;
-
     /**
      * ReceivedUrl constructor.
      *
@@ -67,6 +42,7 @@ class ReceivedUrl extends Action
      * @param Bugsnag         $bugsnag
      * @param LogHelper       $logHelper
      * @param CheckoutSession $checkoutSession
+     * @param OrderHelper     $orderHelper
      */
     public function __construct(
         Context $context,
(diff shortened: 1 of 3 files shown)


Controller/Adminhtml/Order/ReceivedUrl.php
```diff
@@ -17,47 +17,22 @@
...
 use Magento\Sales\Model\Order;
+use Bolt\Boltpay\Helper\Order as OrderHelper;
 
 /**
  * Class ReceivedUrl
  *
  * @package Bolt\Boltpay\Controller\Adminhtml\Order
  */
-class ReceivedUrl extends Action
+class ReceivedUrl extends \Bolt\Boltpay\Controller\Order\ReceivedUrl
```
YevhenShBolt: Usually for adminhtml controllers uses the class `Magento\Backend\App\Action`.
`\Bolt\Boltpay\Controller\Order\ReceivedUrl` - did not extend from adminhtml so we do not have correct session, permissions and etc.

Model/Api/CreateOrder.php
```diff
@@ -638,7 +649,7 @@ protected function getQtyFromTransaction($transactionItem)
      */
     protected function getShippingAmountFromTransaction($transaction)
     {
-        return $transaction->order->cart->shipping_amount->amount;
+        return (int)@$transaction->order->cart->shipping_amount->amount;
```
YevhenShBolt: Please do not use `@`.  It better to use `property_exists` or `isset` and return 0 if it does not exist.

Controller/Adminhtml/Order/ReceivedUrl.php
danac-gs: @YevhenShBolt 
all our admin controllers use Magento\Framework\App\Action instead of Magento\Backend\App\Action. This works and we never had the problems with session or permissions. This was probably the workaround for maintaining the form key for call made from the checkout javascript. Anyway I tested this and it works. Using Magento\Backend\App\Action had thrown errors last week. I'd leave this like this for now and see what can be done in the future.

Model/Api/CreateOrder.php
danac-gs: done

Controller/Adminhtml/Order/ReceivedUrl.php
YevhenShBolt: @danac-gs.
Some code or vulnerability can call our class from frontend and all dependencies will load and we lost security here. We do not have any session check or auth check from admin side now (but in backend action it exists).
We should refactor it now (or create a ticket for it and make refactoring after...)because in some time we can have a security problem.

Model/Api/CreateOrder.php
```diff
@@ -638,7 +649,10 @@ protected function getQtyFromTransaction($transactionItem)
      */
     protected function getShippingAmountFromTransaction($transaction)
     {
-        return $transaction->order->cart->shipping_amount->amount;
+        if (isset($transaction->order->cart->shipping_amount->amount)) {
+            return $transaction->order->cart->shipping_amount->amount;
```
YevhenShBolt: Maybe we can return to the  (int) or (float) value transformation, do you agree?

Controller/Adminhtml/Order/ReceivedUrl.php
danac-gs: @YevhenShBolt what you said is correct and valid.
However, in practice, with the particular calls we make I don't see the problem. Let me explain. 

We only have 2 admin controllers, Cart\Data and Order\ReceivedUrl. Although Order\Save is still there, it is redundant with the pre-auth flow.

Cart\Data creates the cart / immutable quote, operating exclusively based on the session related data, can not be called out of the admin session (it can but it can not find the quote), has no potentially malicious side effect on the parent quote, the only input that it can take is the billing address sent to bolt (this might be redundant, it is there for payment only from front-end that no one uses and our support for that is not confirmed in the past year).

The Order\ReceivedUrl input is generated and hashed by bolt and verified via signing secret. 

Nevertheless, I will create a task to investigate and refactor this.

YevhenShBolt: @danac-gs 
ok. it is good for me with ticket.
Below left  one more review.

Model/Api/CreateOrder.php
danac-gs: done

Controller/Adminhtml/Order/ReceivedUrl.php
danac-gs: I also made Bolt\Boltpay\Controller\Order\ReceivedUrl methods private

PR #282, opened 2019-05-14
Multi-website support

YevhenShBolt:


Helper/Cart.php
```diff
@@ -254,12 +259,23 @@ public function isCheckoutAllowed()
...
+                ->addFilter('main_table.entity_id', $quoteId)->create();
+
+            $collection = $this->quoteRepository
+                ->getList($searchCriteria)
+                ->getItems();
+
+            $this->quotes[$quoteId] = reset($collection);
+        }
+
+        return $this->quotes[$quoteId];
```
danac-gs: What's the reason behind this change?

YevhenShBolt: @danac-gs , In the backend when i tries to load quote from repository, it comes without correct store id (it was always default store id). I looked deeper and found that repository and his method `get` or `getActive` - loads quote data without store id (the option `loadByIdWithoutStore`). It turned out to be critical for the backend.

Regarding caching a `quote` object it is useful because method `getQuoteById` calling minimum twice, and in second call we does not need to load again data from db.

danac-gs: @YevhenShBolt Awesome. Thanks.

PR #284, opened 2019-05-15
Feature implement review api

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/945931188955500

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -1224,7 +1224,7 @@ class Order extends AbstractHelper
                 break;
 
             case self::TS_REJECTED_REVERSIBLE:
-                $orderState = OrderModel::STATE_HOLDED;
+                $orderState = OrderModel::STATE_PAYMENT_REVIEW;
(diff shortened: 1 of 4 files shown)


Helper/Order.php
```diff
@@ -1224,7 +1224,7 @@ public function updateOrderPayment($order, $transaction = null, $reference = nul
                 break;
 
             case self::TS_REJECTED_REVERSIBLE:
-                $orderState = OrderModel::STATE_HOLDED;
+                $orderState = OrderModel::STATE_PAYMENT_REVIEW;
```
danac-gs: How will the merchant know that the payment approve / reject action needs to take place?
How will this orders be easily distinguishable?

johnmccombs1: @danac-gs this uses default Magento functionality. The merchant will see two buttons on the admin's View Order page https://screencast.com/t/1H8v1ee6d6G8.

danac-gs: @johnmccombs1 awesome. thanks for the explanation.

PR #288, opened 2019-05-16
Support Mageplaza gift card

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/1121586614366893

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -166,7 +166,8 @@ class Cart extends AbstractHelper
         Discount::GIFT_CARD_ACCOUNT => '',
         Discount::UNIRGY_GIFT_CERT => '',
         Discount::AMASTY_GIFTCARD => 'Gift Card ',
-        Discount::GIFT_VOUCHER => ''
+        Discount::GIFT_VOUCHER => '',
+        Discount::MAGEPLAZA_GIFTCARD => ''
     ];
     /////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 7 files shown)


Model/Api/DiscountCodeValidation.php
```diff
@@ -282,6 +282,12 @@ public function validate()
                 $giftCard = $this->discountHelper->loadAmastyGiftCard($couponCode);
             }
 
+            // Apply Mageplaza_GiftCard
+            if (empty($giftCard)) {
+                // Load the gift card by code
+                $giftCard = $this->discountHelper->loadMageplazaGiftCard($couponCode);
+            }
+
```
danac-gs: You should add a section below in applyingGiftCardCode method where this $giftCard is applied to quote / immutable quote. Every gift cart module has it's own logic.

johnmccombs1: I've updated it.

Helper/Cart.php
```diff
@@ -166,7 +166,9 @@ class Cart extends AbstractHelper
         Discount::GIFT_CARD_ACCOUNT => '',
         Discount::UNIRGY_GIFT_CERT => '',
         Discount::AMASTY_GIFTCARD => 'Gift Card ',
-        Discount::GIFT_VOUCHER => ''
+        Discount::GIFT_VOUCHER => '',
```
ethanwayda22: The code seems to be duplicated here

johnmccombs1:
> @johnmccombs1
> The code looks OK. However, the part regarding applying GC via Bolt checkout still does not work. At least in the latest M2 version, 2.3.1.

I've set up a new Magento 2.3.1 instance, tested, and confirmed it works fine, see the video [here](https://drive.google.com/file/d/1IjTOGOvVb1H7TF8t3Jjl1n6VY3woS6Ac/view). You can test it to confirm on this site https://m231.bolt.devzn.com/with the admin credentials below:
> https://m231.bolt.devzn.com/admin
> coredev1/Coredev12

PR #291, opened 2019-05-20
preauth improve SKU based validation

danac-gs:


Model/Api/CreateOrder.php
```diff
@@ -422,34 +434,38 @@ public function validateCartItems($quote, $transaction)
...
      * @param int $itemQty
+     * @param int $productId
      * @throws BoltException
      * @throws NoSuchEntityException
      */
-    public function validateItemInventory($itemSku, $itemQty)
+    public function validateItemInventory($itemSku, $itemQty, $productId)
     {
-        $stockItem = $this->stockRegistry->getStockItemBySku($itemSku);
+        $stockItem = $this->stockRegistry->getStockItem($productId);
```
YevhenShBolt: @danac-gs What is the reason to use `getStockItem($productId)` instead of `getStockItemBySku($itemSku)`?

As i see, in case `getStockItemBySku()` we have two query to the DB: 

- run query to get product Id by sku
- run query to get stockItem by product ID

in case  `getStockItem()` we have one query to the DB: 

- run query to get stockItem by product ID

Did you make changes for the same purpose?
Does i miss something?

danac-gs: @YevhenShBolt that is one purpose but more important one is avoiding the SKU because it can not be relied upon. Foe example one of the merchants had the SKU for a product saved with trailing spaces, "FIDDLESTICKBLCK    ". The SKU gets trimmed at Bolt end, so we trim it at plugin level and use it like that. But that would not result in a successful query to the DB where the SKU is not trimmed. Actually Magento does not allow you to save a string like that as a SKU from the admin in the newer versions, but at some point it was possible and that got into DB. We had similar SKU issues before pre-auth.

YevhenShBolt: Nice, thanks

PR #292, opened 2019-05-21
Implement constructor to ensure error code is taken into account (on older versions)

danac-gs:


Exception/BoltException.php
```diff
@@ -18,8 +18,22 @@
...
+     * Bypass the parent and call grandparent's constructor
+     *
+     * @param \Magento\Framework\Phrase $phrase
+     * @param \Exception $cause
+     * @param int $code
+     */
+    public function __construct(Phrase $phrase, \Exception $cause = null, $code = 0)
+    {
+        $this->phrase = $phrase;
+        \Exception::__construct($phrase->render(), intval($code), $cause);
```
YevhenShBolt: I think the best way to do this is:
```
$this->phrase = $phrase;
parent::__construct($phrase->render(), $cause);
$this->code = intval($code);
```

danac-gs: @YevhenShBolt done

PR #294, opened 2019-05-23
Support Mirasvit Store Credit

johnmccombs1:
Asana task: https://app.asana.com/0/564264490825835/1124038059584813

### DIFF
--- a/Helper/Discount.php
+++ b/Helper/Discount.php
@@ -17,6 +17,7 @@
 
 namespace Bolt\Boltpay\Helper;
 
+use Bolt\Boltpay\Model\Payment;
 use Magento\Framework\App\Helper\AbstractHelper;
 use Magento\Framework\App\Helper\Context;
 use Magento\Framework\App\ResourceConnection;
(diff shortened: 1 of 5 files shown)


Helper/Discount.php
```diff
@@ -400,4 +424,92 @@ public function getUnirgyGiftCertBalanceByCode($giftcertCode)
...
+        $miravitBalanceAmount = $this->getMirasvitStoreCreditBalanceAmount($quote);
+        if(!$paymentOnly){
+            /** @var \Mirasvit\Credit\Api\Config\CalculationConfigInterface $miravitConfig */
+            $miravitConfig = $this->mirasvitStoreCreditConfig->getInstance();
+            if ($miravitConfig->isTaxIncluded() || $miravitConfig->IsShippingIncluded()){
+                return $miravitBalanceAmount;
+            }
+        }
+
+        $toUse = $quote->getGrandTotal() + $quote->getCreditAmountUsed();
```
mccottry: @johnmccombs1 
It's unclear to me here what is going on.  I imagine that here you are looking into all the other non-Miravit store credit balances that are used, `$quote->getCreditAmountUsed()`, and subtracting that from the grand total, (i.e. I would expect that value to be negative).

So, `$toUse` could be labeled something like `$outstandingBalance`, or `$unresolvedTotal` and ultimately, you return min($unresolvedTotal, $miravitBalanceAmount).

Is this interpretation correct?

johnmccombs1:
> @johnmccombs1
> I have a question below, but, as I interpret the code, it LGTM.

I agree with your comments. I do like well broken out code but sometimes I lose the forest for the trees :)

PR #297, opened 2019-05-31
PreAuth multi-website: check quoteItem errors.

YevhenShBolt:
multi-website support for create_order, refactoring quoteItem validation.

After research i found that when we load quote items collection, calls an event, where the Magento updates stock info for every quoteItem. So, after Magento was validated, we have error messages if something happened with quote Item.
Example response for Bolt:
``` 
body:[{"status":"failure","error":[{"code":2001001,"data":[{"reason":"(cataloginventory): We don't have as many \"Configurable Product 121 - option 1\" as you requested."}]}]}]
```

### DIFF
--- a/Model/Api/CreateOrder.php
+++ b/Model/Api/CreateOrder.php
@@ -182,10 +182,6 @@ class CreateOrder implements CreateOrderInterface
                 );
             }
 
-            $payload = $this->request->getContent();
-
-            $this->validateHook();
-
             if (empty($order)) {
                 throw new BoltException(
                     __('Missing order data.'),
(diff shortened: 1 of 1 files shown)


Model/Api/CreateOrder.php
```diff
@@ -451,42 +462,44 @@ public function validateCartItems($quote, $transaction)
...
+            $report->setMetaData([
+                'Pre Auth' => [
+                    'quoteItem errors' => $message,
+                ]
+            ]);
+        });
+        throw new BoltException(
+            __($message),
+            null,
+            self::E_BOLT_GENERAL_ERROR
```
danac-gs: Can we throw E_BOLT_ITEM_OUT_OF_INVENTORY error here? Vast majority if not all errors we catch should be exactly that. Or explicitly check $errorInfo code property to match ERROR_QTY, vendor/magento/module-catalog-inventory/Helper/Data.php

YevhenShBolt: Updated

PR #300, opened 2019-06-04
Multi-website: fix disappearing Bolt button at first load in admin side

YevhenShBolt:
Task - https://app.asana.com/0/941920570700290/1125041703113784/f

The problem was at returned 0 instead of null if we have clean quote.
Magento could not find store with id 0 in 1 case (see below).

We have several cases for multi-store which already covered:

- [x] 1. We do not have Bolt's api keys in default store view.
- [x] 2. We have Bolt's api keys in default store view.
- [x] 3. We have different Bolt's ApiKey and Signing Secret keys for different website/stores.

### DIFF
--- a/Block/Js.php
+++ b/Block/Js.php
@@ -22,7 +22,7 @@ use Magento\Framework\View\Element\Template;
 use Magento\Framework\View\Element\Template\Context;
 use Magento\Framework\Session\SessionManager as CheckoutSession;
 use Bolt\Boltpay\Helper\Cart as CartHelper;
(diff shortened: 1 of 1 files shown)


Block/Js.php
```diff
@@ -444,15 +444,13 @@ public function shouldDisableBoltCheckout()
     public function getMagentoStoreId()
     {
         if ($this->checkoutSession instanceof BackendSessionQuote) {
```
danac-gs: you don't need this anymore

YevhenShBolt: updated

PR #301, opened 2019-06-05
MultiWebsite updates.

YevhenShBolt:
- reverted Daniel's commit to proper store_id determination for admin side.
- update magentoStoreId JS variable check.

The main problem was in layout file. The functionality of js or templates and etc should be inside order page scope. If we placed code to `js` container (out of scope of order content) it is out of scope and kill ability to determine store or update data for our plugin, when we choose another store or change some shipping data or address data.


YevhenShBolt:
@matt-thomason But without this fix the Bolt payment button in admin side do not appear.
I hope that fix should be in develop also, but need that Daniel (as he said) fix reorder case.

danac-gs:
@YevhenShBolt @matt-thomason I am closing this PR.

PR #302, opened 2019-06-07
Multiwebsite support fix

danac-gs:


Model/Api/OrderManagement.php
```diff
@@ -133,10 +133,14 @@ public function manage(
...
+            $storeId = $this->orderHelper->getStoreIdByQuoteId($order)
+                    ?: $this->orderHelper->getOrderStoreIdByDisplayId($display_id);
 
-            $this->logHelper->addInfoLog('Order StoreId: ' . $orderStoreId);
+            $this->hookHelper->setMagentoStoreId($storeId);
+
+            $this->logHelper->addInfoLog('StoreId: ' . $storeId);
 
             $this->hookHelper->setCommonMetaData();
             $this->hookHelper->setHeaders();
```
YevhenShBolt: We have special method for checking webhook.
Instead of 
```
  $this->hookHelper->setMagentoStoreId($storeId);
  $this->hookHelper->setCommonMetaData(); 
  $this->hookHelper->setHeaders();
  $this->hookHelper->verifyWebhook();
```

we can simply use

```
  $this->hookHelper->preProcessWebhook($storeId)
```

Model/Api/ShippingMethods.php
```diff
@@ -319,11 +319,10 @@ public function getShippingMethods($cart, $shipping_address)
             // Load quote from entity id
             $quote = $this->cartHelper->getQuoteById($quoteId);
```
YevhenShBolt: In the model class `Model/Api/ShippingMethods.php` exist method `getQuoteById()`.
We can simplify it and just call like `$this->getQuoteById($quoteId)`.

view/adminhtml/templates/boltpay/js/replacejs.phtml
```diff
@@ -318,21 +317,19 @@ require([
...
-            publishableKey = getCheckoutKey();
+    var insertConnectScript = function(publishableKey) {
+        var scriptTag = document.getElementById('bolt-connect');
         if (scriptTag) {
-            scriptTag.setAttribute('data-publishable-key', publishableKey);
-            createOrder();
+            BoltCheckout.configure(cart, hints, callbacks);
             return;
         }
+        if (!publishableKey) return;
```
daisy1754: let's do `console.error`

Controller/Adminhtml/Cart/Data.php
```diff
@@ -115,16 +116,11 @@ public function execute()
             $result = $this->dataObjectFactory->create();
             $result->setData('cart', $cart);
             $result->setData('hints', $hints);
+            $result->setData('publishableKey', @$publishableKey);
```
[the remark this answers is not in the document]
danac-gs: @YevhenShBolt the admin page logic is changed. The connect.js is loaded programmatically only after we get the publishable key. It is not pre-loaded anymore. This way we always first get the right key and then send the request to load connect.js with the appropriate publishable key. The publishable key is retrieved with the cart data response. The cart, and the whole response, is valid only when all the required fields are populated, i.e. items in shopping cart, customer email and address and the shipping method. The cart endpoint is called on every page change with payment sections reloaded (by Magento). I think you did that part. If there's a non empty response there's a cart, eventually hints and the publishable key. On the first successful response the connect.js is loaded. Until then it is expected to not have bolt cart in some results (some of the data still missing) and no publishable key as well. So this is not an error suppression it is just covering the case when we still don't have it.

Controller/Adminhtml/Cart/Data.php
```diff
@@ -115,16 +116,11 @@ public function execute()
...
+            $result->setData('storeId', @$storeId);
```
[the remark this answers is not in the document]
danac-gs: The same as above. Whenever there's a quote, there's also a store id attached to it. If read from the quote table it is never null. Basically we wait for the quote to be created and then read store id with the rest of the cart data, get the publishable key for that particular store and send it back to the front-end with the cart. Then, if there's a key in response and connect.js has not been loaded yet we load it with the right key. Only the first time the key / cart / store_id arrives.

Helper/Cart.php
```diff
@@ -434,7 +433,7 @@ protected function saveCartSession($qouoteId)
     protected function getSessionQuoteStoreId()
     {
         $sessionQuote = $this->checkoutSession->getQuote();
-        return $sessionQuote ? $sessionQuote->getStoreId() : null;
+        return $sessionQuote ? ($sessionQuote->getStoreId() ?: null) : null;
```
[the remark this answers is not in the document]
danac-gs: I agree

Helper/Cart.php
```diff
@@ -283,7 +283,6 @@ public function isCheckoutAllowed()
      * Load Quote by id
      * @param $quoteId
      * @return \Magento\Quote\Model\Quote
-     * @throws NoSuchEntityException
```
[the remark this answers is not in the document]
danac-gs: Because it is not thrown anymore since your (justified and reasonable )changes to how this method works. It returns false when not found and we should update the code when the exception is expected.

Model/Api/OrderManagement.php
```diff
@@ -133,10 +133,14 @@ public function manage(
 
             $this->logHelper->addInfoLog($this->request->getContent());
 
-            $orderStoreId = $this->orderHelper->getOrderStoreIdByDisplayId($display_id);
-            $this->hookHelper->setMagentoStoreId($orderStoreId);
+            // get the store id. try fetching from quote first,
+            // otherwise load it from order if there's one created
+            $storeId = $this->orderHelper->getStoreIdByQuoteId($order)
```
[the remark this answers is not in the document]
danac-gs: Seems correct. However, since the OrderManagement.php is the only place it's used and Helper\Order.php is already loaded I tried to avoid injecting Helper\Cart.php just for that.
I'd leave it like this for now.

Model/Api/OrderManagement.php
danac-gs: correct

Model/Api/ShippingMethods.php
danac-gs: done

view/adminhtml/templates/boltpay/js/replacejs.phtml
danac-gs: @daisy1754 I will add it but I don't think it is necessary. Not having publishable key set is expected behavior. Please read the first two comments to Yevhen above.

Model/Api/DiscountCodeValidation.php
```diff
@@ -918,7 +919,11 @@ public function loadGiftCertData($code)
                 /** @var \Unirgy\Giftcert\Model\Cert $giftCert */
                 $giftCert = $giftCertRepository->get($code);
 
-                $result = ($giftCert->getData('status') !== 'I') ? $giftCert : null;
+                $gcStoreId = $giftCert->getStoreId();
```
ethanwayda22: Could we get comments here breaking down whats going a bit deeper?

PR #304, opened 2019-06-11
Support custom amount for Mirasvit Store Credit

johnmccombs1:


Helper/Discount.php
```diff
@@ -497,7 +497,7 @@ protected function getMirasvitStoreCreditBalanceAmount($quote)
                         ->getInstance()
                         ->getBalance($quote->getCustomerId(), $quote->getQuoteCurrencyCode());
 
-        $amount = $balance->getAmount();
+        $amount = ((float)$quote->getManualUsedCredit() > 0) ? $quote->getManualUsedCredit() : $balance->getAmount();
```
daisy1754: is `getManualUsedCredit` built in method? If not, when this method doesn't exist what would happen?

danac-gs: @daisy1754 @mccottry getters are magic methods in Magento 2. They don't need to be explicitly defined. NULL is returned if property is not found. Which is safe in this case.

matt-thomason: Can we add a comment to this code summarizing what @mccottry explained? Someone coming in fresh looking at this code might not understand this is expected to be dependency injected from `MirasvitStoreCredit` magically.

johnmccombs1: Thank @danac-gs, @daisy1754 it is a magic method. The code works as expected, and even the manual_used_credit doesn't exist on the quote. Magic methods essentially allow us to pass data around without explicitly defining data on the model. @matt-thomason I'll push a new update for the comment.

PR #308, opened 2019-06-20
Fulfills merchant request to configurably minify our JS for site performance metrics

mccottry:
https://app.asana.com/0/941920570700290/1124390364971337

### DIFF
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -25,8 +25,9 @@
 if ($block->shouldDisableBoltCheckout()) return;
 
 $trackCallbacks = $block->getTrackCallbacks();
-?>
 
+ob_start();
+?>
 <script type="text/javascript">
 
     // Store the configuration parameters passed from the php block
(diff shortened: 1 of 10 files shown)


view/frontend/templates/js/replacejs.phtml
```diff
@@ -25,8 +25,9 @@
 if ($block->shouldDisableBoltCheckout()) return;
 
 $trackCallbacks = $block->getTrackCallbacks();
-?>
 
+ob_start();
```
ethanwayda22: I am curious for myself, what are ob_start(); and ob_get_clean(); doing in this piece?

ethanwayda22: Could we possibly leave comments here to explain or is this documented somewhere else?

mccottry: @ethanwayda22  That's a good question. I was thinking the same about some Docker principles.

**ob_start()** is very standard PHP.  **ob** stands for **output buffering**.  It is documented somewhere else, used extensively throughout the underlying  code in all major PHP shopping cart frameworks (and in all major web frameworks), and should be safe to assume that maintainers of this code will be aware of their functionality or will be able to quickly read up on them and understand their purposes.

Find references here:
https://www.php.net/manual/en/book.outcontrol.php
https://www.php.net/manual/en/function.ob-start.php
https://www.php.net/manual/en/function.ob-get-clean.php

The main page is here and is an invaluable reference for any and all PHP development:
https://www.php.net/manual/en/

PR #312, opened 2019-06-22
Php CS: show actual magento 2 errors

roopakv:


roopakv:
@daisy1754 	any idea what the best standard to use in this case would be. Seems like MEQP2 doesnt get installed by default anymore.

PR #316, opened 2019-06-28
Update phpcs to pass

matt-thomason:
Used custom PHPCS file, removed the @ check and fixed/ignored the other failing checks.
https://app.asana.com/0/941920570700290/1128432110428106/f

### DIFF
--- a/Model/Api/ShippingMethods.php
+++ b/Model/Api/ShippingMethods.php
@@ -40,6 +40,7 @@ use Magento\Framework\Pricing\Helper\Data as PriceHelper;
 use Bolt\Boltpay\Model\ErrorResponse as BoltErrorResponse;
 use Bolt\Boltpay\Helper\Session as SessionHelper;
 use Bolt\Boltpay\Exception\BoltException;
+use Magento\Framework\Serialize\SerializerInterface as SerializerInterface;
 
 /**
  * Class ShippingMethods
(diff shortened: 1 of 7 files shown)


matt-thomason:
There are basically 4 categories of errors that this addresses:
- `final` - removed these
- `@` ignore warning - I think we use these appropriately, so I removed this check from `phpcs.xml`.
- `serialize` and `unserialize` insecurity. We still support Magento 2.1 which doesn't have the recommended M2 Serialize library, so we can't use that for backwards compatibility reasons.
- `md5` insecurity. I left in this check so we are encouraged to use a more secure encoding in the future, but we can't change what we have right now for backwards compatibility. I added annotated exceptions for our current `md5` usages.

Model/Api/ShippingMethods.php
```diff
@@ -468,7 +468,9 @@ function ($el) {
 
             $cacheIdentifier .= '_' . $quote->getStoreId();
 
+            // phpcs:disable
```
ethanwayda22: should we remove these comments or are they necessary?

matt-thomason: They are necessary - they are what I referred to in my above comment:
> I added annotated exceptions for our current md5 usages.

Codesniffer ignores the style between `disable` and `enable`.

PR #318, opened 2019-06-29
Aheadworks store credit

danac-gs:


Helper/Discount.php
```diff
@@ -714,4 +723,25 @@ public function applyMageplazaDiscountToQuote($quote)
...
+
+    /**
+     * Get Aheadworks store credit for the user
+     *
+     * @param int $customerId Logged in customer ID
+     * @return float User store credit
+     */
+    public function getAheadworksStoreCredit($customerId)
+    {
+        return $this->aheadworksCustomerStoreCreditManagement->getInstance()->getCustomerStoreCreditBalance($customerId);
```
YevhenShBolt: @danac-gs i guess that `getInstance()` returns `null` if we do not have that instance/class in Magento.
So there we maybe have a php error or warning.

danac-gs: @YevhenShBolt this method is called only after the Aheadworks Store Credit key is found in the quote totals. However, I am adding additional check for clarity.
```php
if (array_key_exists(Discount::AHEADWORKS_STORE_CREDIT, $totals)) {
    $customerId = $this->customerSession->getCustomer()->getId();
    $aheadStoreCredits = $this->discountHelper->getAheadworksStoreCredit($customerId);
    $roundedAmount = abs($this->getRoundAmount($aheadStoreCredits));
    $cart['discounts'][] = [
        'description' => 'Store Credit',
        'amount'      => $roundedAmount,
        'type'        => 'fixed_amount',
    ];
    $totalAmount -= $roundedAmount;
}
```

PR #319, opened 2019-07-01
amasty reward points support

danac-gs:


YevhenShBolt:
@danac-gs .
Does `Amasty_Rewards` have some layer of abstraction under delete?
Does this module have some `Repository` or `Model` instead of `$rewardsTable = $this->resource->getTableName('amasty_rewards_quote');`

Can you research and try to find such abstraction?
I think we should use abstraction and break down responsibility between modules.

Model/Api/ShippingMethods.php
```diff
@@ -430,7 +458,7 @@ public function shippingEstimation($quote, $addressData)
 
             $cacheIdentifier = $parentQuoteId.'_'.round($quote->getSubtotal()*100).'_'.
                 $addressData['country_code']. '_'.$addressData['region'].'_'.$addressData['postal_code']. '_'.
-                @$addressData['street_address1'].'_'.@$addressData['street_address2'];
+                @$addressData['street_address1'].'_'.@$addressData['street_address2'].'_'.$externalData;
```
YevhenShBolt: The `@` - will produce a slow performance and also dismiss error. It is better to check `street_address1 `, `street_address1 ` otherwise return empty string. 
Please change it.

danac-gs:
> @danac-gs . LGTM
> Does `Amasty_Rewards` have some layer of abstraction under delete?
> Does this module have some `Repository` or `Model` instead of `$rewardsTable = $this->resource->getTableName('amasty_rewards_quote');`
> 
> Can you research and try to find such abstraction?
> I think we should use abstraction and break down responsibility between modules.

@YevhenShBolt there is no abstraction layer around delete for this table / model. It is never deleted from by the module. The data is fetched and used as an array in the module itself.  We delete the redundant immutable quote related entries to clear the DB and leave it in the expected state. Also, this is the most efficient way.

Model/Api/ShippingMethods.php
danac-gs: @YevhenShBolt there is no error here. Sometimes there is an address and sometimes not (prefetch shipping from the shopping cart page and probably geolocation, apple pay shipping and tax call..). Also the performance impact here is meaningless. I am leaving it as is.

matt-thomason: @YevhenShBolt Using `@` in this context costs about 0.005 ms. There is certainly a trade-off and a discussion to be had here, but I don't think it's about performance. A better argument against `@` is that it is against the `Magento2` PHPCS rules - we had to write a custom sniffer to get around that. I think for now, we should keep and use these.

YevhenShBolt: @matt-thomason @danac-gs 
1. I would probably agree that this suppressor is not so critical for this piece of code.
2. @matt-thomason  It is not clear for me why we should have our own rule if Magento provides recommendations on best practices and monitors the purity of the code. It also seems Magento checks the code with the same validator before loading it into its marketplace.

I think we should move `phpcs rule` discussion to another place. :-)

YevhenShBolt:
@danac-gs i understand how `DELETE` works and for what is it. 

I am just worried about future versions of `Amasty_Rewards` module and our support. For example, they adds some other table or decides to change table name in their purpose - after that we should supports the both versions or any other table changes with such code.

I am OK with it now, thanks for clarification. It is just thinking out loud.

PR #321, opened 2019-07-02
removed redundant area code emulation

danac-gs:
Turns out this was causing problems. Ran some quick tests with the oldest and the newest Magento version we support.

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -45,7 +45,6 @@ use Bolt\Boltpay\Model\Service\InvoiceService;
 use Zend_Http_Client_Exception;
 use Magento\Sales\Model\Order\Invoice;
 use Magento\Framework\DataObjectFactory;
-use Magento\Framework\App\State;
 use Bolt\Boltpay\Helper\Log as LogHelper;
 use Bolt\Boltpay\Helper\Cart as CartHelper;
 use Magento\Framework\App\ResourceConnection;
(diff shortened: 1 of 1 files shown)


Helper/Order.php
```diff
@@ -1472,16 +1465,12 @@ public function updateOrderPayment($order, $transaction = null, $reference = nul
         $this->setOrderState($order, $orderState);
 
         // Send order confirmation email to customer.
-        // Emulate frontend area in order for email
-        // template to be loaded from the correct path
         if ( ! $order->getEmailSent() ) {
-            $this->appState->emulateAreaCode('frontend', function () use ($order) {
```
mccottry: @danac-gs 
I wonder if 
`Mage::getDesign()->setArea('frontend')` 
is useful in some cases.  Maybe not, but something to keep in mind.

YevhenShBolt: I rememberedю
@mccottry we added this part of codewith Kazuki for specific merchant with Enterprise Edition.
In past we could not establish why it works that way for him.

danac-gs: @YevhenShBolt @mccottry this was called in the past from the adminhtml context via save order controller for the back-office orders. Now it is called only from the webapi_rest which seams to fetch the correct frontend templates by default. Not sure why it does not work with this, to be honest.

PR #322, opened 2019-07-04
Fix for accepting pending discount codes

danac-gs:


Model/Api/DiscountCodeValidation.php
```diff
@@ -921,7 +921,7 @@ public function loadGiftCertData($code, $storeId)
 
                 $gcStoreId = $giftCert->getStoreId();
 
-                $result = ((!$gcStoreId || $gcStoreId == $storeId) && $giftCert->getData('status') !== 'I')
+                $result = ((!$gcStoreId || $gcStoreId == $storeId) && $giftCert->getData('status') === 'A')
```
danac-gs: The problem was that we use to check only if the certificate is not Inactive (**I**), so we allowed for pending ones to be applied which was wrong. Now we explicitly check if it is Active (**A**).

YevhenShBolt: thanks

PR #323, opened 2019-07-05
Fix for the failed payment validation logic

danac-gs:


Model/Api/CreateOrder.php
```diff
@@ -616,7 +624,7 @@ public function validateTotalAmount($quote, $transaction)
         $quoteTotal = $this->cartHelper->getRoundAmount($quote->getGrandTotal());
         $transactionTotal = $this->getTotalAmountFromTransaction($transaction);
 
-        if (abs($quoteTotal - $transactionTotal) > OrderHelper::MISMATCH_TOLERANCE) {
+        if ($quoteTotal != $transactionTotal) {
```
danac-gs: There should be no tolerance for the grand total amount.

Model/Api/CreateOrder.php
```diff
@@ -192,21 +197,24 @@ public function execute(
...
             $quote = $this->orderHelper->prepareQuote($immutableQuote, $transaction);
-            $this->validateQuoteData($quote, $transaction);
 
-            /** @var \Magento\Sales\Model\Order $createOrderData */
-            $createOrderData = $this->orderHelper->preAuthCreateOrder($quote, $transaction);
+            /** @var \Magento\Sales\Model\Order $createdOrder */
+            $createdOrder = $this->orderHelper->processExistingOrder($quote, $transaction);
+
+            if (! $createdOrder) {
+                $this->validateQuoteData($quote, $transaction);
```
danac-gs: The validation was moved after the check for an existing order because it can fail on inventory validation if the last item(s) of the product got reserved in that order. In processExistingOrder we call validation of totals.

Helper/Order.php
```diff
@@ -857,21 +856,20 @@ public function processNewOrder($quote, $transaction)
...
+        $taxAmount = $this->cartHelper->getRoundAmount($order->getTaxAmount());
+        $shippingAmount = $this->cartHelper->getRoundAmount($order->getShippingAmount());
+        $grandTotalAmount = $this->cartHelper->getRoundAmount($order->getGrandTotal());
 
-        $transactionSubtotalAmount = $transaction->order->cart->subtotal_amount->amount;
         $transactionTaxAmount = $transaction->order->cart->tax_amount->amount;
         $transactionShippingAmount = $transaction->order->cart->shipping_amount->amount;
         $transactionGrandTotalAmount = $transaction->order->cart->total_amount->amount;
 
-        return abs($subtotal - $transactionSubtotalAmount) <= self::MISMATCH_TOLERANCE
```
danac-gs: The subtotal check was removed for 2 reasons:
1. there's a grand totals check, so it is redundant
2. in tests I found $transaction->order->cart->subtotal_amount was not set (null) in the transaction data fetched from Bolt. This should probably be reported <- @matt-thomason

Model/Api/CreateOrder.php
```diff
@@ -192,21 +197,24 @@ public function execute(
...
 
-            $payload = $this->request->getContent();
             $transaction = json_decode($payload);
 
             /** @var Quote $quote */
             $quote = $this->orderHelper->prepareQuote($immutableQuote, $transaction);
-            $this->validateQuoteData($quote, $transaction);
 
-            /** @var \Magento\Sales\Model\Order $createOrderData */
-            $createOrderData = $this->orderHelper->preAuthCreateOrder($quote, $transaction);
```
danac-gs: preAuthCreateOrder was split into 2 methods, processExistingOrder and processNewOrder, with validateQuoteData put in between them

Helper/Order.php
```diff
@@ -762,42 +762,30 @@ public function saveUpdateOrder($reference, $storeId = null, $boltTraceId = null
...
-    public function preAuthCreateOrder($quote, $transaction)
+    public function processExistingOrder($quote, $transaction)
     {
         // check if the order has been created in the meanwhile
         if ($order = $this->getExistingOrder($quote->getReservedOrderId())) {
             if ($this->hasSamePrice($order, $transaction)) {
                 return $order;
-            } else {
-                $this->deleteOrder($order);
             }
```
YevhenShBolt: I want to suggest add here an event which will help a merchant's system to avoid suddenly disappearing data.
I am talking about some ERP or Reporting system which collect order data. It is rarely but happened.

Helper/Order.php
```diff
@@ -762,42 +762,30 @@ public function saveUpdateOrder($reference, $storeId = null, $boltTraceId = null
...
-     * @return OrderModel
-     * @throws BoltException
-     * @throws LocalizedException
+     * Try to fetch and price-validate already existing order.
+     * Use case: order has been created but the payment failes due to the ivnalid credit card data.
+     * Then the customer enters the correct card info, the previously created order is used if the amounts match.
+     *
+     * @param Quote $quote
+     * @param \stdClass $transaction
+     * @return bool|false|OrderModel
```
YevhenShBolt: As i see we alway return here of `false` or `OrderModel `.
Can leave here only `false|OrderModel`? How do you think?

Helper/Order.php
YevhenShBolt: As i see our method `deleteOrder()` make cancel at first and then `delete`.
As alternative we can add event directly to the method `deleteOrder()`.

PR #326, opened 2019-07-11
Do not allow checkout after Declined Payment card

danac-gs:
https://app.asana.com/0/941920570700290/1130810862387296/f


matt-thomason:
@danac-gs There's no changes in this now after revert - are you going to add more or should this just be closed?

danac-gs:
> @danac-gs There's no changes in this now after revert - are you going to add more or should this just be closed?

@matt-thomason I will apply the solution discussed with @nirajjayantbolt Let it stay. I will mark it WIP.

PR #329, opened 2019-07-17
Add custom address fields to order token cache identifier

danac-gs:
Custom address field (e.g. commercial / residential flag used by restaurant supply) does not get taken into account when creating bolt order / caching bolt order token.

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -491,33 +491,46 @@ class Cart extends AbstractHelper
     }
 
     /**
+     * Get array of custom address field names converted to PascalCase
+     * used to build method names (i.e. getters)
+     *
+     * @param int $storeId
+     * @return array
+     */
+    protected function getCustomAddressFieldsPascalCaseArray($storeId)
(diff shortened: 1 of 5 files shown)


Helper/Cart.php
```diff
@@ -483,7 +483,49 @@ protected function getCartCacheIdentifier($cart)
...
+        );
+        // trim values and filter out empty strings
+        $prefetchAddressFields = array_filter(array_map('trim', $prefetchAddressFields));
+        // convert to PascalCase
+        $prefetchAddressFields = array_map(
+            function ($el) {
+                return str_replace('_', '', ucwords($el, '_'));
+            },
+            $prefetchAddressFields
+        );
```
YevhenShBolt: Maybe better to extract these three variables `$prefetchAddressFields` and operations with them to the separate method.
After that we have good ability to cover it with unit tests.

danac-gs: @YevhenShBolt Done

PR #337, opened 2019-07-26
Dry order create

danac-gs:
# Description
There is a lot of code repetition when creating orders from pre-auth and non pre-auth flow.

Fixes: (link Asana Task)
https://app.asana.com/0/941920570700290/1129264891331567/f

# Type of change

# How Has This Been Tested?

# Checklist:


Helper/Order.php
```diff
@@ -930,20 +873,28 @@ public function prepareQuote($immutableQuote, $transaction)
...
 
         $email = @$transaction->order->cart->billing_address->email_address ?:
             @$transaction->order->cart->shipments[0]->shipping_address->email_address;
         $this->addCustomerDetails($quote, $email);
 
+        // Check if Mageplaza Gift Card data exist and apply it to the parent quote
+        $this->discountHelper->applyMageplazaDiscountToQuote($quote);
+
         $this->setPaymentMethod($quote);
+        $this->cartHelper->quoteResourceSave($quote);
```
YevhenShBolt: Why we should save quote every time when some data was added to the quote?

danac-gs: @YevhenShBolt some 3rd party plugins (e.g. Bronto) listen to quote after save events and collect various kind of analytics data. They usually expect those events in an ordered, funnel type, sequence, i.e. shipping address -> shipping method -> payment method. As we save all the info at the same time, we simulate the behavior by calling save in between. I know it is not optimal, honestly I am not 100% sure it is necessary, but I know is it was put that way when doing Bronto support and after that Bronto was not complaining anymore. It was there in non-preauth methods but not included in preauth so I included it.

YevhenShBolt: Thanks @danac-gs .

Hm interesting. I think it would be better to replace these saves: 
- or event 
- or move several methods together  to the separate methods and use plugin

instead of several save calls. 

Events or plugins give us ability to determine when we can calling additional save or not. And, as a result, these saves will not affect other merchants.

danac-gs: @YevhenShBolt
If I get it correctly, you recommend that instead of saving the quote multiple times, which triggers sales_quote_save_after event, we just dispatch the event (sales_quote_save_after), passing it the current state of the quote object, not yet persistent to the DB. And reduce the times we call save from 6 to 1. It is a good idea and I think it would work. @matt-thomason ?

YevhenShBolt: not exactly, i think that the event `sales_quote_save_after` too heavy also as a `save()`.

**1st way:**
instead of use `$this->cartHelper->quoteResourceSave($quote);` just need to apply
a method which implement:

```
public function isNeedSaves()
{
    // some logic for Bronto or something else
}
```
We can add custom event here or use plugin to extend our functionality.

**2nd way:**
move strings
```
$this->setShippingAddress($quote, $transaction);
$this->setBillingAddress($quote, $transaction);
```
to the separate method like `updateAddressData()` after that we can use plugin `after` for this method and decide does need to use additional save or not.

The same for other string:
```
$this->setShippingMethod($quote, $transaction);
```
move to `updateShippingMethod()` 

and etc.

after that we can use plugin `after` for these methods and decide does need to use additional save or not.

danac-gs: @YevhenShBolt I see you point and it makes sense. However, I do not agree completely.
1. `sales_quote_save_after` can not be as heavy as saving quote to the DB and than dispatching it.
2. The procedure you described is a good way to make our plugin extendable and easier to customize in the future. The same or similar as it is being done in M1 plugin. On the other hand, since we should support as much as possible 3rd parties out of the box, at least what we supported so far, I'd like to keep this events fired. Here or from the plugins to be implemented, but I'd keep them firing by default. Moreover, we are not sure yet if we would need additional custom events and plugins at this particular places. Until we actually need them maybe we should not add them (YAGNI).
@matt-thomason your thougths?
Ways to go:
1. replace saving the quote with dispatching the event
2. dispatch the event in an observers of a custom bolt events (that can be used to add more functionality), we would need ~5 events and handlers
3. use after plugins and dispatch the event there, 5 handlers again
4. do not dispatch the event by default, implement 2. and / or 3. and create a custom package for all merchants that use Bronto and ask them to install
5. use one of the options 1-3 but make the event dispatching configurable, public or via additional configuration json

YevhenShBolt: I vote for plugin implementation!

https://devdocs.magento.com/guides/v2.3/extension-dev-guide/plugins.html

By default we can disable our plugin customization for Bronto module and enable it only for projects where Bronto exists (or exist such a need).
```
<type name="{ObservedType}">
      <plugin name="{pluginName}" type="{PluginClassName}" sortOrder="1" disabled="false" />
</type>
```

We can easily marked `disabled="true"` and code would not be compiled in production mode. We do not impact any performance in merchant's system where it does not need. Regarding event dispatching we cannot say it with 100%. 

In this case with plugin, we will not have unexpected behavior of events, as we will know exactly where and when we called the event(s) or save. In the usual event dispatch, we really can not immediately find in which chain the change occurred.

danac-gs: @YevhenShBolt I don't remember who uses Bronto, we made it for one merchant but chances are others may use it as well. Moreover, there may be other plugins that use this event that we are not aware of. I vote for adding this event naturally in the code, replacing the save call. @matt-thomason I guess it's up to you.

matt-thomason: Sorry for the delay here. I read this over several times trying to form my opinion - you both have very good points here. This ties into a larger project we're trying to tackle this quarter, decoupling extension code from M2 (similar to how our Woo app accomplishes this), so it was top of mind. In fact, I'd love your opinion as well @mpatin.

I think both ideas are good. My thinking right now is to have a generic plugin (not a Bronto specific plugin, but a generic "we need to fire save events" plugin), that we can enable per customer. This allows us flexibility if we discover other plugins down the road that require this same event fired, and leaves it off the compiled code if merchants do not have a setup that requires it.

mpatin: I'm not sure I follow the exact tradeoffs and nuances to quotes within Magento; however, I'll weigh in at a higher level with some things to consider.

1. Ideally, for all of our platforms we will follow a model similar to the one we've taken with WooCommerce, having two plugins for merchants to install:
-- A "Core" plugin that contains almost all logic necessary to use Bolt - consistent across all merchants on the platform. This plugin will be updated on a regular (~bi-weekly) basis. 
-- An "Extension" plugin that contains customized logic for a very specific use case or new plugin support -- specific to a merchant. This is not expected to regularly need changes.

2. If we believe that the configuration in question (calling `sales_quote_save_after` event with every change) will be necessary for a significant number of merchants (>5), I believe we should include it into the "Core" plugin. 

3. I worry that if we begin to break up our code into separate code bases (e.g. one specifically for Bronto), we add more complexity to merchant integrations and places where merchants could have outdated or conflicting logic.

As mentioned above, could we add the configuration (firing this event at this frequency) as an option within plugin settings?

PR #346, opened 2019-07-31
Implement different success processing for pre-auth and non-pre-auth orders

danac-gs:
# Description
There is a redundant order/save call from the non pre-auth flow still present when pre-auth is turned on. It throws and error on HUF. Removing it fixes the issue.

Removing the call via configuration option, Pre-Auth order creation.

Fixes:
https://app.asana.com/0/941920570700290/1133661540833258/f

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/Config.php
+++ b/Helper/Config.php
@@ -850,7 +850,7 @@ class Config extends AbstractHelper
      *
      * @return  boolean
      */
-    public function getIsPreAUth($store = null)
+    public function getIsPreAuth($store = null)
     {
         return $this->getScopeConfig()->isSetFlag(
             self::XML_PATH_IS_PRE_AUTH,
--- a/Helper/Config.php
+++ b/Helper/Config.php
@@ -101,6 +101,11 @@ class Config extends AbstractHelper
     const XML_PATH_AUTOMATIC_CAPTURE_MODE = 'payment/boltpay/automatic_capture_mode';
 
     /**
+     * Is pre-auth
+     */
+    const XML_PATH_IS_PRE_AUTH = 'payment/boltpay/is_pre_auth';
+
+    /**
      * Prefetch shipping
      */
     const XML_PATH_PREFETCH_SHIPPING = 'payment/boltpay/prefetch_shipping';
@@ -836,6 +841,22 @@ class Config extends AbstractHelper
     {
         $config = $this->getAdditionalConfigObject($storeId);
         return @$config->$name;
+    }
+
+    /**
+     * Get Is Pre-Auth from config
+     *
+     * @param int|string|Store $store
+     *
+     * @return  boolean
+     */
+    public function getIsPreAUth($store = null)
+    {
+        return $this->getScopeConfig()->isSetFlag(
+            self::XML_PATH_IS_PRE_AUTH,
+            ScopeInterface::SCOPE_STORE,
+            $store
+        );
     }
 
     /**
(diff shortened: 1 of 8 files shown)


Helper/Config.php
```diff
@@ -838,6 +843,22 @@ private function getAdditionalConfigProperty($name, $storeId = null)
...
     }
 
+    /**
+     * Get Is Pre-Auth flag from config
+     *
+     * @param int|string|Store $store
+     *
+     * @return  boolean
+     */
+    public function getIsPreAUth($store = null)
```
matt-thomason: Weird capitalization here `AUth`

view/adminhtml/templates/boltpay/js/replacejs.phtml
```diff
@@ -376,21 +376,41 @@ require([
         },
 
         success: function (transaction, callback) {
+
+            /**
+             * Success transaction handler.
+             * Sets the success url for the non-preauth flow.
```
matt-thomason: Isn't this being called if `is_pre_auth` is true?

view/frontend/templates/js/replacejs.phtml
```diff
@@ -553,6 +553,34 @@ ob_start();
             },
 
             success: function (transaction, callback) {
+                /**
+                 * Success transaction handler.
+                 * Sets the success url for the non-preauth flow.
```
matt-thomason: Same here as well.

etc/config.xml
```diff
@@ -56,6 +56,7 @@ tr.shipping.totals td.amount span.price,
                 <bolt_order_caching>1</bolt_order_caching>
                 <api_emulate_session>1</api_emulate_session>
                 <should_minify_javascript>1</should_minify_javascript>
+                <is_pre_auth>0</is_pre_auth>
```
daisy1754: can we make it true by default?

view/adminhtml/templates/boltpay/js/replacejs.phtml
danac-gs: @matt-thomason 

If pre-auth is true then the redirect_url, which is created and returned in result of create order call, is implicitly opened from the checkout, when passed callback() is executed. And it redirects to the success page if passes validation.

In no pre-auth we need to call order/save, which returns the success url that we save as a property of the callbacks object. on the success order/save call, as the last step, after storing the success_url, the callback() is called. One of the things it does is closing the checkout modal. In the close event handler we check for the success_url and open the success page.

To sumarize:
pre-auth - checkout implicitly does the success page redirect
no pre-auth - the redirect is called explicitly in the plugin code

The later was already done in pre-auth and it was redundant.

etc/config.xml
danac-gs: @daisy1754

It is deliberately set as false. Two reasons:

1. When set to 0 It actually works both ways, pre-auth and no pre-auth. At least in most of the cases. If not for HUF we wouldn't notice it yet.

2. If the merchant updates the plugin on their own (as Capillus did today) without turning on the pre-auth flag and setting the create order endpoints we loose backward compatibility if the default is true.

daisy1754: okay fair

view/frontend/templates/js/replacejs.phtml
danac-gs: Same as above.

etc/config.xml
daisy1754: @matt-thomason let's make sure for new merchants we instruct to set this to true

Helper/Config.php
danac-gs: fixed

PR #348, opened 2019-08-02
Only collect total for the quote twice if the discount applies to shipping

johnmccombs1:
# Description
Only collect total for the quote twice if the discount applies to shipping
Fixes: https://app.asana.com/0/564264490825835/1126529642624345

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Model/Api/ShippingMethods.php
+++ b/Model/Api/ShippingMethods.php
@@ -41,6 +41,7 @@ use Bolt\Boltpay\Model\ErrorResponse as BoltErrorResponse;
 use Bolt\Boltpay\Helper\Session as SessionHelper;
 use Bolt\Boltpay\Exception\BoltException;
 use Bolt\Boltpay\Helper\Discount as DiscountHelper;
+use Magento\SalesRule\Model\RuleFactory;
 
 /**
  * Class ShippingMethods
@@ -147,6 +148,11 @@ class ShippingMethods implements ShippingMethodsInterface
     private $taxAdjusted = false;
 
     /**
+     * @var RuleFactory
+     */
+    private $ruleFactory;
+
+    /**
      * Assigns local references to global resources
      *
      * @param HookHelper                      $hookHelper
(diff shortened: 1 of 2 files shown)


Model/Api/ShippingMethods.php
```diff
@@ -743,6 +755,27 @@ public function getShippingOptions($quote, $addressData)
         return $shippingMethods;
     }
 
+    /**
+     * @param $quote
+     * @return bool
+     */
+    protected function doesDiscountApplyToShipping($quote){
+        if ($quote->getAppliedRuleIds()) {
```
oxigenka: Could I propose to define a variable for applied rules before `if` so that it can be used inside `foreach` later? It's only my opinion, but I believe it would help keep code readability and clearance.

danac-gs: Just removing the `if` condition would do.

oxigenka: > Just removing the `if` condition would do.

Since the `getAppliedRuleIds` method could return either string or `null`, there's the sense to check the value it returns, so it would be good to use `if` here to avoid an extra cast its value to the string.

Model/Api/ShippingMethods.php
```diff
@@ -634,12 +643,15 @@ public function getShippingOptions($quote, $addressData)
                 $quote->setCouponCode($appliedQuoteCouponCode)->collectTotals()->save();
             }
 
+            $this->totalsCollector->collectAddressTotals($quote, $shippingAddress);
+
             // In order to get correct shipping discounts the following method must be called twice.
```
matt-thomason: Floating the first call above the comment makes it a little confusing.

Model/Api/ShippingMethods.php
```diff
@@ -634,12 +643,15 @@ public function getShippingOptions($quote, $addressData)
...
             // Being a bug in Magento, or a bug in the tested store version, shipping discounts
             // are not collected the first time the method is called.
             // There was one loop step delay in applying discount to shipping options when method was called once.
```
ethanwayda22: Can we update the comments here to be more reflective of the new change?

PR #357, opened 2019-08-09
Refactor for extending

danac-gs:
# Description
The customization for the merchant has already been done for 1.x version but not in a preferable way. Now it's done via plugins, without overriding classes. In order to do so I needed to refactor 2 methods and extract some of their logic into a separate methods.

Fixes: (link Asana Task)
https://app.asana.com/0/0/1135091400919922/f

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -27,6 +27,7 @@ use Magento\Framework\App\Helper\AbstractHelper;
 use Magento\Framework\App\Helper\Context;
 use Magento\Framework\Exception\LocalizedException;
 use Magento\Quote\Model\Quote;
+use Magento\Quote\Model\Quote\Address\Total as AddressTotal;
 use Magento\Sales\Api\Data\OrderInterface;
 use Zend_Http_Client_Exception;
 use Bolt\Boltpay\Helper\Log as LogHelper;
@@ -184,8 +185,8 @@ class Cart extends AbstractHelper
     ];
     /////////////////////////////////////////////////////////////////////////////
 
-    // Totals adjustment treshold
-    private $treshold = 0.01;
+    // Totals adjustment threshold
+    private $threshold = 0.01;
 
     /**
      * @var array
@@ -540,7 +541,7 @@ class Cart extends AbstractHelper
     public function convertCustomAddressFieldsToCacheIdentifier($quote)
     {
         $customAddressFields = $this->getCustomAddressFieldsPascalCaseArray($quote->getStoreId());
-        $address = $quote->isVirtual() ? $quote->getBillingAddress() : $quote->getShippingAddress();
+        $address = $this->getCalculationAddress($quote);
 
         $cacheIdentifier = "";
 
@@ -1071,6 +1072,17 @@ class Cart extends AbstractHelper
     }
 
     /**
+     * Get the address for totals calculation based on the quote physical / virtual type
+     *
+     * @param Quote $quote
+     * @return QuoteAddress
+     */
+    protected function getCalculationAddress($quote)
+    {
+        return $quote->isVirtual() ? $quote->getBillingAddress() : $quote->getShippingAddress();
+    }
+
+    /**
      * Get cart data.
      * The reference of total methods: dev/tests/api-functional/testsuite/Magento/Quote/Api/CartTotalRepositoryTest.php
      *
@@ -1128,10 +1140,8 @@ class Cart extends AbstractHelper
             return $cart;
         }
 
+        $this->setLastImmutableQuote($immutableQuote);
         $immutableQuote->collectTotals();
-        $totals = $immutableQuote->getTotals();
-
-        $this->logHelper->addInfoLog('### CartTotals: ' . json_encode(array_keys($totals)));
 
         // Set order_reference to parent quote id.
         // This is the constraint field on Bolt side and this way
@@ -1198,7 +1208,7 @@ class Cart extends AbstractHelper
             $cart['billing_address'] = $cartBillingAddress;
         }
 
-        $address = $immutableQuote->isVirtual() ? $billingAddress : $shippingAddress;
+        $address = $this->getCalculationAddress($immutableQuote);
 
         // payment only checkout, include shipments, tax and grand total
         if ($paymentOnly) {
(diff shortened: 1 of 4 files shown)


Helper/Cart.php
```diff
@@ -1284,22 +1286,115 @@ public function getCartData($paymentOnly, $placeOrderPayload, $immutableQuote =
...
+                        'cart' => $cart,
+                    ]
+                ]);
+            });
+            $this->bugsnag->notifyError('Cart Totals Mismatch', "Totals adjusted by $diff.");
+        }
+
+        $this->sessionHelper->cacheFormKey($immutableQuote);
+
+        // $this->logHelper->addInfoLog(json_encode($cart, JSON_PRETTY_PRINT|JSON_UNESCAPED_SLASHES));
```
matt-thomason: Commented code here.

Helper/Cart.php
```diff
@@ -1284,22 +1286,115 @@ public function getCartData($paymentOnly, $placeOrderPayload, $immutableQuote =
...
+            // add the diff to first item total to pass bolt order create check
+            $cart['items'][0]['total_amount'] += round($diff);
+            $totalAmount += round($diff);
+        }
+        /////////////////////////////////////////////////////////////////////////////////
+
+        $cart['total_amount'] = $totalAmount;
+        $cart['tax_amount']   = $taxAmount;
+
+        if (abs($diff) >= $this->treshold) {
```
matt-thomason: I know this is old code but I just noticed the typo. Can you fix this to `$this->threshold`?

Helper/Cart.php
```diff
@@ -1357,14 +1453,15 @@ public function getCartData($paymentOnly, $placeOrderPayload, $immutableQuote =
...
-            $amount = $this->discountHelper->getMirasvitStoreCreditAmount($immutableQuote, $paymentOnly);
+            $amount = abs($this->discountHelper->getMirasvitStoreCreditAmount($immutableQuote, $paymentOnly));
             $roundedAmount = $this->getRoundAmount($amount);
             $cart['discounts'][] = [
                 'description' => 'Store Credit',
                 'amount'      => $roundedAmount,
                 'type'        => 'fixed_amount',
             ];
 
+            $diff -= $amount * 100 - $roundedAmount;
```
matt-thomason: Shouldn't this be `$diff = $amount * 100 - $roundedAmount;`? Same for the other diffs down below.

danac-gs: @matt-thomason for the items, shipping and tax prices this $diff collector increment is positive, for the discounts is negative. The same is with the $totalAmount aggregator.

Helper/Cart.php
danac-gs: @matt-thomason this is here for debugging purposes and always was. I turn it on / of during development. I will remove it if you say so.

Helper/Cart.php
danac-gs: Done.

Helper/Cart.php
matt-thomason: I don't love it but if it has a purpose then it's fine for this PR. We can perhaps revisit it when we're more generically cleaning up the plugin.

Helper/Cart.php
matt-thomason: Gotcha 👀

danac-gs:
started adding unit tests

Model/Api/ShippingMethods.php
```diff
@@ -310,27 +313,27 @@ public function getShippingMethods($cart, $shipping_address)
...
-            if (!$quote) {
+            if (!$this->quote) {
                 $this->throwUnknownQuoteIdException($quoteId);
             }
 
-            $this->preprocessHook($quote->getStoreId());
+            $this->preprocessHook($this->quote->getStoreId());
 
-            $this->checkCartItems($cart, $quote);
+            $this->checkCartItems($cart, $this->quote);
```
YevhenShBolt: Does it make sense to avoid here using `$this->quote`?
Can we just call `$this->quote` directly in protected method instead of transfer it through parameter?

The same for `$this->preprocessHook($this->quote->getStoreId());`

Model/Api/ShippingMethods.php
```diff
@@ -343,7 +346,7 @@ public function getShippingMethods($cart, $shipping_address)
 
             /** @var \Magento\Quote\Model\Quote $parentQuote */
             $parentQuote = $this->getQuoteById($cart['order_reference']);
-            if ($this->couponInvalidForShippingAddress($parentQuote->getCouponCode(), $quote)){
+            if ($this->couponInvalidForShippingAddress($parentQuote->getCouponCode(), $this->quote)){
```
YevhenShBolt: Can we avoid here transfer `$this->quote` parameter?
Can we safely call `$this->quote` directly in protected method?

Helper/Cart.php
```diff
@@ -1334,19 +1443,20 @@ public function getCartData($paymentOnly, $placeOrderPayload, $immutableQuote =
...
                     $roundedAmount = $this->getRoundAmount($amount);
 
-                    $cart['discounts'][] = [
+                    $discounts[] = [
                         'description' => 'Store Credit',
                         'amount'      => $roundedAmount,
                         'type'        => 'fixed_amount',
                     ];
 
+                    $diff -= $amount * 100 - $roundedAmount;
```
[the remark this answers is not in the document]
YevhenShBolt: Also, i would be glad if there appears unit tests which covered different variable types here (e.x `float`, `int` and etc).

CC: @danac-gs @matt-thomason

Helper/Cart.php
```diff
@@ -1290,19 +1300,118 @@ public function getCartData($paymentOnly, $placeOrderPayload, $immutableQuote =
...
+
+    /**
+     * Round amount helper
+     *
+     * @param $amount
+     * @return  int
+     */
+    public function getRoundAmount($amount)
+    {
+        return round($amount * 100);
```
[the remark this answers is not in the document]
danac-gs: I'll cast result to int.

Helper/Cart.php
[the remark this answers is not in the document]
danac-gs: No need for $amount and $roundedAmount to be of the same type. $roundedAmount is float with no decimal part, ~ integer. Will be explicitly casted to int per request above. @amount is some number, int or float. $diff can be 0 or some float. Later in the code, only if the accumulated $diff is over some threshold we take it into account. We don't check for equality so floating point precision can't get in our way.

Helper/Cart.php
```diff
@@ -1373,26 +1484,27 @@ public function getCartData($paymentOnly, $placeOrderPayload, $immutableQuote =
...
-            $cart['discounts'][] = [
+            $aheadStoreCredits = abs($this->discountHelper->getAheadworksStoreCredit($quote->getCustomerId()));
+            $roundedAmount = $this->getRoundAmount($aheadStoreCredits);
+            $discounts[] = [
                 'description' => 'Store Credit',
                 'amount'      => $roundedAmount,
                 'type'        => 'fixed_amount',
             ];
 
+            $diff -= $amount * 100 - $roundedAmount;
```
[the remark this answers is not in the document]
danac-gs: The same as above.

Model/Api/ShippingMethods.php
danac-gs: Makes sense

Model/Api/ShippingMethods.php
danac-gs: Makes sense

Helper/Cart.php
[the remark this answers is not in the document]
YevhenShBolt: @danac-gs  if we already have `$roundedAmount` as `int`.
Does it make sense to convert `$amount` to `int` also.
i mean something like that: `$diff -= (int) $amount * 100 - $roundedAmount;`

What the type return `$balanceModel->getAmount()` ?

As i remember correct about `abs()` we should have the following:
1. `$diff -= abs(12.3)*100 - 100;`
The result would be `float(1130)`
2. `$diff -= abs(123)*100 - 100;`
The result would be `int(12200)`

I do not see a problem here, and i am ok with such implementation, but if we work only on `int` values like `100` and `2235` without comma, i think it would be better to use only `int` types.

What do you think? ;-)

matt-thomason:
@danac-gs Can you rebase this on master - codecov is broken here but fixed on master HEAD.

PR #365, opened 2019-09-04
Merchant metrics v0

ethanwayda22:
# Description
This is the initial commit for capturing merchant metrics on all calls directed towards Bolt. It counts and times the latency for all these requests and pushes them to a file for batching. For the purposes of this initial push, I will be removing the batching aspect, posting the metrics after every call. Additionally, I created a feature flag so that no merchant should be directly affected by these functions.

Project: https://app.asana.com/0/1135083386077000/1135083386077000

# Type of change

# How Has This Been Tested?

(I have a flag set so we can begin testing on Sandbox servers and will not impact live merchants)

# Checklist:


Helper/MerchantMetrics.php
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    /**
+     * Based off the environment the project is ran in, determines the endpoint to add merchant metrics
+     *
+     * @return \GuzzleHttp\Client
+     */
+    protected function setClient() {
+        // determines if we are in the Sandbox env or not
+        $base_uri =  $this->configHelper->isSandboxModeSet() ? 'https://api-sandbox.bolt.com/' : 'https://api.bolt.com/' ;
```
danac-gs: Helper\Config::getApiUrl()

Helper/MerchantMetrics.php
```diff
@@ -0,0 +1,343 @@
...
+    public function addLatencyMetric($key, $value)
+    {
+        $data = [
+                'value' => $value,
+                "metric_type" => "latency",
+                "timestamp" => $this->getCurrentTime(),
+            ];
+        $this->metrics[$key] = $data;
+
+    }
```
danac-gs: The two methods above can be easily joined into one, avoiding repeating the code.

Controller/Cart/Data.php
```diff
@@ -144,14 +160,19 @@ public function execute()
                 'message' => $e->getMessage(),
                 'backUrl' => '',
             ]);
+            $latency = round(microtime(true) * 1000) - $startTime;
+            $this->merchantMetrics->processMetrics("order_token.failure", 1, "order_token.latency", $latency);
+            $this->merchantMetrics->postMetrics();
         } catch (Exception $e) {
-            $this->bugsnag->notifyException($e);
-
+            $this->bugsnag->notifyException($e);-
```
daisy1754: edit by mistake?

Helper/MerchantMetrics.php
```diff
@@ -0,0 +1,343 @@
...
+
+
+/**
+ * Boltpay Merchant Metrics wrapper helper
+ *
+ * @SuppressWarnings(PHPMD.TooManyFields)
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
+ */
+
+class MerchantMetrics extends AbstractHelper
```
daisy1754: *Merchant* prefix is redundunt as plugin is designed to run on merchant server on the first place. I would probably call it MetricsClient.

Helper/MerchantMetrics.php
```diff
@@ -0,0 +1,343 @@
...
+use Magento\Store\Model\StoreManagerInterface;
+use Bolt\Boltpay\Helper\Log as LogHelper;
+
+
+
+/**
+ * Boltpay Merchant Metrics wrapper helper
+ *
+ * @SuppressWarnings(PHPMD.TooManyFields)
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
```
daisy1754: are these valid warnings? Ideally we don't ignore warnings in new code or remove this rule all together via some config

Helper/MerchantMetrics.php
```diff
@@ -0,0 +1,343 @@
...
+    ) {
+        parent::__construct($context);
+
+        $this->storeManager = $storeManager;
+        $this->bugsnag = $bugsnag;
+        $this->logHelper = $logHelper;
+        //////////////////////////////////////////
+        // Composerless installation.
+        // Make sure libraries are in place:
+        // lib/internal/Bolt/bugsnag
```
daisy1754: looks like you're not autoloading bugsnag here?

Helper/MerchantMetrics.php
```diff
@@ -0,0 +1,343 @@
...
+        // lib/internal/Bolt/guzzle
+        //////////////////////////////////////////
+        if (!class_exists('\GuzzleHttp\Client')) {
+            // phpcs:ignore
+            require_once $directoryList->getPath('lib_internal') . '/Bolt/guzzle/autoloader.php';
+        }
+        $this->configHelper = $configHelper;
+        $this->metrics = array();
+        $this->metricsFile = null;
+        $this->guzzleClient = null;
```
daisy1754: why not initialize client here?

Helper/MerchantMetrics.php
```diff
@@ -0,0 +1,343 @@
...
+     * Based off the environment the project is ran in, determines the endpoint to add merchant metrics
+     *
+     * @return \GuzzleHttp\Client
+     */
+    protected function setClient() {
+        // determines if we are in the Sandbox env or not
+        $base_uri =  $this->configHelper->isSandboxModeSet() ? 'https://api-sandbox.bolt.com/' : 'https://api.bolt.com/' ;
+
+        // Creates a Guzzle Client and Headers needed for Metrics Requests
+        return new \GuzzleHttp\Client(['base_uri' => $base_uri]);
```
daisy1754: could we use sendRequest method in Helper/Api? There are existing boilerplate code to set API key, plugin version header, etc so ideally we should unify

Helper/MerchantMetrics.php
```diff
@@ -0,0 +1,343 @@
...
+            'x-api-key' =>  $this->configHelper->getApiKey()
+        ];
+    }
+
+    /**
+     * Fetches the current root directory for magento and creates a bolt metrics file in that location
+     *
+     * @return string
+     */
+    protected function setFile() {
```
daisy1754: does this actually create file? it looks like it just returns the path of file; if so let's rename to getFilePath

Helper/MerchantMetrics.php
```diff
@@ -0,0 +1,343 @@
...
+        //////////////////////////////////////////
+        if (!class_exists('\GuzzleHttp\Client')) {
+            // phpcs:ignore
+            require_once $directoryList->getPath('lib_internal') . '/Bolt/guzzle/autoloader.php';
+        }
+        $this->configHelper = $configHelper;
+        $this->metrics = array();
+        $this->metricsFile = null;
+        $this->guzzleClient = null;
+        $this->metricsFile = null;
```
daisy1754: why not initialize it here?

Helper/MerchantMetrics.php
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    /**
+     * Attempts to open a file and if it cannot open in 5 seconds it will return null
+     *
+     * @return Stream
+     */
+    public function waitForFile(){
+        $count = 0;
+        $timeoutMsecs = 10; //number of miliseconds of timeout (10 * 500) so 5 seconds
```
daisy1754: how about $maxRetryCount?

Helper/MerchantMetrics.php
```diff
@@ -0,0 +1,343 @@
...
+     * Adds a count and latency for a given event
+     *
+     * @param string        $countKey           name of count metric
+     * @param int           $countValue         the count value of the metric
+     * @param string        $latencyKey         name of latency metric
+     * @param int           $latencyValue  the total time of the metric
+     *
+     * @return void
+     */
+    public function processMetrics($countKey, $countValue, $latencyKey, $latencyValue)
```
daisy1754: hm, why not expose two public methods, processCount and processLatency?

daisy1754: I think it's too strong assumption that you always track both metrics and count.

Helper/MerchantMetrics.php
```diff
@@ -0,0 +1,343 @@
...
+    /**
+     * Post Metrics Collected in File to Merchant Metrics Endpoint, returning a 200 response if successful
+     *
+     *
+     * @return int
+     */
+    public function postMetrics()
+    {
+        // logic for properly grabbing file and locking it
+        if ($this->configHelper->shouldCaptureMerchantMetrics()) {
```
daisy1754: in general I recommend return early for readability. ie.,

```
if (!$this->configHelper->shouldCaptureMerchantMetrics()) {
  return;
}
```

Helper/MerchantMetrics.php
```diff
@@ -0,0 +1,343 @@
...
+        if ($test) {
+            $this->addCountMetric($countKey, $countValue);
+            $this->addLatencyMetric($latencyKey, $latencyValue);
+
+            if ($this->metricsFile == null) {
+                $this->metricsFile = $this->setFile();
+            }
+            $workingFile = $this->waitForFile();
+            if ($workingFile) {
+                file_put_contents($this->metricsFile, json_encode($this->metrics), FILE_APPEND);
```
daisy1754: I don't think using member variable metrics here is a good idea. suppose there are two thread executing processMetrics, you could be writing this content into file twice? you use file system to lock things but member variable live independent of file system.

I think `addCountMetric` and `addLatencyMetric` should be renamed to `formatCountMetric` which just returns php object, and in this processMetrics method we maintain list of metrics to write.

Model/Api/CreateOrder.php
```diff
@@ -171,7 +180,7 @@ public function execute(
         $currency = null
     ) {
         try {
-
+            $startTime = round(microtime(true) * 1000);
```
daisy1754: up to you but we could have simple timer class here for better redability. Timer would have two method, startTimer() and getElapsedTime()

Helper/MerchantMetrics.php
ethanwayda22: I did originally do that, however, I found this hard to mock for unit tests due to it not being passed in the constructor either.

Helper/MerchantMetrics.php
ethanwayda22: I spoke to @matt-thomason about this at first and he mentioned that we would be transitioning over to guzzle and this would be a great first use case for it. What are your thoughts, Matt?

Helper/MerchantMetrics.php
ethanwayda22: Again here is for mocking, any suggestions to avoid?

Helper/MerchantMetrics.php
ethanwayda22: I added this logic and split the dependency of count and latency. Is it worth keeping the combined one as well to reduce the amount of file writes we are doing? That is the one benefit there.

Model/Api/CreateOrder.php
ethanwayda22: Adding a timer and a metric class. I am putting them in the helper function classes. Do we have a separate directory I should be adding these in?

Helper/MetricsClient.php
```diff
@@ -0,0 +1,364 @@
...
+use Bolt\Boltpay\Helper\Log as LogHelper;
+
+
+/**
+ * Boltpay Metric Client Helper
+ */
+
+class MetricsClient extends AbstractHelper
+{
+    const STAGE_DEVELOPMENT = 'development';
```
roopakv: these seem unused

Helper/Timer.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+class Timer
+{
+    /**
+     * @var int
+     */
+    protected $startTime;
+
+    function __construct() {
+        $this->startTime = round(microtime(true) * 1000);
```
danac-gs: DRY, you can call $this->startTimer() here

Controller/Adminhtml/Cart/Data.php
```diff
@@ -94,13 +104,23 @@ public function __construct(
...
                 $responseData = json_decode(json_encode($boltpayOrder->getResponse()), true);
+                $latency = $timer->getElapsedTime();
+                $this->metricsClient->processCountMetric("back_office_order_token.success", 1);
+                $this->metricsClient->processLatencyMetric("back_office_order_token.latency", $latency);
+                $this->metricsClient->postMetrics();
+            } else {
+                $latency = $timer->getElapsedTime();
+                $this->metricsClient->processCountMetric("back_office_order_token.failure", 1);
+                $this->metricsClient->processLatencyMetric("back_office_order_token.latency", $latency);
+                $this->metricsClient->postMetrics();
```
danac-gs: This 4 line code pattern keeps repeating. Is it reasonable to wrap it into a single method call?

Helper/Timer.php
```diff
@@ -0,0 +1,52 @@
...
+ * @package    Bolt_Boltpay
+ * @copyright  Copyright (c) 2018 Bolt Financial, Inc (https://www.bolt.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+
+namespace Bolt\Boltpay\Helper;
+
+/**
+ * Timer class used to measure latency of requests in ms
```
daisy1754: comment doesn't have to mention specific use case. this class could be used for anything

Helper/Timer.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+class Timer
+{
+    /**
+     * @var int
+     */
+    protected $startTime;
```
daisy1754: private?

Helper/Timer.php
```diff
@@ -0,0 +1,52 @@
...
+    function __construct() {
+        $this->startTimer();
+    }
+
+    /**
+     * Restarts the timer to the current time
+     *
+     * @return void
+     */
+    public function startTimer() {
```
daisy1754: nit: this is Timer class so method can be simply named `start`. `$timer->start()` makes sense.

Helper/Timer.php
```diff
@@ -0,0 +1,52 @@
...
+        $this->startTime = round(microtime(true) * 1000);
```
daisy1754: nit: you can hold startTime as micro and only round in `getElapsedTime`

Model/Api/CreateOrder.php
daisy1754: Helper directory seems good

Helper/MerchantMetrics.php
daisy1754: I start to wonder if you even need count metrics. you can just use count of latency metrics if you always report both count and metrics at the same location. I suggest you start to build sample dashboard in datadog and talk with merchant exp team to figure out what metrics are valuable. 

For scope of this PR, I'm ok with creating utility function to report both count and metrics but as I said above reporting only latency should probably just work fine in that case.

Helper/MerchantMetrics.php
daisy1754: I don't know how magento's DI works but in general you can workaround that kind of issue by injecting factory class. Factory class will look like below:

```
class GuzzleClientFactory {
   public getInstance() {
    return new GuzzleClient();
  }
}
```

In test you inject MockClientFactory which returns mock client.

Helper/Metric.php
```diff
@@ -0,0 +1,64 @@
...
+ * @copyright  Copyright (c) 2018 Bolt Financial, Inc (https://www.bolt.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+namespace Bolt\Boltpay\Helper;
+
+/**
+ * Boltpay Metrics used by the Metric Client
+ */
+class Metric implements \JsonSerializable
```
daisy1754: I wonder if we should define this in MetricsClient.php file. It seems it's kind of implementation detail of MetricsClient

PR #372, opened 2019-09-12
validateItemPrice fix - multiple items with the same SKU and different prices, (multivariant product)

danac-gs:
# Description
validateItemPrice fix for the case with multiple items with same SKU and different prices, (multivariant product)

Fixes:
https://app.asana.com/0/474713171217346/1139465075665305/f

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Model/Api/CreateOrder.php
+++ b/Model/Api/CreateOrder.php
@@ -558,30 +558,32 @@ class CreateOrder implements CreateOrderInterface
      * @param array     $transactionItems
      * @throws BoltException
      */
-    public function validateItemPrice($itemSku, $itemPrice, $transactionItems)
+    public function validateItemPrice($itemSku, $itemPrice, &$transactionItems)
     {
-        foreach ($transactionItems as $transactionItem) {
+        foreach ($transactionItems as $index => $transactionItem) {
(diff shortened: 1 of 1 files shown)


Model/Api/CreateOrder.php
```diff
@@ -558,30 +558,32 @@ public function hasItemErrors($quoteItem)
...
-                            'transaction.unit_price' => $transactionUnitPrice,
-                        ]
-                    ]);
-                });
-                throw new BoltException(
-                    __('Price does not match. Item sku: ' . $itemSku),
-                    null,
-                    self::E_BOLT_ITEM_PRICE_HAS_BEEN_UPDATED
-                );
+                unset ($transactionItems[$index]);
```
YevhenShBolt: Why we should unset some item?

danac-gs: @YevhenShBolt if we don't we could have false positive results. Here we match quote items to bolt transaction items. Theoretically two or more quote items with the same SKU and price (and some different options) could be matched to a single one item within the bolt transaction in this loop. We could solve this differently, for example add another loop and match them in the opposite direction, but this way it is more optimal and simpler.

PR #367, opened 2019-09-09
Remove state transition checks to support paypal

roopakv:
# Description
This PR removes state transition checks. The reason being that as we add more processors and introduce new state, keeping the state in check involves releasing a new version of the plugin making roll outs harder.

In this new flow, we will not check state transitions and rather implicitly trust the bolt modal.

Fixes: https://app.asana.com/0/1120201793092521/1135349256216762

# Type of change

- [ ] Bug fix (change which fixes an issue)
- [X] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?
Please validate that you have tested your change in at least one of the following areas:

- [ ] Successfully tested locally (or docker image)
- [X] Successfully tested on a staging or sandbox server
- [ ] Successfully tested on a merchant's staging server

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] New and existing unit tests pass locally with my changes
- [x] I have created or modified unit tests to sufficiently cover my changes.

PR #383, opened 2019-09-27
Revert "Remove state transition checks to support paypal (#367)"

roopakv:
This reverts commit 8320dc1410adc1b923884d1901ee49a7ab516e92.

We believe something about these state transition checks was necessary to prevent Bolt from handling state transitions from non-Bolt payment methods. We're rolling back the changes as we investigate the issues.

# Description

Fixes: (link Asana Task)

# Type of change

# How Has This Been Tested?

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] New and existing unit tests pass locally with my changes
- [X] I have created or modified unit tests to sufficiently cover my changes.

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -83,6 +83,56 @@ class Order extends AbstractHelper
     const BOLT_ORDER_STATUS_PENDING = 'bolt_pending';
     const MAGENTO_ORDER_STATUS_PENDING = 'pending';
 
+    // Posible transation state transitions
+    private $validStateTransitions = [
+        null => [
+            self::TS_ZERO_AMOUNT,
+            self::TS_PENDING,
+            self::TS_COMPLETED, // back office
+            // for historic data (order placed before plugin update) does not have "previous state"
+            self::TS_CREDIT_COMPLETED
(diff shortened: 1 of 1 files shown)


matt-thomason:
Context: We believe something about these state transition checks was necessary to prevent Bolt from handling state transitions from non-Bolt payment methods. We're rolling back the changes as we investigate the issues.

danac-gs:
@matt-thomason @roopakv there's a payment method check in Helper\Order::checkPaymentMethod(). Anyway, we need a (better) way to test paypal transactions and some time spent investigating what has really happened?

roopakv:
Agree. I spent a bit of time (albeit, i know very little about M2, so one of you might be able to talk better about this).

What im noticing is that we rely on state transitions to block some flows, rather than relying on it to just validate. It looks like this method is called for more than just Bolt flows and we sort of just "hope" they have an incorrect transition.

Would be nice to go thru and understand what cases and how this works.

PR #385, opened 2019-10-03
Ensure customer email is set to the quote before saving the order

johnmccombs1:
If we enable Automatic Assignment to Customer Group https://screencast.com/t/wL9IfyT627M , the quote customer email is set to null before creating the order. This ensures the email is set back to the quote before saving the order

Fixes: https://app.asana.com/0/564264490825835/1143022171544385

# Type of change

# How Has This Been Tested?

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -976,10 +976,6 @@ class Order extends AbstractHelper
         $this->setShippingMethod($quote, $transaction);
         $this->quoteAfterChange($quote);
 
-        $email = @$transaction->order->cart->billing_address->email_address ?:
-            @$transaction->order->cart->shipments[0]->shipping_address->email_address;
-        $this->addCustomerDetails($quote, $email);
-
         // Check if Mageplaza Gift Card data exist and apply it to the parent quote
         $this->discountHelper->applyMageplazaDiscountToQuote($quote);
 
@@ -994,6 +990,10 @@ class Order extends AbstractHelper
                 'cc_type' => @$transaction->from_credit_card->network
             ]
         );
+
+        $email = @$transaction->order->cart->billing_address->email_address ?:
+            @$transaction->order->cart->shipments[0]->shipping_address->email_address;
+        $this->addCustomerDetails($quote, $email);
 
         $quote->setReservedOrderId($quote->getBoltReservedOrderId());
         $this->cartHelper->quoteResourceSave($quote);


Helper/Order.php
```diff
@@ -995,6 +991,10 @@ public function prepareQuote($immutableQuote, $transaction)
             ]
         );
 
+        $email = @$transaction->order->cart->billing_address->email_address ?:
+            @$transaction->order->cart->shipments[0]->shipping_address->email_address;
+        $this->addCustomerDetails($quote, $email);
+
```
YevhenShBolt: @johnmccombs1, What is the difference between moving these lines here or before the current lines (see below part of code) or to some other place?:
```
        $this->quoteAfterChange($quote);
```
???

Maybe better to add additional check for an email for checking is empty email or customer groups or something like that?


As i see, the config "auto_group_assign" use only in:
`\Magento\Customer\Model\Vat::getCustomerGroupIdBasedOnVatNumber()`

And it depends from addresses (billing and shipping), as i see next, that the `getCustomerGroupIdBasedOnVatNumber` only calling twice in:
1. `\Magento\Quote\Observer\Frontend\Quote\Address\CollectTotalsObserver::execute` - an event `sales_quote_address_collect_totals_before`
2. `\Magento\Customer\Observer\AfterAddressSaveObserver::execute` - an event - `customer_address_save_after`


So, I dare to assume that only during save or collect totals we should be aware of an email and assignment of customer groups. 

My suggestion is to add additional check for email to be sure that we have correct email always (if needed), and i do not see any reason to move this part of code yet.

YevhenShBolt: One reason to move this code before save is avoid calling `$this->quoteAfterChange($quote);`

so it means that we insert too many dispatches of event `sales_quote_save_after` and needs to refactor the whole logic. 
Or another way to find in code side which 3rd party plugin erases our data from email field.

danac-gs:
@johnmccombs1 @YevhenShBolt @matt-thomason please resolve this so it can go into 2.0.3. I think it's good enough to be merged as it currently is.

Helper/Order.php
johnmccombs1: @YevhenShBolt , thanks for the feedback. 

The reason I moved addCustomerDetails method call to after setPaymentMethod method call is that the method setPaymentMethod collects the total and then set the email to null if we enable Automatic Assignment to Customer Group. 

I confirm this isn't a 3rd party plugin issue. 

FYI, method addCustomerDetails is still called before the quote is saved here https://github.com/BoltApp/bolt-magento2/blob/master/Helper/Order.php#L999 and I don't think the change that I made dispatches any sales_quote_save_after event. It just sets the email for the quote after setting the payment method. 

Am I missing something here?

PR #386, opened 2019-10-04
Skip inventory verification in post checkout events

danac-gs:
# Description
Skip inventory verification in success redirect call and hooks. At this point the order is already created and inventory reserved.

Magento core team intended the checkout_submit_all_after  event and $this->quoteManagement->submit($quote, $orderData) to be raised/called in the same execution.

Quote management submit method raises sales_model_service_quote_submit_before, which executes \Magento\CatalogInventory\Observer\SubtractQuoteInventoryObserver.

Event checkout_submit_all_after calls \Magento\CatalogInventory\Observer\CheckoutAllSubmitAfterObserver which, in turn calls \Magento\CatalogInventory\Observer\SubtractQuoteInventoryObserver.

Magento core team avoided  this by using inventory_processed data property on Quote object (if set, skip observer execution). Because we separated the two calls and because this property is not persistent in the quote table the inventory subtraction logic was executed twice.

Fixes: https://app.asana.com/0/474713171217346/1142870009972713/f

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -761,6 +761,9 @@ class Order extends AbstractHelper
 
         $this->applyExternalQuoteData($quote);
 
+        // do not verify inventory, it is already reserved
+        $quote->setInventoryProcessed(true);
+
         $this->logHelper->addInfoLog('[-= dispatchPostCheckoutEvents =-]');
         $this->_eventManager->dispatch(
             'checkout_submit_all_after', [


danac-gs:
This bug has more impact then reported. There need to be twice as much items in inventory than ordered for orders to pass. For newer Magento versions it is just the verification issue, inventory is not affected. For older ones (Taos,  M 2.1.9) the inventory gets decreased twice.

matt-thomason:
Thanks for the detailed explanation @danac-gs.

PR #391, opened 2019-10-10
Support Bss StoreCredit plugin

oldDriver:
# Description
Support Bss StoreCredit plugin

Fixes: https://app.asana.com/0/0/1142368638742826/f

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/Discount.php
+++ b/Helper/Discount.php
@@ -50,6 +50,8 @@ class Discount extends AbstractHelper
     const MAGEPLAZA_GIFTCARD_QUOTE_KEY = 'mp_gift_cards';
     const AHEADWORKS_STORE_CREDIT = 'aw_store_credit';
     const BSS_STORE_CREDIT = 'bss_storecredit';
+    const AMASTY_COUPON_AMOUNT = 'amasty_coupon_amount';
+    const AHEADWORKS_GIFTCARD = 'aw_giftcard';
 
     /**
      * @var ResourceConnection $resource
--- a/Helper/Discount.php
+++ b/Helper/Discount.php
@@ -49,6 +49,7 @@ class Discount extends AbstractHelper
     const MAGEPLAZA_GIFTCARD = 'gift_card';
     const MAGEPLAZA_GIFTCARD_QUOTE_KEY = 'mp_gift_cards';
     const AHEADWORKS_STORE_CREDIT = 'aw_store_credit';
+    const BSS_STORE_CREDIT = 'bss_storecredit';
 
     /**
      * @var ResourceConnection $resource
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -182,7 +182,9 @@ class Cart extends AbstractHelper
         Discount::AMASTY_GIFTCARD => 'Gift Card ',
         Discount::GIFT_VOUCHER => '',
         Discount::MAGEPLAZA_GIFTCARD => '',
-        Discount::BSS_STORE_CREDIT => ''
+        Discount::BSS_STORE_CREDIT => '',
+        Discount::AMASTY_COUPON_AMOUNT => '',
+        Discount::AHEADWORKS_GIFTCARD => ''
     ];
     /////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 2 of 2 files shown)


matt-thomason:
@oldDriver I am not referring to your test phrase, I'm asking if you're actually tested it with a merchant's BSS storecredit plugin. This does not look like it would work to me.

oldDriver:
@matt-thomason i tested this with @eugene-shab on my local instance and on the Merchant's stage server

danac-gs:
LGTM so far. @matt-thomason I believe that if Store Credit amount is added to the quote totals then this approach might actually work. In most of the cases. @oldDriver However, consider the case when the customer Store Credit total available balance is greater than the current shopping cart items subtotal. If store credit can be used to pay for shipping and / or tax then the total available amount should be passed to cart as a fixed_amount type discount. This can not be read from the session / quote / totals but you need to fetch it from some customer / store credit resource. Similar as the support for other store credit / reward point modules. If, on the other hand, this store credit can only be used to pay for the cart items, this solution may be sufficient.

YevhenShBolt:
@matt-thomason ,
As i see all calculation of discount happen in

https://github.com/BoltApp/bolt-magento2/blob/ed59919460aefc2bbd4c2fe0890eeee51ad6729f/Helper/Cart.php#L1546

We always add new total type in such way (you can see it on example of AMASTY_GIFTCARD and etc). 

Previously we have a special method `$this->getDiscountTypes()`which returned a private variable  `$this->discountTypes` and have easy to extend through Magento plugin logic (see before plugin), but now we only left private variable.

So, in that way we can only adds items to the `discountTypes` array.

Also, often we add additional logic to the `Helper/Cart` as describe Daniel.

In this helper a lot of example like AMASTY_GIFTCARD, MAGEPLAZA_GIFTCARD, UNIRGY_GIFT_CERT and etc.

danac-gs:
@YevhenShBolt @oldDriver please check how other Store Credit / Reward Point modules are supported as fixed_amount discounts:
https://github.com/BoltApp/bolt-magento2/blob/ed59919460aefc2bbd4c2fe0890eeee51ad6729f/Helper/Cart.php#L1428
https://github.com/BoltApp/bolt-magento2/blob/ed59919460aefc2bbd4c2fe0890eeee51ad6729f/Helper/Cart.php#L1469
https://github.com/BoltApp/bolt-magento2/blob/ed59919460aefc2bbd4c2fe0890eeee51ad6729f/Helper/Cart.php#L1503
To test your solution create a cart with subtotal less then a customer Store Credit balance and shipping and tax greater than 0.

oldDriver:
Last changes are needed for Ambush\BuyWake

Helper/Discount.php
```diff
@@ -49,6 +49,9 @@ class Discount extends AbstractHelper
     const MAGEPLAZA_GIFTCARD = 'gift_card';
     const MAGEPLAZA_GIFTCARD_QUOTE_KEY = 'mp_gift_cards';
     const AHEADWORKS_STORE_CREDIT = 'aw_store_credit';
+    const BSS_STORE_CREDIT = 'bss_storecredit';
+    const AMASTY_COUPON_AMOUNT = 'amasty_coupon_amount';
+    const AHEADWORKS_GIFTCARD = 'aw_giftcard';
```
roopakv: it seems like previously we already added `AHEADWORKS_STORE_CREDIT` is there anyway we can find all `aheadworks` discounts and add them now? same for amasty. we should just be able to whitelist another module once and for all as a whole?

danac-gs: @roopakv those are different modules of the same vendor. Support for a module can be a not so trivial task and so far it was done on demand (when a merchant needs it we build it into our plugin as an integral part).
@oldDriver did you test the case I explained above with all of the newly added discounts?

roopakv: @danac-gs thank you for explaining

oldDriver: @roopakv as for me better way to change property $discountTypes from "private" to "protected"
I this way we will not need to add all gift/discounts/coupons/store credit/voucher in the Helper\Cart.
We could simply add Botpay\Custom\Helper\Cart:
namespace Bolt\Custom\Helper
use Bolt\Boltpay\Helper\Cart as BaseCart;
class Cart extends BaseCart
{
    protected $discountTypes = [
        // Types we need in the current integration
    ];
}
Does it make sense?

roopakv: sort of. not following a 100% but i think i get the idea

matt-thomason:
Closing this as it was addressed in #420

PR #396, opened 2019-10-15
[PFS][Part 1] Introduce feature switch repository

roopakv:
This PR introduces the feature switch repository. Nothing uses this repo yet. This repo has the capability to save, read from and delete from the table. Most of the code is boilerplate code. look at feature switch repository which has the only bit of logic.

If trying to run this locally, please make sure to compile again.

Note: Magento2 strictly requires class names to be in full and not imported using `use` when dealing with repositories in order to generate the factories. This is non negotiable, and hence the code might not look exactly how we want it to. 

# Description

Fixes: Part of https://app.asana.com/0/1141250355664307/1144434411596969

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Model/FeatureSwitch.php
+++ b/Model/FeatureSwitch.php
@@ -1,0 +1,61 @@
+<?php
+
+
+namespace Bolt\Boltpay\Model;
+
+use Magento\Framework\Model\AbstractModel;
+use Bolt\Boltpay\Api\Data\FeatureSwitchInterface;
+
+class FeatureSwitch extends AbstractModel implements \Bolt\Boltpay\Api\Data\FeatureSwitchInterface
+{
+    const NAME = 'switch_name';
+    const VALUE = 'switch_value';
+    const DEFAULT_VALUE = 'default_value';
+    const ROLLOUT_PERCENTAGE = 'rollout_percentage';
+
+    protected function _construct()
+    {
+        parent::_construct();
+        $this->_init(ResourceModel\FeatureSwitch::class);
+    }
+
+    public function getName()
+    {
+        return $this->_getData(self::NAME);
+    }
+
+    public function setName($name)
+    {
+        $this->setData(self::NAME, $name);
+    }
+
+    public function getValue()
+    {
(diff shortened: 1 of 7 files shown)


Model/FeatureSwitch.php
```diff
@@ -0,0 +1,61 @@
...
+    const DEFAULT_VALUE = 'default_value';
+    const ROLLOUT_PERCENTAGE = 'rollout_percentage';
+
+    protected function _construct()
+    {
+        parent::_construct();
+        $this->_init(ResourceModel\FeatureSwitch::class);
+    }
+
+    public function getName()
```
daisy1754: my understanding is you don't have to define these getters/setters because magento has magic method (see "Basic Model Operations" section https://devdocs.magento.com/guides/m1x/magefordev/mage-for-dev-5.html) - yay dynamic language

roopakv: it was pretty hard to get here though :P their documentation though rich didnt add a sample for for this

Model/FeatureSwitchRepository.php
```diff
@@ -0,0 +1,37 @@
...
+
+    public function __construct(
+        FeatureSwitchFactory $featureSwitchFactory
+    ) {
+        $this->featureSwitchFactory = $featureSwitchFactory;
+    }
+
+    public function getByName($name) {
+        $switch = $this->featureSwitchFactory->create();
+        $switch->getResource()->load($switch, $name, "switch_name");
```
matt-thomason: I would use the `switch_name` enum you defined in `FeatureSwitch`.

Model/FeatureSwitch.php
roopakv: ok i understand how this is supposed to work but also thoughts on making it explicit?

this will make the method named getSwitchName() etc, which is fine, but i sort of like this typing. going ot let you and @matt-thomason decide

roopakv: discussed offline, that was M1 and M2 doesnt support this.

PR #398, opened 2019-10-15
[PFS][Part 3]  Get from bolt and store in DB

roopakv:
Get feature switches from bolt and store in DB.

The next PR in this series will call this from the API hook and from upgrade data.

Fixes: https://app.asana.com/0/1141250355664307/1144434411596969

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/FeatureSwitch/Manager.php
+++ b/Helper/FeatureSwitch/Manager.php
@@ -63,13 +63,7 @@ class Manager extends AbstractHelper {
         $switches = array();
         if ($switchesResponse->getData()) {
             $data = $switchesResponse->getData();
-            if (isset($data["response"])
-                && isset($data["response"]->data)
-                && isset($data["response"]->data->plugin)
-                && isset($data["response"]->data->plugin->features)
-            ) {
-                $switches = $data["response"]->data->plugin->features;
-            }
+            $switches = @$data["response"]->data->plugin->features;
         }
 
         if (!is_array($switches) || count($switches) == 0) {
--- a/Helper/FeatureSwitch/Manager.php
+++ b/Helper/FeatureSwitch/Manager.php
@@ -1,0 +1,90 @@
+<?php
+/**
+ * Bolt magento2 plugin
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ *
+ * @category   Bolt
+ * @package    Bolt_Boltpay
+ * @copyright  Copyright (c) 2018 Bolt Financial, Inc (https://www.bolt.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+namespace Bolt\Boltpay\Helper\FeatureSwitch;
+
+
+use Magento\Framework\App\Helper\Context;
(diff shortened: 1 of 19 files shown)


Helper/FeatureSwitch/Manager.php
```diff
@@ -0,0 +1,90 @@
...
+
+        $switches = array();
+        if ($switchesResponse->getData()) {
+            $data = $switchesResponse->getData();
+            if (isset($data["response"])
+                && isset($data["response"]->data)
+                && isset($data["response"]->data->plugin)
+                && isset($data["response"]->data->plugin->features)
+            ) {
+                $switches = $data["response"]->data->plugin->features;
```
daisy1754: you could also do

```
@$switches = $data["response"]->data->plugin->features;
```

it will work like a optional chain (@ is error control operator https://www.php.net/manual/en/language.operators.errorcontrol.php )

Helper/FeatureSwitch/Manager.php
```diff
@@ -0,0 +1,90 @@
...
+     *
+     * @throws \Magento\Framework\Exception\LocalizedException
+     */
+    public function updateSwitchesFromBolt() {
+        $switchesResponse = $this->gql->getFeatureSwitches();
+
+        $switches = array();
+        if ($switchesResponse->getData()) {
+            $data = $switchesResponse->getData();
+            if (isset($data["response"])
```
daisy1754: IMO this is redundunt, I think it's ok to assume some contract (response has specific format in case of successful response)

Helper/FeatureSwitch/Manager.php
matt-thomason: `@` only works on expressions, need it on the RHS if you want to do this approach. This is probably fine - you are checking on `switches` directly after this.
```
$switches = @$data["response"]->data->plugin->features;
```

Helper/FeatureSwitch/Manager.php
```diff
@@ -0,0 +1,90 @@
...
+                $switches = $data["response"]->data->plugin->features;
+            }
+        }
+
+        if (!is_array($switches) || count($switches) == 0) {
+            return;
+        }
+
+        foreach ($switches as $fs) {
+            if (isset($fs->name)) {
```
matt-thomason: While this check is safe, why have it? All 4 values are necessary for `upsertByName`, right? When would a name be missing but not the other 3 values?

Helper/FeatureSwitch/Manager.php
roopakv: WOW. apparently this is the stfu operator! LOVE IT

Helper/FeatureSwitch/Manager.php
roopakv: i sort of agree, but also think would make our code bullet proof just incase someone by mistake changes bolt then we have to revert. this just makes us bullet proof and saves us from random errors. now given that we have the stfu operator id rather it be this way. please unresolve if you care a lot about this

Helper/FeatureSwitch/Manager.php
roopakv: TBH probably overkill. I am trying hard to make sure the code will not barf by checking everything. however value and default value are booleans so unset they are false. that said i could also check roll out %age.

let me know what you decide

matt-thomason: That's prooobably fine, and I suppose if name is present then %age likely will too. I thought raising this to get your thoughts was worth it, but this doesn't block me.

roopakv:
@matt-thomason @daisy1754 lost my approvals since i changed branch to master due to stacked commits.

PR #400, opened 2019-10-15
Fixed validation of capture amount by using correct rounding algorithm

oxigenka:
# Description
I faced a rounding issue in the validation method when working on merchant integration and saw an exception was raised. Debug let me see the values of an order object might not be rounded.

---------
Example 1.

As of the previous behavior, we had:

```
$captureAmount: 203.29
$order->getTotalInvoiced(): [empty]
$order->getGrandTotal(): 203.2874

    203.29 + 0 > 203.2874 [ = true ] -> leads to raise an exception
```

This way, an exception is raised here. To avoid it, the same rounding algorithm was applied on each of operands and now it works well. The webhook was failed but it shouldn't. After the fix has been applied the webhook succeeded.

Now it works as it should:

```
getRoundAmount($order->getTotalInvoiced()) [0] + getRoundAmount($captureAmount) [20329]: 20329
getRoundAmount($order->getGrandTotal()): 20329

    20329 > 20329 [ = false ] -> validation passed
```

---------
Example 2.

```
$captureAmount: 203.49
$order->getTotalInvoiced(): 2.49
$order->getGrandTotal(): 205.9877

    203.49 + 2.49 ( = 205.98 ) > 205.9877 [ = false ] -> validation passed
```

Applying rounding, we got:

```
getRoundAmount($order->getTotalInvoiced()) [249] + getRoundAmount($captureAmount) [20349]: 20598
getRoundAmount($order->getGrandTotal()): 20599

    20598 > 20599 [ = false ] -> validation also passed
```
---
There was no Asana task created for this issue.

# Type of change

# How Has This Been Tested?

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -1719,7 +1719,9 @@ class Order extends AbstractHelper
      */
     protected function validateCaptureAmount(OrderInterface $order, $captureAmount)
     {
-        $isInvalidAmount = !isset($captureAmount) || !is_numeric($captureAmount) || $captureAmount < 0;
+        if (!isset($captureAmount) || !is_numeric($captureAmount) || $captureAmount < 0) {
+            throw new \Exception( __('Capture amount is invalid'));
+        }
 
         /**
          * Due to grand total sent to Bolt is rounded, the same operation should be used
@@ -1727,12 +1729,14 @@ class Order extends AbstractHelper
          * Rounding operations are applied to each operand in order to avoid cases when the grand total
          * is formally less (before it has been rounded) than the sum of the captured amount and the total invoiced.
          */
-        $isInvalidAmountRange = $this->cartHelper->getRoundAmount($order->getTotalInvoiced())
-            + $this->cartHelper->getRoundAmount($captureAmount)
-            > $this->cartHelper->getRoundAmount($order->getGrandTotal());
+        $captured = $this->cartHelper->getRoundAmount($order->getTotalInvoiced())
+            + $this->cartHelper->getRoundAmount($captureAmount);
+        $grandTotal = $this->cartHelper->getRoundAmount($order->getGrandTotal());
 
-        if ($isInvalidAmount || $isInvalidAmountRange) {
-            throw new \Exception( __('Capture amount is invalid'));
+        if ($captured > $grandTotal) {
+            throw new \Exception(
+                __('Capture amount is invalid: captured [%s], grand total [%s]', $captured, $grandTotal)
+            );
         }
     }
 
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -1654,45 +1654,10 @@ class Order extends AbstractHelper
         $isInvalidAmount = !isset($captureAmount) || !is_numeric($captureAmount) || $captureAmount < 0;
 
         /**
-         * Due to grand total sent to Bolt is rounded,
-         * the same operation should be used when validating captured amount.
-         *
-         * The rounding operations are applied to each operand to avoid cases when grand total
-         * is formally lower than the sum of captured amount and total invoiced.
-         *
-         * ---------
-         * Example 1.
-         *
-         * As of the previous behavior, we had:
-         *
-         * $captureAmount: 203.29
-         * $order->getTotalInvoiced(): [empty]
-         * $order->getGrandTotal(): 203.2874
-         *
-         *      203.29 > 203.2874 [ = true ] -> leads to raise an exception
-         *
-         * After applying rounding, got:
-         *
-         * getRoundAmount($order->getTotalInvoiced()) [0] + getRoundAmount($captureAmount) [20329]: 20329
-         * getRoundAmount($order->getGrandTotal()): 20329
-         *
-         *      20329 > 20329 [ = false ] -> validation passed
-         *
-         * ---------
-         * Example 2.
-         *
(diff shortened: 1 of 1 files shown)


Helper/Order.php
```diff
@@ -1644,15 +1644,20 @@ protected function isCaptureHookRequest($newCapture)
...
      * @throws \Exception
      */
-    protected function validateCaptureAmount(OrderInterface $order, $captureAmount) {
+    protected function validateCaptureAmount(OrderInterface $order, $captureAmount)
+    {
         $isInvalidAmount = !isset($captureAmount) || !is_numeric($captureAmount) || $captureAmount < 0;
-        $isInvalidAmountRange = $order->getTotalInvoiced() + $captureAmount > $order->getGrandTotal();
 
-        if($isInvalidAmount || $isInvalidAmountRange) {
+        $isInvalidAmountRange = $this->cartHelper->getRoundAmount($order->getTotalInvoiced())
```
roopakv: a case where this can fail even with this fix.

$captureAmount: 203.29
$order->getTotalInvoiced(): 
$order->getGrandTotal(): 203.2844
$captureAmount + $order->getTotalInvoiced(): 20329 > 20328 -> leads to raise an exception

roopakv: also fwiw i think it might be better tp add and then round to lose as little precision as possible?

oxigenka: I believe in the case you've provided where `$order->getGrandTotal(): 203.2844` the captured amount will be 203.28, not 203.29. It's because of the operation of rounding will be done before the order information (including amounts) would be sent to Bolt. Then, when Bolt calls the order management webhook, it will be `$captureAmount: 203.28`

roopakv: what if

captureAmount = 203.49
getTotalInvoiced = 2.49
getGrandTotal = 205.9877

= (20349+249) > 20599?

i think it might just be safer to acount for edge cases than to say we send something different etc, because there are many ways to come with a case to break the above logic since you round before adding / checking?

roopakv: you might be right in terms of the operations taking place, but it is not obvious to a new person entering the codebase?

oxigenka: Sorry, I've mistaken in the previous comment:
> the condition will remain `true`.

should be
> the condition will remain `false`, and no exception raises.

What about example with total invoiced, it would be:
> = (20349+249 = 20598) > 20599

In this case, the condition is also `false`, and it's not an error here :)

Finally, you're absolutely right about the lack of comments on what these rounding operations are applied for, and I will add it to the code.

Helper/Order.php
```diff
@@ -1644,15 +1644,61 @@ protected function isCaptureHookRequest($newCapture)
...
      * @throws \Exception
      */
-    protected function validateCaptureAmount(OrderInterface $order, $captureAmount) {
+    protected function validateCaptureAmount(OrderInterface $order, $captureAmount)
+    {
         $isInvalidAmount = !isset($captureAmount) || !is_numeric($captureAmount) || $captureAmount < 0;
-        $isInvalidAmountRange = $order->getTotalInvoiced() + $captureAmount > $order->getGrandTotal();
 
-        if($isInvalidAmount || $isInvalidAmountRange) {
+        /**
```
ethanwayda22: This is wonderful documentation, however, this is too verbose for inline documentation. Please add this to a readme or update internal documentation.

oxigenka: Updated comments (moved examples to the pull request description)

oxigenka:
> Thanks for cleaning this up @oxigenka. Can you add tests - `OrderTest.php` exists already so this should be pretty quick.

The updated method is `protected`. Do we need to create tests on protected methods?

oxigenka:
> please add a test

@roopakv Please take into account that the incapsulation level of an updated method is `protected`, and @matt-thomason agreed that there's not obligated to create a separate test for this. It should just be considered when the test on the `updateOrderPayment` method will be creating as it does not exist now.

Helper/Order.php
```diff
@@ -1644,15 +1644,26 @@ protected function isCaptureHookRequest($newCapture)
...
+         * when validating captured amount.
+         * Rounding operations are applied to each operand in order to avoid cases when the grand total
+         * is formally less (before it has been rounded) than the sum of the captured amount and the total invoiced.
+         */
+        $isInvalidAmountRange = $this->cartHelper->getRoundAmount($order->getTotalInvoiced())
+            + $this->cartHelper->getRoundAmount($captureAmount)
+            > $this->cartHelper->getRoundAmount($order->getGrandTotal());
+
+        if ($isInvalidAmount || $isInvalidAmountRange) {
             throw new \Exception( __('Capture amount is invalid'));
```
daisy1754: can you add amount in error message to make future debugging easier?

oxigenka: It's a good point, thank you!

oxigenka:
> can we add test for updateOrderPayment? I think you can write test case that would fail without this change?

@daisy1754 I've talked about such possibility with @matt-thomason and we've agreed that the `updateOrderPayment` method should be refactored as it contains too many code to be a fine testable.
Sure, I can write a test without taking into account other logic in the `update...` method, but IMHO that would not be as helpful as we might wish for the good testing of the method.

PR #401, opened 2019-10-15
Convert attribute to string if it's a boolean before sending to the Bolt API

johnmccombs1:
When the attribute value is a boolean, Bolt recognizes it as an invalid JSON and doesn’t allow the order to be created on Bolt. This ensures the attribute value is converted to string if it's a boolean when sending it to Bolt.

Fixes: https://app.asana.com/0/564264490825835/1142017681006690

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -1033,9 +1033,11 @@ class Cart extends AbstractHelper
                 if(isset($item_options['attributes_info'])){
                     $properties = [];
                     foreach($item_options['attributes_info'] as $attribute_info){
+                        // Convert attribute to string if it's a boolean before sending to the Bolt API
+                        $attributeValue = is_bool($attribute_info['value']) ? var_export($attribute_info['value'], true) : $attribute_info['value'];
                         $properties[] = (object) [
                             "name" => $attribute_info['label'],
-                            "value" => $attribute_info['value']
+                            "value" => $attributeValue
                         ];
                     }
                     $product['properties'] = $properties;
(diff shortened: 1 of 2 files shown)


Helper/Cart.php
```diff
@@ -1032,9 +1032,10 @@ function ($item) use ($imageBlock, &$totalAmount, &$diff, $storeId) {
                 if(isset($item_options['attributes_info'])){
                     $properties = [];
                     foreach($item_options['attributes_info'] as $attribute_info){
+                        $attributeValue = is_string($attribute_info['value']) ? $attribute_info['value'] : var_export($attribute_info['value'], true);
```
roopakv: could we add some tests for this. Also why does `var_export` fix this issue? if this was an array var this would srtill break right?

would love to see tests + examples for this

matt-thomason: Agree with @roopakv here. If `$attribute_info['value']` is JSON, then wouldn't converting that JSON value into a string break something later on down the line as well?

YevhenShBolt:
@johnmccombs1 any thoughts or fixes?

johnmccombs1:
> @johnmccombs1 curious what this PR is trying to fix? i sort of feel like converting to string will only break a lot fo cases. Could you please help me understand what is up

@roopakv Bolt throws an exception if the body of the create order API request has a boolean property value. See here -  https://screencast.com/t/ACFPrsuZ

Here is the Bugsnag link for the error response from Bolt: https://app.bugsnag.com/bolt/magento-2/errors/5dd792e130aa700018e5b30c?filters[event.since][0][type]=eq&filters[event.since][0][value]=30d&filters[search][0][type]=eq&filters[search][0][value]=Invalid

In order to fix it, we need to convert the boolean to its string representation. E.g. false to "false"

Please see the Asana task for more detail about the issue.

PR #414, opened 2019-10-28
added tests for /Cart/Email.php

njfaries:
# Description
Added tests for ../Cart/Email.php. 

Part of https://app.asana.com/0/1131034199022718/1137538625478186/f

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Test/Unit/Controller/Cart/EmailTest.php
+++ b/Test/Unit/Controller/Cart/EmailTest.php
@@ -1,0 +1,182 @@
+<?php
+
+namespace Bolt\Boltpay\Test\Unit\Controller\Cart;
+
+use Bolt\Boltpay\Controller\Cart\Email;
+use Bolt\Boltpay\Helper\Bugsnag;
+use Bolt\Boltpay\Helper\Cart as CartHelper;
+use Magento\Checkout\Model\Session as CheckoutSession;
+use Magento\Customer\Model\Session as CustomerSession;
+use Magento\Framework\App\Action\Context;
+use Magento\Framework\App\RequestInterface;
+use Magento\Framework\Exception\LocalizedException;
+use Magento\Quote\Model\Quote;
+use PHPUnit\Framework\TestCase;
+
+class EmailTest extends TestCase
+{
+    /**
+     * @var Context
+     */
+    private $context;
+
+    /**
+     * @var CheckoutSession
+     */
+    private $checkoutSession;
+
+    /**
+     * @var CustomerSession
+     */
+    private $customerSession;
+
+    /**
+     * @var Bugsnag
+     */
+    private $bugsnag;
+
+    /**
+     * @var CartHelper
+     */
+    private $cartHelper;
+
+    /**
+     * @var Quote quote
+     */
+    private $quote;
+
+    /**
+     * @var RequestInterface request
+     */
+    private $request;
+
+    /**
+     * @var Email
+     */
+    private $currentMock;
+
+    public function setUp()
+    {
+        $this->initRequiredMocks();
+        $this->initCurrentMock();
+
+    }
+
+    /**
+     * @covers Email::execute
+     */
+    public function testQuoteDoesNotExist()
+    {
+        $this->checkoutSession->method('getQuote')->willReturn(null);
+        $exception = new LocalizedException(__('Quote does not exist.'));
+        $this->bugsnag->expects($this->once())
+            ->method('notifyException')
+            ->with($this->equalTo($exception));
+        $this->currentMock->execute();
+    }
+
+    /**
+     * @covers Email::execute
+     */
+    public function testQuoteIdFails()
+    {
+        $this->quote->method('getId')->willReturn(false);
+        $exception = new LocalizedException(__('Quote does not exist.'));
+        $this->bugsnag->expects($this->once())
+                    ->method('notifyException')
+                    ->with($this->equalTo($exception));
+        $this->currentMock->execute();
+    }
+
+    /**
+     * @covers Email::execute
+     */
+    public function testNoEmail()
+    {
+        $this->checkoutSession->method('getQuote')->willReturn($this->quote);
+        $this->request->method('getParam')->willReturn('');
+        $exception = new LocalizedException(__('No email received.'));
+        $this->bugsnag->expects($this->once())
+                    ->method('notifyException')
+                    ->with($this->equalTo($exception));
+        $this->currentMock->execute();
+    }
+
+    /**
+     * This doesn't really seem to test invalid emails as the validateEmail method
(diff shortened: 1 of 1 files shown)


Test/Unit/Controller/Cart/EmailTest.php
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * @var Email
+     */
+    private $currentMock;
+
+    public function setUp()
+    {
+        $this->initRequiredMocks();
+        $this->initCurrentMock();
+
```
roopakv: Not remove extra line

Test/Unit/Controller/Cart/EmailTest.php
```diff
@@ -0,0 +1,182 @@
...
+        $this->request = $this->createMock(RequestInterface::class);
+
+        //methods
+//        $this->checkoutSession->method('getQuote')->willReturn($this->quote);
+        $this->cartHelper->method('quoteResourceSave')->willReturn(null);
+        $this->context->method('getRequest')->willReturn($this->request);
+    }
+
+    private function initCurrentMock()
```
roopakv: Nit: let's move this to the top. When reading a test file all initializations etc tend to be at the top usually which makes it easy to understand what thebtest Is doing. Let's stick with that pattern.

Test/Unit/Controller/Cart/EmailTest.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        $this->currentMock = $this->getMockBuilder(Email::class)
```
roopakv: Can we use the object manager to initialize the object?

https://github.com/BoltApp/bolt-magento2/blob/9586db85834ee0e20fb2eb9c0c555bfea055c616/Test/Unit/Model/FeatureSwitchRepositoryTest.php#L44

If not why do we need a mock based object to test the current method? Won't it mock out things we should be testing?

roopakv: I'm assuming `enableProxyingToOriginalMethods()` alleviates some issues but curious

Test/Unit/Controller/Cart/EmailTest.php
```diff
@@ -0,0 +1,167 @@
...
+        $this->cartHelper = $this->createMock(CartHelper::class);
+
+        //objects needed for method
+        $this->quote = $this->createMock(Quote::class);
+        $this->quote->method('getId')->willReturn('1234');
+
+        $this->request = $this->createMock(RequestInterface::class);
+
+        //methods
+//        $this->checkoutSession->method('getQuote')->willReturn($this->quote);
```
roopakv: Nit: remove if not being used

njfaries: Oops. Fixed

Test/Unit/Controller/Cart/EmailTest.php
```diff
@@ -0,0 +1,166 @@
...
+        $this->request->method('getParam')->willReturn('invalidemail');
+        $this->cartHelper->method('validateEmail')->willReturn(false);
+        $exception = new LocalizedException(__('Invalid email: %1', 'invalidemail'));
+        $this->bugsnag->expects($this->once())
+                    ->method('notifyException')
+                    ->with($this->equalTo($exception));
+        $this->currentMock->execute();
+    }
+
+    public function testOtherException()
```
matt-thomason: Nit: Can you revisit some of these test names? A common pattern is to have the function being tested, and then what makes it special. For example, this would be: `testExecute_notifyException`.

Test/Unit/Controller/Cart/EmailTest.php
njfaries: I was following what was done in `Model/PaymentTest.php` for all the test structure. `enableProxyingToOriginalMethods()` does make sure the method is called correctly.

Test/Unit/Controller/Cart/EmailTest.php
```diff
@@ -0,0 +1,166 @@
...
+                    ->with($this->equalTo($exception));
+        $this->currentMock->execute();
+    }
+
+    /**
+     * This doesn't really seem to test invalid emails as the validateEmail method
+     * needs to be mocked. Generally just making sure that if validateEmail comes
+     * back false we error handle properly
+     */
+
```
matt-thomason: Don't have a blank line between comments and function pls

Test/Unit/Controller/Cart/EmailTest.php
roopakv: I feel like this method is prone to skipping the actual methods. @daisy1754 @matt-thomason what's the preferred way here? I sort of prefer new Object manager way but that's with no deep thought into the issue ( only bcos the other seems like it could lead to trouble)

Test/Unit/Controller/Cart/EmailTest.php
```diff
@@ -0,0 +1,164 @@
...
+            ->setConstructorArgs([
+                $this->context,
+                $this->checkoutSession,
+                $this->customerSession,
+                $this->bugsnag,
+                $this->cartHelper
+            ])
+            ->enableProxyingToOriginalMethods()
+            ->getMock();
+        return $this->currentMock;
```
danac-gs: This line seems redundant, the returned value is never used directly.

Test/Unit/Controller/Cart/EmailTest.php
danac-gs: enableProxyingToOriginalMethods should be OK

Test/Unit/Controller/Cart/EmailTest.php
njfaries: As with the discussion about mocking the email class, this was just something I followed from `PaymentTest.php`, though I see it's not in your recent commits. Removing.

Test/Unit/Controller/Cart/EmailTest.php
```diff
@@ -0,0 +1,163 @@
...
+     * needs to be mocked. Generally just making sure that if validateEmail comes
+     * back false we error handle properly
+     */
+    public function testExecute_invalidEmail()
+    {
+        $this->checkoutSession->method('getQuote')->willReturn($this->quote);
+        $this->request->method('getParam')->willReturn('invalidemail');
+        $this->cartHelper->method('validateEmail')->willReturn(false);
+        $exception = new LocalizedException(__('Invalid email: %1', 'invalidemail'));
+        $this->bugsnag->expects($this->once())
```
ethanwayda22: nit: can you have these two tied to a single variable so that a change in one does not break the other?

njfaries: Done

PR #418, opened 2019-10-30
Restock product quantity when the payment is irreversibly rejected

johnmccombs1:
# Description
When Bolt rejects the payment irreversibly, a webhook is sent to Magento to cancel the order. The order state and status are updated to canceled but the product isn’t restocked as the underlining logic for canceling the order isn’t run. More specifically, Magento doesn’t allow the order to cancel when its status is payment review. This PR adds the logic to restock the product quantity for the rejected irreversible hook.

Fixes: https://app.asana.com/0/564264490825835/1147250679267337

#changelog
Restock product quantity when the payment is irreversibly rejected

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -1336,8 +1336,19 @@ class Order extends AbstractHelper
                 $order->setStatus($order->getConfig()->getStateDefaultStatus(OrderModel::STATE_HOLDED));
             }
         } elseif ($state == OrderModel::STATE_CANCELED) {
-            $this->cancelOrder($order);
-            return;
+            if ($order->canCancel()){
+                $this->cancelOrder($order);
+                return;
+            }
+
+            try{
+                // Restock product quantity when the payment is irreversibly rejected
+                $order->registerCancellation('', false);
+            } catch (\Exception $e){
+                // Put the order in "cancelled" state even if the previous call fails
+                $order->setState(OrderModel::STATE_CANCELED);
+                $order->setStatus($order->getConfig()->getStateDefaultStatus(OrderModel::STATE_CANCELED));
+            }
         } else {
             $order->setState($state);
             $order->setStatus($order->getConfig()->getStateDefaultStatus($state));
--- a/Test/Unit/Helper/OrderTest.php
+++ b/Test/Unit/Helper/OrderTest.php
@@ -259,6 +259,8 @@ class OrderTest extends TestCase
                 'setState',
                 'setStatus',
                 'isCanceled',
+                'canCancel',
+                'registerCancellation',
                 'addStatusHistoryComment',
             ]
         );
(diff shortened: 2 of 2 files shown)


Helper/Order.php
```diff
@@ -1336,8 +1336,19 @@ public function setOrderState($order, $state)
...
+            if ($order->canCancel()){
+                $this->cancelOrder($order);
+                return;
+            }
+
+            try{
+                // Restock product quantity when the payment is irreversibly rejected
+                $order->registerCancellation('', false);
+            } catch (\Exception $e){
+                // Put the order in "cancelled" state even if the previous call fails
```
roopakv: Let's bump a metric here stating that we fell back.

I'm curious as to why this happens in the first place

Helper/Order.php
```diff
@@ -1336,8 +1336,19 @@ public function setOrderState($order, $state)
                 $order->setStatus($order->getConfig()->getStateDefaultStatus(OrderModel::STATE_HOLDED));
             }
         } elseif ($state == OrderModel::STATE_CANCELED) {
-            $this->cancelOrder($order);
-            return;
+            if ($order->canCancel()){
+                $this->cancelOrder($order);
+                return;
```
roopakv: Just for my understanding:

What is the difference between cancelling order here and the registerCancel below. Why can't we use the below one for all cases

Helper/Order.php
johnmccombs1: @roopakv, I didn't see any exceptions thrown while doing the development work. This is just in case an exception occurs and I just followed the same pattern that existed before. See here https://github.com/BoltApp/bolt-magento2/blob/master/Helper/Order.php#L1335

Let me know if you think I should remove the try-catch block.

Helper/Order.php
johnmccombs1: The difference here is that we’re dealing with 2 separate cases for canceling the order:

1. Normal order cancelation (e.g. void hook request)

2. Canceling the order when the payment is rejected irreversibly (payment is denied)

For #1, Magento cancels the payment and then registers cancelation. See here https://screencast.com/t/c5OHnLfv5lC

For #2, Magento just registers cancelation

This is how default Magento works, we can’t use #1 for #2 as the order state is payment review and we can’t cancel the order in that case https://screencast.com/t/KwjNEKdPQ3Y

We can use #2 for #1, but the logic to cancel the payment won’t be run and I’m not sure if it will cause any issues. I think it’s better to keep the logic like it is here. Thoughts?

Helper/Order.php
roopakv: Let's keep it but bump a metric.this Isc The perfect case of something we don't expect to happen happening. We can then track if this happens and how many times

johnmccombs1: @roopakv can you explain what you mean by "bump a metric"? Are you saying that I should add a Bugsnag notification, or bump the Bolt version number, or something else?

PR #423, opened 2019-11-06
Add JavaScript to track checkout funnel in onepage checkout

daisy1754:
# Description
This PR adds a JavaScript running on one page checkout. Lots of (hacky) DOM scraping to detect checkout funnel transition and send it to Bolt via `BoltTrack.recordEvent`. 

I was looking for better way to hook into native magento's event but wasn't able to find - welcome for comment and suggestions

I'll add test with subsequent PR. We currently don't have JavaScript test so adding test requires good amount of new code and config files.

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/view/frontend/layout/checkout_index_index.xml
+++ b/view/frontend/layout/checkout_index_index.xml
@@ -63,7 +63,8 @@
             <block class="Bolt\Boltpay\Block\Js"
                    name="checkout_funnel.js"
                    template="Bolt_Boltpay::js/checkout_funnel.js.phtml"
-                   ifconfig="payment/boltpay/track_checkout_funnel" />
+                   ifconfig="payment/boltpay/track_checkout_funnel" >
+            </block>
         </referenceContainer>
     </body>
 </page>
--- a/view/frontend/layout/checkout_index_index.xml
+++ b/view/frontend/layout/checkout_index_index.xml
@@ -63,8 +63,7 @@
             <block class="Bolt\Boltpay\Block\Js"
                    name="checkout_funnel.js"
                    template="Bolt_Boltpay::js/checkout_funnel.js.phtml"
-                   ifconfig="payment/boltpay/track_checkout_funnel" >
-            </block>
+                   ifconfig="payment/boltpay/track_checkout_funnel" />
         </referenceContainer>
     </body>
 </page>
--- a/view/frontend/templates/js/checkout_funnel.js.phtml
+++ b/view/frontend/templates/js/checkout_funnel.js.phtml
@@ -1,9 +1,5 @@
 <?php
 if (!$block->shouldTrackCheckoutFunnel()) return;
-
-$objectManager =  \Magento\Framework\App\ObjectManager::getInstance();
-$repo = $objectManager->get('\Magento\Framework\View\Asset\Repository');
-
 ?>
 
-<script type="text/javascript" src="<?php echo $repo->getUrl("Bolt_Boltpay::js/track-checkout.js") ?>"></script>
+<script type="text/x-magento-init"> { "*" : { "Bolt_Boltpay/js/track-checkout" : {} } } </script>
--- a/view/frontend/templates/js/checkout_funnel.js.phtml
+++ b/view/frontend/templates/js/checkout_funnel.js.phtml
@@ -1,7 +1,9 @@
 <?php
 if (!$block->shouldTrackCheckoutFunnel()) return;
+
+$objectManager =  \Magento\Framework\App\ObjectManager::getInstance();
+$repo = $objectManager->get('\Magento\Framework\View\Asset\Repository');
+
 ?>
 
-<script type="text/javascript">
-    // TODO: add logic here
-</script>
+<script type="text/javascript" src="<?php echo $repo->getUrl("Bolt_Boltpay::js/track-checkout.js") ?>"></script>
--- a/view/frontend/web/js/track-checkout.js
+++ b/view/frontend/web/js/track-checkout.js
@@ -20,8 +20,8 @@ function trackFunnel(step) {
         // prevent duplicate tracking
         return;
     }
-    currentStep = step;
     BoltTrack.recordEvent(step);
+    currentStep = step;
 }
 
 function setupListnerForShippingForm() {
--- a/view/frontend/web/js/track-checkout.js
+++ b/view/frontend/web/js/track-checkout.js
@@ -2,6 +2,15 @@
 
 var currentStep = '';
 
+function waitFor(condition, doWhenReady) {
+    var waitForInterval = setInterval(function () {
+        if (condition()) {
+            doWhenReady();
+            clearInterval(waitForInterval);
+        }
+    }, 50);
+}
+
 function trackFunnel(step) {
     if (step === currentStep) {
         // prevent duplicate tracking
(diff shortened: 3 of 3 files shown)


view/frontend/layout/checkout_index_index.xml
```diff
@@ -63,7 +63,8 @@
             <block class="Bolt\Boltpay\Block\Js"
                    name="checkout_funnel.js"
                    template="Bolt_Boltpay::js/checkout_funnel.js.phtml"
-                   ifconfig="payment/boltpay/track_checkout_funnel" />
+                   ifconfig="payment/boltpay/track_checkout_funnel" >
```
johnmccombs1: I don’t think we need this if the block doesn’t have any children

view/frontend/templates/js/checkout_funnel.js.phtml
```diff
@@ -1,7 +1,9 @@
 <?php
 if (!$block->shouldTrackCheckoutFunnel()) return;
+
+$objectManager =  \Magento\Framework\App\ObjectManager::getInstance();
```
johnmccombs1: We shouldn’t use the Object Manager to initialize an object. The best practice here is to inject the track-checkout.js using RequireJs
Something like this:
> <script type="text/x-magento-init">
>     {
>         "*" : {
>             "Bolt_Boltpay/js/track-checkout" : {}
>         }
>     }
> </script>

view/frontend/web/js/track-checkout.js
```diff
@@ -0,0 +1,68 @@
...
+    });
+
+    // on method can detect checkout button added after this query is executed.
+    jQuery(".page-wrapper").on("click", ".action.primary.checkout", function() {
+        trackFunnel("onPaymentSubmit")
+        return false;
+    });
+}
+
+waitFor(
```
johnmccombs1: It looks like you're using plain JS to wait when JQuery is available and then using it for initializing the events. That might not work if JQuery isn’t available on the page as Magento uses RequireJS to inject the dependency (E.g. JQuery) and sometimes the global JQuery object isn’t available or they might use the $ variable as the global object for JQuery.

It might be better to create an ADM RequireJS module and inject the JQuery library as a dependency. Then you can remove the logic for waiting for JQuery to be available.

view/frontend/web/js/track-checkout.js
```diff
@@ -0,0 +1,68 @@
...
+function trackFunnel(step) {
+    if (step === currentStep) {
+        // prevent duplicate tracking
+        return;
+    }
+    currentStep = step;
+    BoltTrack.recordEvent(step);
+}
+
+function setupListnerForShippingForm() {
```
johnmccombs1: I don’t think this will work with a logged-in customer that has a default shipping address.

view/frontend/templates/js/checkout_funnel.js.phtml
daisy1754: nice I had hard time figuring out this; thank you

view/frontend/web/js/track-checkout.js
daisy1754: yes this only handles guest checkout. added todo

view/frontend/web/js/track-checkout.js
daisy1754: cool I updated code to use requriejs

view/frontend/web/js/track-checkout.js
```diff
@@ -0,0 +1,70 @@
+/* Track checkout funnel transition in magento's default checkout */
+
+// TODO: support logged in user
+
+var currentStep = '';
+
+function waitFor(condition, doWhenReady) {
```
roopakv: curious: doesnt jquery have its own way to wait for conditions? whats thwe advantage of this ^

view/frontend/web/js/track-checkout.js
```diff
@@ -0,0 +1,70 @@
...
+    });
+
+    // on method can detect checkout button added after this query is executed.
+    jQuery(".page-wrapper").on("click", ".action.primary.checkout", function() {
+        trackFunnel("onPaymentSubmit")
+        return false;
+    });
+}
+
+require(["jquery"], function(){
```
roopakv: nit: would this work?
```suggestion
require(["jquery"], init);
```

view/frontend/web/js/track-checkout.js
daisy1754: hm I don't think jquery has this kind of polling wait

view/frontend/web/js/track-checkout.js
```diff
@@ -0,0 +1,68 @@
+/* Track checkout funnel transition in magento's default checkout */
+
+// TODO: support logged in user
+
+var currentStep = '';
+
+function waitFor(condition, doWhenReady) {
+    var waitForInterval = setInterval(function () {
```
roopakv: discussed. lets use `setTimeout`

view/frontend/web/js/track-checkout.js
```diff
@@ -0,0 +1,70 @@
...
+// TODO: support logged in user
+
+var currentStep = '';
+
+function waitFor(condition, doWhenReady) {
+    var checkCondition = function () {
+        if (condition()) {
+            doWhenReady();
+        } else {
+            setTimeout(condition, 50);
```
roopakv: this will just call condition in 50ms? you want to call `checkCondition`?

daisy1754: woops fixed thank you

view/frontend/web/js/track-checkout.js
```diff
@@ -0,0 +1,70 @@
...
+    checkCondition();
+}
+
+function trackFunnel(step) {
+    if (step === currentStep) {
+        // prevent duplicate tracking
+        return;
+    }
+    currentStep = step;
+    BoltTrack.recordEvent(step);
```
roopakv: can this ever fail? do we want to set `currentStep = step;` after this?

view/frontend/web/js/track-checkout.js
johnmccombs1: I think we should add a new argurment to specify the number of retry attempts to check the condition. If the condition always returns false, we might end up running this check forever in the checkout page. We should also assign a default value for that argument like 50.

view/frontend/web/js/track-checkout.js
```diff
@@ -0,0 +1,70 @@
...
+        });
+    });
+}
+
+function init() {
+    trackFunnel("onCheckoutStart");
+
+    waitFor(
+        // wait for shipping form to be fully rendered, ie., we have 9 required fields in the page.
+        function() { return jQuery("#shipping-new-address-form .field._required input,select").length === 9; },
```
johnmccombs1: I would estimate that 1/3 of the merchants I've worked with customize their checkout to include new required fields or to remove some of the required fields. If they do, this won’t work on their sites.

Maybe you should use a required value range as the condition instead of a specific number. Something like this:

> function() { return jQuery("#shipping-new-address-form .field._required input,select").length >== 5; },

daisy1754: I updated to `>= 9`. We'll working closely with first few beta merchant and hopefully we can find better approach during beta

view/frontend/web/js/track-checkout.js
```diff
@@ -0,0 +1,72 @@
+/* Track checkout funnel transition in magento's default checkout */
+
+// TODO: support logged in user
+
+var currentStep = '';
+
+function waitFor(condition, doWhenReady, pollIntervalMillis) {
+    var defaultPollIntervalMillis = 100;
+    var maxRetryCount = 30;
+    var checkCondition = function (retryCount) {
```
johnmccombs1: Not really functionally much different. Just might be a little cleaner to do something like this but I'm find without it.

> function waitFor(condition, doWhenReady, pollIntervalMillis = 100, maxRetryCount = 30) {    
>     var checkCondition = function (retryCount) {
>         if (condition()) {
>             doWhenReady();
>         } else if (retryCount < maxRetryCount) {
>             setTimeout(function(){checkCondition(retryCount + 1)}, pollIntervalMillis);
>         }
>     };
>     checkCondition(0);
> }

daisy1754: unfortunately IE doesn't support default argument value

PR #425, opened 2019-11-07
Load the correct frontend division key configuration

johnmccombs1:
# Description
Currently, the logic to get the publishable key is to check the multi-step key and use it if it’s not empty. If it is empty then the payment only key will be used.

This causes an issue when we want to use both checkout divisions for the storefront. This PR loads the correct key based on the request. If the request is from the checkout page, then use the payment only key, otherwise, use the multiple-step key.

Fixes: https://app.asana.com/0/564264490825835/1148566476960345

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Block/Js.php
+++ b/Block/Js.php
@@ -102,7 +102,11 @@ class Js extends Template
      */
     public function getCheckoutKey()
     {
-        return $this->configHelper->getAnyPublishableKey();
+        if($this->configHelper->isPaymentOnlyCheckoutEnabled() && $this->_request->getFullActionName() == Config::CHECKOUT_PAGE_ACTION){
+            return $this->configHelper->getPublishableKeyPayment();
+        }
+
+        return $this->configHelper->getPublishableKeyCheckout();
     }
 
     /**
@@ -112,7 +116,8 @@ class Js extends Template
      */
     public function getReplaceSelectors()
     {
-        $subject = trim($this->configHelper->getReplaceSelectors());
+        $isBoltUsedInCheckoutPage = $this->configHelper->isPaymentOnlyCheckoutEnabled() && $this->_request->getFullActionName() == Config::CHECKOUT_PAGE_ACTION;
+        $subject = ($isBoltUsedInCheckoutPage) ? '' : trim($this->configHelper->getReplaceSelectors());
 
         return array_filter(explode(',', preg_replace('/\s+/', ' ', $subject)));
     }
--- a/Test/Unit/Block/JsTest.php
+++ b/Test/Unit/Block/JsTest.php
@@ -21,7 +21,7 @@ use Bolt\Boltpay\Block\Js as BlockJs;
 use Bolt\Boltpay\Helper\Bugsnag;
 use Bolt\Boltpay\Helper\Config as HelperConfig;
 use Bolt\Boltpay\Helper\Cart as CartHelper;
-
+use Magento\Framework\App\Request\Http;
 /**
  * Class JsTest
  *
@@ -41,6 +41,12 @@ class JsTest extends \PHPUnit\Framework\TestCase
      * @var \Magento\Framework\View\Element\Template\Context
      */
     protected $contextMock;
+
+    /**
+     * @var Http
+     */
+    private $requestMock;
+
     /**
      * @var \Magento\Checkout\Model\Session
      */
@@ -86,7 +92,7 @@ class JsTest extends \PHPUnit\Framework\TestCase
             'getPublishableKeyPayment', 'getPublishableKeyCheckout', 'getPublishableKeyBackOffice',
             'getReplaceSelectors', 'getGlobalCSS', 'getPrefetchShipping', 'getQuoteIsVirtual',
             'getTotalsChangeSelectors', 'getAdditionalCheckoutButtonClass', 'getAdditionalConfigString', 'getIsPreAuth',
-            'shouldTrackCheckoutFunnel'
+            'shouldTrackCheckoutFunnel','isPaymentOnlyCheckoutEnabled'
         ];
 
         $this->configHelper = $this->getMockBuilder(HelperConfig::class)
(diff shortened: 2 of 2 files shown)


danac-gs:
@johnmccombs1  am not aware if we ever used payment only key in Magento 2. Our support for for default Magento checkout page, although implemented (long time ago) up to some point, was never really tested and I guess the requests for that were successfully pushed back if ever required from merchants.
The method you are updating is only used here https://github.com/BoltApp/bolt-magento2/blob/fixbug/load_the_correct_frontend_division_key_configuration/view/frontend/templates/js/boltjs.phtml#L27 for loading track.js and connect.js.
With the implementation without the fallback we would not load bolt javascript files on checkout page for majority if not all existing M2 merchants simply because the payment only divisions were not needed thus never created so there's no the payment only publishable key.

Block/Js.php
```diff
@@ -62,13 +69,15 @@ public function __construct(
...
         Bugsnag $bugsnag,
+        Http $request,
         array $data = []
     ) {
         parent::__construct($context, $data);
         $this->configHelper = $configHelper;
         $this->checkoutSession = $checkoutSession;
         $this->cartHelper = $cartHelper;
         $this->bugsnag = $bugsnag;
+        $this->request = $request;
```
YevhenShBolt: @johnmccombs1 
Wow! Why did you inject in such way?

Every block have `$this->_request` variable. The variable contain object of class `Magento\Framework\App\Request\Http`.
Even if your block do not have such variable, so every `$context` in block have it - you can inject it
`$this->request = $context->getRequest()` 
(the inherited class is `Magento\Framework\App\Request\Http`).

Please remove this extra injection.

Test/Unit/Block/JsTest.php
```diff
@@ -104,6 +109,10 @@ protected function setUp()
 
         $this->cartHelperMock = $this->createMock(CartHelper::class);
         $this->bugsnagHelperMock = $this->createMock(Bugsnag::class);
+        $this->requestMock = $this->getMockBuilder(Http::class)
+            ->disableOriginalConstructor()
+            ->setMethods(['getFullActionName'])
+            ->getMock();
```
YevhenShBolt: Please update a test regarding comment above.

Test/Unit/Block/JsTest.php
```diff
@@ -169,18 +179,39 @@ public function testGetConnectJsUrlForProductionMode()
...
+        $this->assertStringStartsWith('pKv_', $result, '"Publishable Key Payment Only" doesn\'t work properly');
+        $this->assertEquals(strlen($key), strlen($result), '"Publishable Key Payment Only" has an invalid length');
+    }
 
-        $this->assertStringStartsWith('pKv_', $result, '"Any Publishable Key" not working properly');
-        $this->assertEquals(strlen($key), strlen($result), '"Any Publishable Key" have invalid length');
+    /**
+     * @inheritdoc
+     */
+    public function testGetCheckoutKey_shoppingCartPage()
```
YevhenShBolt: I suggest to add a `@dataProvider someName` to simplify test cases. Also it would be more simply in future to add new scenarious especially negative scenarious (e.x when we do not have a publishable key here)

Block/Js.php
```diff
@@ -102,7 +102,11 @@ public function getConnectJsUrl()
      */
     public function getCheckoutKey()
     {
-        return $this->configHelper->getAnyPublishableKey();
+        if($this->_request->getFullActionName() == Config::CHECKOUT_PAGE_ACTION){
+            return $this->configHelper->getPublishableKeyPayment();
```
danac-gs: @johnmccombs1 there's no payment-only division for majority of the merchants. So no publishable key as well. This would affect Bolt logic on the checkout page, setting up track.js and connect.js, configuring the checkout button in minicart if available. When the plugin was initially developed there was an option to dynamically change the publishable key on the page depending on the frontend route - shipping or payment. On the shipping step the multi-step was used, and payment only on the payment form. This functionality was removed on Bolt side so we can not support it anymore. I guess that is one of the reasons we do not support magento default (payment only) checkout.
Perhaps we should discuss this on a call. cc @matt-thomason @YevhenShBolt @njfaries @ethanwayda22

danac-gs:
> LGTM

@YevhenShBolt please take my comments above into consideration and let us know what you think.

Block/Js.php
johnmccombs1: @danac-gs, Nature Hills Nursery wants to use Bolt in both the cart and the checkout page. See this task for more information about the request https://app.asana.com/0/564264490825835/1140792639357263

On Magento 1, we can both have divisions working on the cart and the checkout page and I used both of the divisions for several M1 merchants in the past: Akira and  Shop Rescue Spa.

Can you explain to me why this PR causes the issues on the checkout page on M2? The mini cart shouldn’t be used in the checkout page, right?

danac-gs: @johnmccombs1 
1. I agree the minicart should be removed from the checkout page. However, by default it is shown and so it can be for some live stores.
Proposal: merchants that want to use Bolt on the default checkout should have minicart disabled on the page.
2. My concern was more regarding backward compatibility and loading track.js (we may or may not need connect.js depending on minicart, see #1 above) 
Proposal: I've just realized that we have `Enable Payment Only Checkout` configuration option, false by default. You can include it in the condition, alongside 
`$this->_request->getFullActionName() == Config::CHECKOUT_PAGE_ACTION`
and everything should be fine. This way the multi-step key would load as a fallback, as it does for the old stores, and track.js would work as expected.

matt-thomason:
@danac-gs @johnmccombs1 John is right here - we should support the payment only option here consistently through our plugins, especially when it comes to options like these. Unfortunately, this means we probably need to communication through Support this change and coordinate setting up new divisions for the merchants that rely on the existing functionality.

danac-gs:
> @danac-gs @johnmccombs1 John is right here - we should support the payment only option here consistently through our plugins, especially when it comes to options like these. Unfortunately, this means we probably need to communication through Support this change and coordinate setting up new divisions for the merchants that rely on the existing functionality.

@matt-thomason after taking another look I realized we can add this change and make it backward compatible without the necessity of creating new divisions if we don't really need them (old merchants and the new ones that don't use default checkout). Point 2. in my previous comment, https://github.com/BoltApp/bolt-magento2/pull/425#discussion_r345248973

PR #427, opened 2019-11-07
Add tracking observer to M2

akulkapoor:
# Description
This adds an observer that fires every time tracking is added and sends relevant data to the Bolt API

Fixes: (link Asana Task)

\#changelog

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/Api.php
+++ b/Helper/Api.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Bolt magento2 plugin
  *
@@ -61,6 +62,11 @@ class Api extends AbstractHelper
      * Api create order
      */
     const API_CREATE_ORDER = 'merchant/orders';
+
+    /**
+     * Api create tracking
+     */
+    const API_CREATE_TRACKING = 'merchant/track_shipment';
 
     /**
      * Api void transaction
@@ -174,7 +180,7 @@ class Api extends AbstractHelper
      */
     private function isResponseError($response)
     {
-        $arr = (array)$response;
+        $arr = (array) $response;
         return array_key_exists('errors', $arr) || array_key_exists('error_code', $arr);
     }
 
@@ -209,7 +215,7 @@ class Api extends AbstractHelper
             $this->configHelper->getModuleVersion(),
             $requestData,
             $apiKey,
-            (array)$request->getHeaders()
+            (array) $request->getHeaders()
         );
 
         $request->setHeaders($headers);
@@ -291,8 +297,8 @@ class Api extends AbstractHelper
         $requestMethod = empty($requestData->getRequestMethod()) ? 'POST' : $requestData->getRequestMethod();
         $apiUrl        = $this->getFullApiUrl($dynamicApiUrl);
 
-        $headers    = (array)$requestData->getHeaders();
-        $statusOnly = (bool)$requestData->getStatusOnly();
+        $headers    = (array) $requestData->getHeaders();
+        $statusOnly = (bool) $requestData->getStatusOnly();
 
         $request = $this->requestFactory->create();
         $request->setApiData($apiData);
--- a/Helper/FeatureSwitch/Decider.php
+++ b/Helper/FeatureSwitch/Decider.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Bolt magento2 plugin
  *
@@ -26,7 +27,8 @@ use Magento\Framework\App\State;
 use Magento\Framework\App\Helper\AbstractHelper;
 use Magento\Framework\App\Helper\Context;
 
-class Decider extends AbstractHelper {
+class Decider extends AbstractHelper
+{
     /**
      * @var Manager
      */
@@ -87,7 +89,8 @@ class Decider extends AbstractHelper {
      * @param int $rolloutPercentage
      * @return bool
      */
-    private function _isInBucket(string $switchName, int $rolloutPercentage) {
+    private function _isInBucket(string $switchName, int $rolloutPercentage)
+    {
         $this->_session->start();
         $boltFeatureSwitchId = $this->_session->getBoltFeatureSwitchId();
         if (!$boltFeatureSwitchId) {
@@ -104,7 +107,8 @@ class Decider extends AbstractHelper {
         return $position < $rolloutPercentage;
     }
 
-    private function _switchFromConst($switchDef) {
+    private function _switchFromConst($switchDef)
+    {
         $switch = $this->_fsFactory->create();
         $switch->setName($switchDef[Definitions::NAME_KEY]);
         $switch->setValue($switchDef[Definitions::VAL_KEY]);
@@ -121,7 +125,8 @@ class Decider extends AbstractHelper {
      * @throws LocalizedException
      * @return bool
      */
-    public function isSwitchEnabled($switchName) {
+    public function isSwitchEnabled($switchName)
+    {
         $defaultDef = isset(Definitions::DEFAULT_SWITCH_VALUES[$switchName]) ?
             Definitions::DEFAULT_SWITCH_VALUES[$switchName] : null;
         if (!$defaultDef) {
@@ -156,7 +161,13 @@ class Decider extends AbstractHelper {
      * Switch Helpers below
      ***************************************************/
 
-    public function isSampleSwitchEnabled() {
+    public function isSampleSwitchEnabled()
+    {
         return $this->isSwitchEnabled(Definitions::M2_SAMPLE_SWITCH_NAME);
+    }
+
+    public function isTrackingSaveEventsEnabled()
+    {
+        return $this->isSwitchEnabled(Definitions::TRACKING_EVENT_OBSERVER_SWITCH);
     }
 }
--- a/Helper/FeatureSwitch/Decider.php
+++ b/Helper/FeatureSwitch/Decider.php
@@ -161,11 +161,17 @@ class Decider extends AbstractHelper
      * Switch Helpers below
      ***************************************************/
 
+    /**
+     * Checks whether the sample feature switch is enabled
+     */
     public function isSampleSwitchEnabled()
     {
         return $this->isSwitchEnabled(Definitions::M2_SAMPLE_SWITCH_NAME);
     }
 
+    /**
+     * Checks whether the feature switch for saving/sending tracking data is enabled
+     */
     public function isTrackingSaveEventsEnabled()
     {
         return $this->isSwitchEnabled(Definitions::TRACKING_EVENT_OBSERVER_SWITCH);
--- a/Helper/FeatureSwitch/Definitions.php
+++ b/Helper/FeatureSwitch/Definitions.php
@@ -21,7 +21,7 @@ namespace Bolt\Boltpay\Helper\FeatureSwitch;
 /**
  * All feature switches are defined here. By default all new switches should be
  * false and not rolled out. These values are only fallbacks for when something
- * isn't defined on the Bolt side. Anything values coming from Bolt take presedence.
+ * isn't defined on the Bolt side. Anything values coming from Bolt take precedence.
  *
  * Class Definitions
  * @package Bolt\Boltpay\Helper\FeatureSwitch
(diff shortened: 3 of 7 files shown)


Helper/TrackingSaveObserver.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+namespace Bolt\Boltpay\Observer;
```
matt-thomason: Namespace doesn't seem to match filepath.

Helper/TrackingSaveObserver.php
```diff
@@ -0,0 +1,75 @@
...
+use Bolt\Boltpay\Helper\Api as ApiHelper;
+use Magento\Framework\Event\ObserverInterface;
+use Magento\Framework\Event\Observer;
+use Bolt\Boltpay\Helper\Config as ConfigHelper;
+use Magento\Framework\DataObjectFactory;
+
+/**
+ * Class TrackingSaveObserver
+ *
+ * @package Bolt\Boltpay\Observer\Adminhtml\Sales
```
matt-thomason: Old package from `OrderCreateProcessDataObserver`

Helper/TrackingSaveObserver.php
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * @param Observer $observer
+     */
+    public function execute(Observer $observer)
+    {
+        $tracking = $observer->getEvent()->getTrack();
+        $shipment = $tracking->getShipment();
+        $order = $shipment->getOrder();
+        $payment = $order->getPayment();
+        $transactionReference = $payment->getAdditionalInformation('transaction_reference');
```
matt-thomason: This appears to work but I have the feeling there should be an easier way to work backwards to the transaction reference from here. Is there a helper I'm forgetting @danac-gs @YevhenShBolt ?

If there isn't perhaps we should create it.

Observer/TrackingSaveObserver.php
```diff
@@ -0,0 +1,75 @@
...
+
+        //Request Data
+        $requestData = $this->dataObjectFactory->create();
+        $requestData->setApiData($trackingData);
+        $requestData->setDynamicApiUrl(ApiHelper::API_CREATE_TRACKING);
+        $requestData->setApiKey($apiKey);
+
+        //Build Request
+        $request = $this->apiHelper->buildRequest($requestData);
+        $this->apiHelper->sendRequest($request);
```
danac-gs: I see a potential problem here.
Observer execution in Magento is sequential (synchronous) and so is this call (via Zend_Http_Client). Side effects:
1. waiting for response from Bolt in order to continue Magento processing
2. breaks the Magento flow in case of an error (exception thrown in \Bolt\Boltpay\Helper\Api::sendRequest)
Depending on how important this data is we could:
1. find a way to send asynchronous request or
2. wrap it in try / catch and decide how to approach the error
3. maybe we can do this similar to how the metrics is sent - log requests to file and send batches on interval, leaving  unsent on error

Helper/TrackingSaveObserver.php
danac-gs: Not sure there's a shorter path to transaction_reference in this case.

Observer/TrackingSaveObserver.php
akulkapoor: I'm new to the codebase and so I could totally be wrong, but why would this cause an issue if there's an error? We're just subscribing to an event so I would assume there would be no side effects even if an error ocurred

danac-gs: @akulkapoor I was trying to find some documentation regarding event dispatching and observer execution in M2 but could not find any real quick to support my point. So, based on experience, events can be fired somewhere, anywhere in the code and at that point a sequence of observers start to execute, synchronously, one by one, until the last one finishes and the execution continues from where it was "interrupted". If an exception is thrown and not caught in an observer (or around the the event) the execution stops.
@matt-thomason @YevhenShBolt @johnmccombs1 @njfaries please correct me if I am wrong.

johnmccombs1: @akulkapoor I agree with Daniel. If there is an error thrown, the execution will break. I think we should at least add a try/catch to keep from breaking the execution in case there is an exception thrown in the new event that you added.

akulkapoor: got it, will add something

akulkapoor: thanks for the explanation!

Observer/TrackingSaveObserver.php
```diff
@@ -0,0 +1,94 @@
...
+
+            //Request Data
+            $requestData = $this->dataObjectFactory->create();
+            $requestData->setApiData($trackingData);
+            $requestData->setDynamicApiUrl(ApiHelper::API_CREATE_TRACKING);
+            $requestData->setApiKey($apiKey);
+
+            //Build Request
+            $request = $this->apiHelper->buildRequest($requestData);
+            $this->apiHelper->sendRequest($request);
```
roopakv: let us grab the response, see if it is a 200, if it is not a 200 we should bump a metric and/or log

Observer/TrackingSaveObserver.php
```diff
@@ -0,0 +1,94 @@
...
+        $this->apiHelper = $apiHelper;
+        $this->bugsnag = $bugsnag;
+    }
+
+    /**
+     * @param Observer $observer
+     */
+    public function execute(Observer $observer)
+    {
+        try {
```
roopakv: lets use a feature switch to wrap all of this especially since we are listening to all trackevents.

we also probably want to track only bolt orders?

akulkapoor: agreed, is there an easy way to just track bolt orders?

akulkapoor: this is resolved apart from only tracking bolt orders. Not sure how that would work

akulkapoor: ended up checking for transaction_referenc to see if this was a bolt order

Helper/Api.php
```diff
@@ -291,8 +297,8 @@ public function buildRequest($requestData)
         $requestMethod = empty($requestData->getRequestMethod()) ? 'POST' : $requestData->getRequestMethod();
         $apiUrl        = $this->getFullApiUrl($dynamicApiUrl);
 
-        $headers    = (array)$requestData->getHeaders();
-        $statusOnly = (bool)$requestData->getStatusOnly();
+        $headers    = (array) $requestData->getHeaders();
```
roopakv: mostly just curious: is this some PHPCS thing? whats the official guideline on these spaces? @daisy1754 or @matt-thomason

Helper/FeatureSwitch/Definitions.php
```diff
@@ -25,25 +26,27 @@
...
-    );
-}
\ No newline at end of file
+  const DEFAULT_SWITCH_VALUES = array(
+    self::M2_SAMPLE_SWITCH_NAME =>  array(
+      self::NAME_KEY            => self::M2_SAMPLE_SWITCH_NAME,
+      self::VAL_KEY             => true,
+      self::DEFAULT_VAL_KEY     => false,
+      self::ROLLOUT_KEY         => 0
+    )
```
roopakv: you need to add a definition with the base values for `TRACKING_EVENT_OBSERVER_SWITCH` similar to `M2_SAMPLE_SWITCH_NAME`

Test/Unit/Observer/TrackingSaveObserverTest.php
```diff
@@ -0,0 +1,198 @@
...
+            $this->bugsnag,
+            $this->metricsClient,
+            $this->decider
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function testExecute()
```
roopakv: also add a test to see that it does nothing if the feature switch is off?

Observer/TrackingSaveObserver.php
```diff
@@ -0,0 +1,131 @@
...
+        if (!$this->decider->isTrackingSaveEventsEnabled()) {
+            return;
+        }
+
+        try {
+            $startTime = $this->metricsClient->getCurrentTime();
+            $tracking = $observer->getEvent()->getTrack();
+            $shipment = $tracking->getShipment();
+            $order = $shipment->getOrder();
+            $payment = $order->getPayment();
```
danac-gs: @akulkapoor a better way of checking if it is not a bolt payment (or no payment at all):
```php
!$payment || $payment->getMethod() != \Bolt\Boltpay\Model\Payment::METHOD_CODE
```

Helper/Api.php
```diff
@@ -174,7 +180,7 @@ private function getFullApiUrl($dynamicUrl)
      */
     private function isResponseError($response)
```
YevhenShBolt: Where is we are using this method?

Helper/Api.php
```diff
@@ -174,7 +180,7 @@ private function getFullApiUrl($dynamicUrl)
...
     {
-        $arr = (array)$response;
+        $arr = (array) $response;
```
YevhenShBolt: What is the class or type of `$response` variable?

If `$reponse` extend from `Magento\Framework\DataObject` so we can call
`$response->getErrors()` (or `$response->hasErrors()` depends from object) and `$response->getErrorCode()`.

I think it will be more correct instead of convert object to array.

Helper/FeatureSwitch/Decider.php
```diff
@@ -156,7 +161,13 @@ public function isSwitchEnabled($switchName) {
      * Switch Helpers below
      ***************************************************/
 
-    public function isSampleSwitchEnabled() {
+    public function isSampleSwitchEnabled()
```
YevhenShBolt: please add PHPDoc Comment here.

Helper/FeatureSwitch/Decider.php
```diff
@@ -156,7 +161,13 @@ public function isSwitchEnabled($switchName) {
...
+    {
         return $this->isSwitchEnabled(Definitions::M2_SAMPLE_SWITCH_NAME);
     }
-}
\ No newline at end of file
+
+    public function isTrackingSaveEventsEnabled()
```
YevhenShBolt: please add PHPDoc Comment here.

Helper/Api.php
akulkapoor: we're not, my linter just made changes

Helper/Api.php
akulkapoor: I don't want to change the functionality here in this PR, this is again my linter just making changes

Helper/FeatureSwitch/Decider.php
```diff
@@ -156,7 +161,19 @@ public function isSwitchEnabled($switchName) {
      * Switch Helpers below
      ***************************************************/
 
-    public function isSampleSwitchEnabled() {
+    /**
+     * Checks whether the sample feature switch is enabled
+     */
```
YevhenShBolt: Please add `@return bool` to PHPDoc Comment

Helper/FeatureSwitch/Decider.php
```diff
@@ -156,7 +161,19 @@ public function isSwitchEnabled($switchName) {
...
+    public function isSampleSwitchEnabled()
+    {
         return $this->isSwitchEnabled(Definitions::M2_SAMPLE_SWITCH_NAME);
     }
-}
\ No newline at end of file
+
+    /**
+     * Checks whether the feature switch for saving/sending tracking data is enabled
+     */
```
YevhenShBolt: Please add `@return bool` to PHPDoc Comment

Helper/Api.php
matt-thomason: This isn't explicitly covered by PSR-2, and that's why our codesniffer doesn't care either way here. Personally I like the spaces, as it better matches what we expect in control structures with parens/braces, and with things like closures.
https://www.php-fig.org/psr/psr-2/

matt-thomason: Though since it's the linter making changes, perhaps it's an addition to MEQP-2 or I'm misreading the spec. Either way, I like this 👍

Helper/FeatureSwitch/Decider.php
roopakv: @akulkapoor this please :)

Helper/FeatureSwitch/Decider.php
roopakv: this too

danac-gs:
@akulkapoor what's the status here? Should this be merged soon?

matt-thomason:
@akulkapoor This is pretty stale now and has conflicts. I'm going to close this for now - please reopen and re-review after fixing conflicts if this is still needed.

PR #429, opened 2019-11-10
Load Test M2 Plugin 

kajan-bolt:
# Description
Load tests that will test how the M2 plugin handles high loads. 

These load tests will test the user flow through checkout: 
- Getting to the cart page (which creates the order token and order reference)
- Calling the shipping and tax web hook set up on the merchant's site
- Calling the Pre-auth web hook set up on the merchant's site 

The purpose behind these load tests is to see whether our plugin is the actual problem when a merchant has a significant number of users accessing their store. 

Fixes: https://app.asana.com/0/941920570700290/1149771740816851/f

\#changelog

# Type of change

- [ x ] New feature (change which adds functionality)

# How Has This Been Tested?

- [ x ] Successfully tested locally (or docker image)
- [ x ] Successfully tested on a staging or sandbox server

# Checklist:

- [ x ] My code follows the style guidelines of this project.
- [ x ] I have performed a self-review of my own code.

### DIFF
--- a/Test/Load/config.js
+++ b/Test/Load/config.js
@@ -1,0 +1,15 @@
+import crypto from "k6/crypto";
+
+export const URL = "http://34.220.243.58";
+export const CART_PATH = "/rest/V1/bolt/boltpay/order/cartdata";
+export const SHIPPING_PATH = "/rest/V1/bolt/boltpay/shipping/methods";
+export const PREAUTH_PATH = "/rest/V1/bolt/boltpay/order/create";
+const SIGNING_SECRET = "76037c6eeaf7005580f4541e3e5397017e153d5f7ce97ae30656f8a830c4908f";
+
+export function create_header( data ) {
+    const hmac = crypto.hmac( 'sha256', SIGNING_SECRET, JSON.stringify( data ), 'base64' );
+    return {
+        "X-Bolt-Hmac-Sha256": hmac,
+        "Content-Type": "application/json"
+    };
+}
--- a/Test/Load/config.js
+++ b/Test/Load/config.js
@@ -1,9 +1,21 @@
 import crypto from "k6/crypto";
 
+// The constants below should be replaced with the
+// values from the store you want to load test
+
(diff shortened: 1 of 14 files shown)


Test/Load/inputs/shippingtax_payload.js
```diff
@@ -0,0 +1,12 @@
+import { address } from "./address.js";
+import { create_shippingtax_cart } from "./shippingtax_cart.js";
+
+export function create_shipping_info( order_token, order_reference ) {
+    return {
+        "order_token": order_token,
+        "cart": create_shippingtax_cart( order_reference, false ),
+        "shipping_address": address,
+        "shipping_options": [],
```
YevhenShBolt: @kajan-bolt, Can you add more details why we have empty `shipping_options`?
It is not clear for me.

Test/Load/config.js
```diff
@@ -0,0 +1,15 @@
+import crypto from "k6/crypto";
+
+export const URL = "http://34.220.243.58";
```
matt-thomason: Can you add comments here similar to the ones added for the M1 PR?

Test/Load/inputs/shippingtax_payload.js
kajan-bolt: This code is supposed to be the request that is sent to the shipping_and_tax webhook that is on the merchant's site. And from what I can see on DataDog, most requests are sent to the shipping_and_tax webhook with an empty `shipping_options` (since the shipping options won't be available to the Bolt Server until we get the response from the shipping_and_tax webhook)

Test/Load/config.js
kajan-bolt: Sounds good, I added in the comments

kajan-bolt:
@matt-thomason I could add in the non-preauth flow, but is the only difference between the two flows the final pre-auth request? If so would it make a difference for load testing, because if we can handle the current flow then we should be able to handle the non-preauth flow?

PR #430, opened 2019-11-11
Fix Bug - the invoice is created with no item

dumega:
# Description
When creating order invoice, the plugin would first compare ($order->getTotalInvoiced() + $amount) with $order->getGrandTotal() , but neither of the two values get rounded before comparison, so for some instance, eg. $order->getTotalInvoiced() + $amount = 121.64 and $order->getGrandTotal() = 121.6406, they are not equal and we call function $this->invoiceService->prepareInvoiceWithoutItems($order, $amount); to create inoivce without item.

Solution: round the value before comparison.

Fixes: https://app.asana.com/0/941920570700290/1148920165401345/f

\#changelog

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -1666,7 +1666,7 @@ class Order extends AbstractHelper
     private function createOrderInvoice($order, $transactionId, $amount)
     {
         try {
-            if ($this->cartHelper->getRoundAmount($order->getTotalInvoiced() + $amount) === $this->cartHelper->getRoundAmount($order->getGrandTotal())) {
+            if (round(($order->getTotalInvoiced() + $amount), 2) == round($order->getGrandTotal(), 2)) {
                 $invoice = $this->invoiceService->prepareInvoice($order);
             } else {
                 $invoice = $this->invoiceService->prepareInvoiceWithoutItems($order, $amount);
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -1666,7 +1666,7 @@ class Order extends AbstractHelper
     private function createOrderInvoice($order, $transactionId, $amount)
     {
         try {
-            if (round(($order->getTotalInvoiced() + $amount), 2) == round($order->getGrandTotal(), 2)) {
+            if ($this->cartHelper->getRoundAmount($order->getTotalInvoiced() + $amount) === $this->cartHelper->getRoundAmount($order->getGrandTotal())) {
                 $invoice = $this->invoiceService->prepareInvoice($order);
             } else {
                 $invoice = $this->invoiceService->prepareInvoiceWithoutItems($order, $amount);
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -1666,7 +1666,7 @@ class Order extends AbstractHelper
     private function createOrderInvoice($order, $transactionId, $amount)
     {
         try {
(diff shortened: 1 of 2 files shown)


dumega:
@danac-gs I have added the test

Helper/Order.php
```diff
@@ -1666,7 +1666,7 @@ public function updateOrderPayment($order, $transaction = null, $reference = nul
     private function createOrderInvoice($order, $transactionId, $amount)
     {
         try {
-            if ($order->getTotalInvoiced() + $amount == $order->getGrandTotal()) {
+            if (round(($order->getTotalInvoiced() + $amount), 2) == round($order->getGrandTotal(), 2)) {
```
YevhenShBolt: Why not strict comparison?
`
if (round(($order->getTotalInvoiced() + $amount), 2) === round($order->getGrandTotal(), 2))
`

Test/Unit/Helper/OrderTest.php
```diff
@@ -737,4 +771,34 @@ public function getTransactionState_APMInitialAuthorized()
...
     }
+    
+    /**
+     * @test
+     * @covers ::createOrderInvoice
+     */
+    public function createOrderInvoice_amountWithDifferentDecimals() {
+        $totalInvoiced = 0;
+        $amount = 12.64;
+        $grandTotal = 12.6402;
```
YevhenShBolt: I guess you can add here a `@dataProvider someVariable` to manage several cases.
as example we can check behaviour like
    * `$amount = 12`;
    * `$grandTotal = 12.00`;
or
    * `$amount = '12'`;
    * `$grandTotal = 12.00`;

and etc

Helper/Order.php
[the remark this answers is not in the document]
oxigenka: Thanks

oxigenka:
LGTM, but the only thing: please fix the alignments in the data provider method

oxigenka:
It seems something has been missed in the list of mocked methods

dumega:
@YevhenShBolt  @oxigenka  as we can see in the failed circle tests (https://github.com/BoltApp/bolt-magento2/pull/430/commits/c850b0bf5baf2739b9b771b47f85fea7bf6dae42), only `(round(($order->getTotalInvoiced() + $amount), 2) == round($order->getGrandTotal(), 2))` can get desired results, while `($this->cartHelper->getRoundAmount($order->getTotalInvoiced() + $amount) === $this->cartHelper->getRoundAmount($order->getGrandTotal()))` would fail with different amount.

YevhenShBolt:
@dumega for the first look on unit test i see such error:
`Error: Call to a member function setRequestedCaptureCase() on null`
I guess need to refactor/fix test for.

@oxigenka @dumega What do you think?

oxigenka:
@YevhenShBolt Could it be a cause of misfunction the `cartHelper->getRoundAmount` method? I don't think so. Although, surely, this also needs to be fixed.
Also, I'm wondering why the correct and widely used rounding method in `cartHelper` doesn't work as designed?

dumega:
@YevhenShBolt @oxigenka well, I have fixed the tests, we need to stub the method `getRoundAmount` of cartHelper to return proper values.

PR #456, opened 2019-12-09
Add tests for Controller/Order/Save

njfaries:
# Description
Added tests for `Controller/Order/Save.php`. Some methods were changed from `private` to `protected` to make testing easier. Also added some coverage ignore decorators to methods in `ReceivedUrl.php` that were performing no logic, just retrieving things.

Fixes: https://app.asana.com/0/1131034199022718/1137538625478187/f

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Controller/Order/Save.php
+++ b/Controller/Order/Save.php
@@ -138,7 +138,7 @@ class Save extends Action
      * @param Quote $quote
      * @return void
      */
-    private function replaceQuote($quote)
+    protected function replaceQuote($quote)
     {
         $this->checkoutSession->replaceQuote($quote);
     }
@@ -150,7 +150,7 @@ class Save extends Action
      *
      * @return void
      */
-    private function clearQuoteSession($quote)
+    protected function clearQuoteSession($quote)
     {
         $this->checkoutSession->setLastQuoteId($quote->getId())
                               ->setLastSuccessQuoteId($quote->getId())
@@ -164,7 +164,7 @@ class Save extends Action
      *
      * @return void
      */
-    private function clearOrderSession($order)
+    protected function clearOrderSession($order)
     {
         $this->checkoutSession->setLastOrderId($order->getId())
                               ->setLastRealOrderId($order->getIncrementId())
--- a/Controller/Order/Save.php
+++ b/Controller/Order/Save.php
@@ -138,7 +138,7 @@ class Save extends Action
      * @param Quote $quote
      * @return void
      */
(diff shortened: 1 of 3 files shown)


Controller/Order/Save.php
```diff
@@ -138,7 +138,7 @@ public function execute()
      * @param Quote $quote
      * @return void
      */
-    private function replaceQuote($quote)
+    protected function replaceQuote($quote)
```
danac-gs: Although this is a handy way to solve the problem (and a sort of accepted practice in M2 to use protected methods for easier testing) it is not a good practice in general OOP. You could probably leave the methods as private and set the expectations on `$this->checkoutSession` object methods. Anyway, it is a matter of preference, so I am not requesting the change.

Test/Unit/Controller/Order/SaveTest.php
```diff
@@ -0,0 +1,228 @@
...
+            ->disableOriginalConstructor()
+            ->getMock();
+        $jsonFactory->method('create')->willReturn($json);
+
+        $url = $this->createMock(UrlInterface::class);
+        $url->method('getUrl')->willReturn(self::SUCCESS_URL);
+        $this->context->method('getUrl')->willReturn($url);
+
+        $this->configHelper->method('getSuccessPageRedirect');
+
```
danac-gs: This seems unfinished. You should probably set a return value here (checkout/onepage/success ?) and set the same as the expectation to the `$url->method('getUrl')` above.

njfaries:
Rebased on master, there were some config changes that didn't get to this branch that were causing circleCI failures

Test/Unit/Controller/Order/SaveTest.php
njfaries: The method is actually configured in the init method, its presence here was an un-removed artifact of writing this. Removing it.

Controller/Order/Save.php
njfaries: Reviewing this, it makes a lot more sense this way. @matt-thomason and I were trying to get this to work before as the `expected` calls refused to work properly and after trial and error we came to this solution. However, without the expectations set on the methods called inside the `private`/`protected` methods the contents aren't actually being tested. Functionally for the top level method being tested (`execute`), it doesn't really matter which private method these `checkoutSession` methods are called it, just that they are called at all.

So with that all said, I'm switching in line with @danac-gs's suggestions because it does actually fix the testing coverage.

PR #457, opened 2019-12-09
Add description field to the quote item collection.

YevhenShBolt:
# Description
Payment transactions processed via Stripe require a field "description" for level 3 processing. Now this field is missing in the quote item collection

Fixes: https://app.asana.com/0/941920570700290/1153065154514665/f

\#changelog

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/etc/catalog_attributes.xml
+++ b/etc/catalog_attributes.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0"?>
+<!--
+/**
+ * Bolt magento2 plugin
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ *
+ * @category   Bolt
+ * @package    Bolt_Boltpay
+ * @copyright  Copyright (c) 2018 Bolt Financial, Inc (https://www.bolt.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+-->
+<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Catalog:etc/catalog_attributes.xsd">
+    <group name="quote_item">
+        <attribute name="description"/>
+    </group>
+</config>


nirajjayantbolt:
Any way we can write unit tests for this? Might not be possible but just curious

nirajjayantbolt:
@YevhenShBolt also can you explain to me what this is doing? I assume its adding description to the attributes of these items but where is the code that will actually send this information to Bolt when we create the order?

isklyarenko:
> @YevhenShBolt also can you explain to me what this is doing? I assume its adding description to the attributes of these items but where is the code that will actually send this information to Bolt when we create the order?

@nirajjayantbolt such code was added before in Helper/Cart.php, line 1048, but it always returns null since the field "description" isn't in the collection

`$product['description'] = strip_tags($_product->getDescription());`

nirajjayantbolt:
@isklyarenko thanks!

danac-gs:
@YevhenShBolt 
After a second look I realize that we are loading the description from the underlying product object, not from the item itself. Would it be reasonable / safe to add another, `catalog_product`, group with the same attribute, like it is done in `vendor/magento/module-bundle/etc/catalog_attributes.xml`?

nirajjayantbolt:
@isklyarenko once we merge this, how quickly can we get this to Lund?

YevhenShBolt:
@danac-gs Can you clarify your question? It is not clear for me what you suggest.
>Would it be reasonable / safe to add another, catalog_product, group with the same attribute, like it is done in vendor/magento/module-bundle/etc/catalog_attributes.xml?

danac-gs:
> @danac-gs Can you clarify your question? It is not clear for me what you suggest.
> 
> > Would it be reasonable / safe to add another, catalog_product, group with the same attribute, like it is done in vendor/magento/module-bundle/etc/catalog_attributes.xml?

Probably unnecessary safeguarding. I was thinking something like this:

```xml
<group name="catalog_product">
    <attribute name="description"/>
</group>
<group name="quote_item">
    <attribute name="description"/>
</group>
```
And perhaps even
```php
$product['description'] = strip_tags($_product->getDescription() ?: $item->getDescription());
```

matt-thomason:
@YevhenShBolt @danac-gs Resurfacing this - are we going to go ahead with @danac-gs's suggestion?

isklyarenko:
@matt-thomason  sorry for the delay here

@danac-gs When creating a cart item we only work with a quote item and don't use a catalog product directly for it (please see \Bolt\Boltpay\Helper\Cart::getCartItems). On the other hand, the product description can be large and we should avoid its loading everywhere until we don't need it.
Therefore I would prefer to keep loading of the description in the quote item only

danac-gs:
@njfaries @matt-thomason @isklyarenko @YevhenShBolt can we merge this?

YevhenShBolt:
i think yes

PR #458, opened 2019-12-09
Ee gift cards fix

danac-gs:
# Description
_"Customers have the option to buy giftcards. They have been fine until they turned on Pre-Auth. Magento has a gift card setting called "Generate Gift card Acount when order is" there are two options (Ordered or Invoiced).
This feature creates the order in the store before moving to bolt. Long before they used bolt we had the Gift Card feature set to Invoiced due to a Bug in Magento, causing duplicate gift cards getting created for every order.
So when we turned the pre-auth condition on what ended up happening was it enabled a loop hole in the Magento software that allowed people to get infinite gift cards for free no matter the value."_

Solution: Prevent gift card creation until the order is approved from bolt (order status not pending_payment or canceled)

Fixes: https://app.asana.com/0/941920570700290/1151947407395393/f

\#changelog Prevent gift card creation until the order is approved from Bolt

# Type of change

# How Has This Been Tested?

# Checklist:


Plugin/Magento/GiftCard/GenerateGiftCardAccountsOrderPlugin.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    public function aroundExecute(
+        $subject,
+        callable $proceed,
+        Observer $observer
+    ) {
+        $event = $observer->getEvent();
+        /** @var Order $order */
+        $order =  $event->getOrder();
+        if ($order->getPayment()->getMethod() != BoltPayment::METHOD_CODE ||
```
matt-thomason: Having the check here is good, but I still would like to hear about your testing here. Did you test both Bolt-enabled and Bolt-disabled paths here to make sure we didn't modify functionality for non-Bolt orders?

Test/Unit/Plugin/Magento/GiftCard/GenerateGiftCardAccountsOrderPluginTest.php
```diff
@@ -0,0 +1,124 @@
...
+
+        $this->callback->expects($this->$expects())->method('__invoke');
+
+        $this->plugin->aroundExecute($this->subject, $this->proceed, $this->observer);
+    }
+
+    public function methodStatusExpects()
+    {
+        return [
+            [self::OTHER_METHOD, Order::STATE_PENDING_PAYMENT, self::EXPECTS_ONCE],
```
matt-thomason: Great tests

Plugin/Magento/GiftCard/GenerateGiftCardAccountsOrderPlugin.php
danac-gs: @matt-thomason The \Magento\GiftCard\Observer\GenerateGiftCardAccountsOrder observer listens to sales_order_save_after event, which fires multiple times for every order. In our plugin we suppress its execution only in cases when payment method is Bolt in combination with pending_payment or canceled order statuses. Maybe it would be more clear I had written it like this:
```php
if ($order->getPayment()->getMethod() === BoltPayment::METHOD_CODE && 
    in_array($order->getStatus(), [Order::STATE_PENDING_PAYMENT,Order::STATE_CANCELED])
) {
    return;
}
return $proceed($observer);
```
Anyway, I did not test with Bolt disabled.

danac-gs:
>  Assuming corresponding testing - would love to hear your manual testing plan.
> 
> I modified your description to add this to the changelog - so this is highlighted for manual QA testing on plugin release.

@matt-thomason 
1. Configure Gift Card to be created on order creation (if set on invoice then there's no potential issue)
2. Add a gift card to cart and make a successful Bolt payment - after a while you should verify that:
   2.1 Gift card is created in Magento backend
   2.2 You receive a gift card code email
3. Add a gift card to cart and make a Bolt payment that fails, resulting in failed_payment hook - after a while you should verify that:
   3.1 Gift card is NOT created in Magento backend
   3.2 NO gift card code email is received

Plugin/Magento/GiftCard/GenerateGiftCardAccountsOrderPlugin.php
danac-gs: Updated as above.

matt-thomason: Thanks @danac-gs - I did not think the previous code was unclear though. I just wanted to confirm what you manually tested. Could you do a manual test with a non-Bolt payment method and confirm back here? Testing with Bolt disabled probably is overkill, I agree.

danac-gs: @matt-thomason I've checked, it works. Bought a GC with check / money order payment and it went through. Please re-approve.

Plugin/Magento/GiftCard/GenerateGiftCardAccountsOrderPlugin.php
```diff
@@ -0,0 +1,39 @@
...
+    public function aroundExecute(
+        $subject,
+        callable $proceed,
+        Observer $observer
+    ) {
+        $event = $observer->getEvent();
+        /** @var Order $order */
+        $order =  $event->getOrder();
+        if ($order->getPayment()->getMethod() === BoltPayment::METHOD_CODE &&
+            in_array($order->getStatus(), [Order::STATE_PENDING_PAYMENT, Order::STATE_CANCELED])) {
```
YevhenShBolt: @danac-gs , as i see a class `\Magento\GiftCard\Observer\GenerateGiftCardAccountsOrder` call on event `sales_order_save_after`.
So, as i see the aroundPlugin functionality calling every time on order save. 

I guess we add here a status `Order::STATE_CLOSED` and `Order::STATE_HOLDED`, regarding that we already checking `Order::STATE_CANCELED`.

What do you think?

danac-gs: @YevhenShBolt what we are trying to achieve here is to prevent creating gift cards when order record is created in pre-auth call, before passing the approval at Bolt (successful authorize call). Initially the order is created in pending_payment state, and we have two cases to cover: 
1. Failed payment - failed_payment hook triggers order deletion, no new state after pending_payment
2. Blacklisted cc - rejected_irreversible hook triggers order cancellation - new and final order state is canceled
Anything after pending_payment that is not canceled should not be a problem. After that point the order lifetime is similar as with other payment methods. We could be more strict but I don't think we need to. In the end we could allow gift card creation only if the order state is processing, but I did not want to alter the system more than necessary.

PR #459, opened 2019-12-10
List all credit cards in the admin customer detail page

johnmccombs1:
# Description
List all credit cards in the admin customer detail page
See the demo screenshot: https://www.screencast.com/t/y9tn5BXLW
Fixes: https://app.asana.com/0/564264490825835/1152758055340089

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Model/CustomerCreditCard.php
+++ b/Model/CustomerCreditCard.php
@@ -126,4 +126,29 @@ class CustomerCreditCard extends AbstractModel
 
         return $response;
     }
+
+    /**
+     * @return object
+     */
+    public function getCardInfoObject(){
+        $cardInfoArray = json_decode($this->getCardInfo(), true);
+        $cartInfo = $this->dataObjectFactory->create();
+        $cartInfo->setData($cardInfoArray);
+
+        return $cartInfo;
+    }
+
+    /**
+     * @return string
+     */
+    public function getCardType(){
+        return $this->getCardInfoObject()->getData('display_network');
+    }
+
+    /**
+     * @return string
+     */
+    public function getCardLast4Digit(){
+        return ($last4 = $this->getCardInfoObject()->getData('last4')) ? 'XXXX-'.$last4 : '';
+    }
 }
--- a/Block/Adminhtml/Customer/CreditCard/Tab/CreditCard.php
+++ b/Block/Adminhtml/Customer/CreditCard/Tab/CreditCard.php
@@ -1,0 +1,125 @@
+<?php
+/**
+ * Bolt magento2 plugin
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ *
+ * @category   Bolt
+ * @package    Bolt_Boltpay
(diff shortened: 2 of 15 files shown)


Model/CustomerCreditCard.php
```diff
@@ -126,4 +126,25 @@ public function recharge(OrderModel $order)
...
+     */
+    public function getCardInfoObject(){
+        return json_decode($this->getCardInfo(), true);
+    }
+
+    /**
+     * @return string
+     */
+    public function getCardType(){
+        return @$this->getCardInfoObject()['display_network'];
```
YevhenShBolt: @johnmccombs1 Can you remove suppression here?

Model/CustomerCreditCard.php
```diff
@@ -126,4 +126,25 @@ public function recharge(OrderModel $order)
...
+    }
+
+    /**
+     * @return string
+     */
+    public function getCardLast4Digit(){
+        return ($last4 = @$this->getCardInfoObject()['last4']) ? 'XXXX-'.$last4 : '';
```
YevhenShBolt: @johnmccombs1 Can you remove suppression here?

Model/CustomerCreditCard.php
```diff
@@ -126,4 +126,25 @@ public function recharge(OrderModel $order)
 
         return $response;
     }
+
+    /**
+     * @return object | bool
+     */
+    public function getCardInfoObject(){
+        return json_decode($this->getCardInfo(), true);
```
YevhenShBolt: I suggest ta bit refactor it to avoid using suppressions in code below.
1. Inject to the constructor additional class`Magento\Framework\Serialize\Serializer\Json $serializer`
2.  As you already injected `DataObjectFactory` the code should look like:
```
$cardInfo = $this->serializer->unserialize($this->getCardInfo());
$data = $this->dataObjectFactory->create();
$data->setData('card_info',  $cardInfo);

return $data;
```

After that we does not need to use suppressions, because:
- In object does not exist method `display_network ` it will automatically return `null`
- We should not aware about production server settings regarding suppressions
- We have predictable code, no error suppression
- We pass Magento validator because we do not have suppressions

danac-gs: @YevhenShBolt @johnmccombs1 
`Magento\Framework\Serialize\Serializer\Json` is not available in M2.1.x that, I guess, we still need to support. We have active merchants on old Magento versions.
Keeping the plain old `json_decode` could be beneficial.
@matt-thomason @njfaries thoughts?

YevhenShBolt: @danac-gs @matt-thomason @njfaries - i checked and i agree with Daniel because the Magento 2.1.x do not support `Magento\Framework\Serialize\Serializer\Json` - we can easily change it to `json_decode($this->getCardInfo(), true)`

The main idea of my request is avoid suppressions and have _"clean and clear"_ look of object, instead of concatenate the object and array in one call and with suppression here.
As i see a bit above in class already used Magento's best-practice and already injected `dataObjectFactory ` to the constructor and using for work with request api object.
https://github.com/BoltApp/bolt-magento2/blob/ba9de4f0ebdb211b0ca986e917f11f2b69daa25c/Model/CustomerCreditCard.php#L101
https://github.com/BoltApp/bolt-magento2/blob/ba9de4f0ebdb211b0ca986e917f11f2b69daa25c/Model/CustomerCreditCard.php#L102

johnmccombs1: @YevhenShBolt I updated the code. Can you review again?

PR #461, opened 2019-12-13
Update capture logic to handle different currency

daisy1754:
# Description
Magento's Payment class pass currency in store currency. This PR adds logic to get capture amount in proper (order's) currency when calling Bolt's capture API

Fixes: https://app.asana.com/0/1145430477677801/1152698132445867/f

\#changelog support invoice creation in different currency

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Model/Payment.php
+++ b/Model/Payment.php
@@ -386,12 +386,10 @@ class Payment extends AbstractMethod
                 );
             }
 
-            $captureAmount = CurrencyUtils::toMinor($amount, "USD");
-
             //Get capture data
             $capturedData = [
                 'transaction_id' => $realTransactionId,
-                'amount'         => $captureAmount,
+                'amount'         => $this->getCaptureAmount($order, $amount),
                 'currency'       => $order->getOrderCurrencyCode(),
                 'skip_hook_notification' => true
             ];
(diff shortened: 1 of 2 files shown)


Model/Payment.php
```diff
@@ -430,6 +428,18 @@ public function capture(InfoInterface $payment, $amount)
         }
     }
 
+    private function getCaptureAmount($order, $amountInStoreCurrency)
+    {
+        $orderCurrency = $order->getOrderCurrencyCode();
+        if ($order->getStoreCurrencyCode() == $orderCurrency) {
+            return CurrencyUtils::toMinor( $amountInStoreCurrency, $orderCurrency );
+        } else {
+            // Magento passes $amount in store currency but not in order currency - we have to grab amount from invoice
```
danac-gs: @daisy1754 Does this work? The invoice should be created after the capture, not before, so we might not have the data needed. We can probably store the currency with payment, from bolt transaction (I suppose we have it there), by adding it to the $paymentData here: https://github.com/BoltApp/bolt-magento2/blob/master/Helper/Order.php#L1590

daisy1754: It worked for me when I tested against my server (2.3.2). Invoice existed but did not have ID at this point

daisy1754: We already have currency and convertion rate information in order but u would love to avoid doing own convertion because it can lead to rounding error and price mismatch

PR #466, opened 2019-12-15
Update refund to handle multi-currency

daisy1754:
# Description
Magento's refund method pass amount in store's currency. In order to handle different currency (eg US store that also sell things in CAD), this PR get currency from credit memo.

Fixes: https://app.asana.com/0/941920570700290/1130539126613441/f

\#changelog update refund to handle different currency

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Model/Payment.php
+++ b/Model/Payment.php
@@ -44,8 +44,6 @@ use Bolt\Boltpay\Helper\Bugsnag;
 use Bolt\Boltpay\Helper\MetricsClient;
 use Bolt\Boltpay\Helper\Cart as CartHelper;
 use \Magento\Sales\Model\Order\Payment\Transaction\Repository as TransactionRepository;
-use Magento\Setup\Exception;
-use u2flib_server\Error;
 
 /**
  * Class Payment.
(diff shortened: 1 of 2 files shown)


Model/Payment.php
```diff
@@ -468,13 +468,14 @@ public function refund(InfoInterface $payment, $amount)
                 );
             }
 
-            $refundAmount = CurrencyUtils::toMinor($amount, "USD");
+            $orderCurrency = $order->getOrderCurrencyCode();
+            // $amount argument of refund method is in store currency, we need to get amount from credit memo to get the value in order's currency.
+            $refundAmount = CurrencyUtils::toMinor( $payment->getCreditMemo()->getGrandTotal(), $orderCurrency );
```
roopakv: does the check on L459 still make sense then?

why not remove $amount as an argument then? it sort of seems unnecessary right?

daisy1754: I think it still serves as check against invalid call. Also we cannot remove $amount argument because this class is implementing magento's interface so method signature has to be this type

roopakv: ahh thank you for explaining

PR #492, opened 2020-01-03
Handle cart.create API call

vitaliyreznikov:
Handle cart.create API call. We need it for Product page checkout
Fixes: https://app.asana.com/0/0/1155249819640873/f

Restrictions
- for guest users only
- doesn't work for multistore

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -52,6 +52,7 @@ use Magento\Quote\Api\Data\CartInterface;
 use Magento\Framework\App\ResourceConnection;
 use Magento\Sales\Model\Order;
 use Magento\Store\Model\ScopeInterface;
+use Magento\Quote\Api\CartManagementInterface;
 
 /**
  * Boltpay Cart helper
@@ -206,6 +207,11 @@ class Cart extends AbstractHelper
     private $orderData;
 
     /**
+     * @var CartManagementInterface
+     */
+    private $quoteManagement;
+
+    /**
      * @param Context           $context
      * @param CheckoutSession   $checkoutSession
      * @param ProductRepository $productRepository
@@ -228,6 +234,7 @@ class Cart extends AbstractHelper
      * @param DiscountHelper $discountHelper
      * @param CacheInterface $cache
      * @param ResourceConnection $resourceConnection
+     * @param CartManagementInterface $quoteManagement
      *
      * @codeCoverageIgnore
      */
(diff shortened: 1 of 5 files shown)


Helper/Cart.php
```diff
@@ -1747,4 +1756,33 @@ public function hasProductRestrictions($quote = null)
...
+        $quote = $this->quoteFactory->create()->load($quoteId);
+
+        //add item in quote
+        $product = $this->productRepository->getbyId($item['reference']);
+        $quote->addProduct($product, $item['quantity']);
+
+        $quote->reserveOrderId();
+        $quote->collectTotals()->save();
+
+        $cart_data = $this->getCartData(false,'',$quote);
```
danac-gs: Something strange might happen here https://github.com/BoltApp/bolt-magento2/blob/master/Helper/Cart.php#L1133 and beyond in the getCartData method. I assume your code works but it's not clear how and why and what exactly happens in the aforementioned method. Perhaps you should add a check / use a immutableQuote quoteId instead of boltParentQuoteId when boltParentQuoteId is not set in the above line.

danac-gs:
@vitaliyreznikov 
1. Why are we using order management endpoint for creating the cart?
2. Is the order assigned to logged in customer?

Helper/Cart.php
vitaliyreznikov: @danac-gs `$quote` we created is immutable. We created it and nobody will change it. It's a reason while I pass `$quote` as the third parameter into getCartData. 

getCartData is already developed to work with the case, we can see in comments that it's for "i.e. discount code validation, bugsnag report generation".

We try to find the parent quote for $immutableQuote, but it doesn't exist so $quote is false.
Again, it's a normal situation for function, it designed to handle it.
Actually it just means we don't try to get address from parentQuote in this line
https://github.com/BoltApp/bolt-magento2/blob/0121c9ef21bb6e2eba03346288fd11c2f000f041/Helper/Cart.php#L1259

So it seems like it's enough to add a couple comments into getCartData to explain how it handles **cart.create** case. What do you think?

vitaliyreznikov:
@danac-gs 
1. I just copied approach we use in WooC https://github.com/BoltApp/bolt-woocommerce/blob/699448f95361e46cd6e12bd45c2cf1c7d9cb7fff/includes/bolt-payment-gateway-helpers.php#L104

Let me know if we need any refactoring there. For example, we can move method **createCartByItem** into a separate new class.

2. No, I plan to add support for logged in users in separate PR.

Helper/Cart.php
danac-gs: I looked again and you're right. This is OK.

danac-gs:
@vitaliyreznikov 
1. I assume the order management (web hook) endpoint is used out of a convenience. Already there, no need to add another configuration option in merchant dashboard / Bolt admin. We are moving towards having one endpoint for all API requests so I guess this is OK. createCartByItem should probably stay in Cart helper.

vitaliyreznikov:
@danac-gs Yes, method saveUpdateOrder I extracted is 100% covered y unit tests

PR #496, opened 2020-01-05
Add hints support for product page checkout

vitaliyreznikov:
Add hints support for product page checkout

Fixes: https://app.asana.com/0/0/1119824466711711/f

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Controller/Cart/Hints.php
+++ b/Controller/Cart/Hints.php
@@ -62,8 +62,9 @@ class Hints extends Action
     }
 
     /**
-     * @return void
-     * @throws Exception
+     * Get hints for Product page checkout
+     *
+     * @return Json
      */
     public function execute()
     {
--- a/Controller/Cart/Hints.php
+++ b/Controller/Cart/Hints.php
@@ -69,7 +69,7 @@ class Hints extends Action
     {
         try {
             $result = $this->resultJsonFactory->create();
-            $hints = $this->cartHelper->getHints(null, 'cart');
+            $hints = $this->cartHelper->getHints(null, 'product');
             $result->setData([
                 'hints' => $hints
             ]);
--- a/Controller/Cart/Hints.php
+++ b/Controller/Cart/Hints.php
@@ -1,0 +1,87 @@
+<?php
+/**
+ * Bolt magento2 plugin
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ *
+ * @category   Bolt
+ * @package    Bolt_Boltpay
+ * @copyright  Copyright (c) 2018 Bolt Financial, Inc (https://www.bolt.com)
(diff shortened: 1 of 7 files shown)


view/frontend/templates/button_product_page.phtml
```diff
@@ -256,7 +256,21 @@ $trackCallbacks = $block->getTrackCallbacks();
 
                 var buttonClass = "bolt-product-checkout-button";
                 var hints = {prefill:{}};
```
danac-gs: @vitaliyreznikov 
Perhaps easier and (to me at least) straightforward way to do this was to just inject hints into the template,
something like:
```javascript
var hints = '<?=$block->getHints();?>';
```
without the need of a controller.

vitaliyreznikov: @danac-gs Are we speaking about "empty" hints?
So do you propose to create `function getHints() { return "{prefill:{}}"; }` ?
In general, we need to receive Hints from server by Ajax call.

danac-gs: @vitaliyreznikov 
No. I mean the complete hints (if there are any). The hints value would be the same at the time the page renders as it would be later, when fetched via Ajax call. No need to receive it from server, can add it to the page when the page is generated on the server. We can only get hints for a logged in customer. No controller, less code to maintain. $block::getHints() that calls \Bolt\Boltpay\Helper\Cart::getHints and formats the data accordingly. Perhaps add another flag to Cart::getHints to prevent it to load the quote. It creates one otherwise ($this->checkoutSession->getQuote()).

vitaliyreznikov: @danac-gs Correct me if I wrong but product pages are cached, so we can't receive hints when the page renders.

danac-gs: @vitaliyreznikov 100% correct! Scratch everything I wrote above except the last two lines. You don't need a flag, you can pass different $checkoutType ('product') and skip fetching the quote.

vitaliyreznikov: @danac-gs Ok, sounds reasonable. Done.

Test/Unit/Block/JsTest.php
```diff
@@ -318,7 +318,7 @@ public function getSettings()
         $this->assertJson($result, 'The Settings config do not have a proper JSON format.');
 
         $array = json_decode($result, true);
-        $this->assertCount(17, $array, 'The number of keys in the settings is not correct');
+        $this->assertCount(18, $array, 'The number of keys in the settings is not correct');
```
ethanwayda22: this should be made into a global constant to easily keep track of

vitaliyreznikov: Done

Controller/Cart/Hints.php
```diff
@@ -0,0 +1,87 @@
...
+        CartHelper $cartHelper
+    ) {
+        parent::__construct($context);
+        $this->resultJsonFactory = $resultJsonFactory;
+        $this->cartHelper = $cartHelper;
+        $this->bugsnag = $bugsnag;
+    }
+
+    /**
+     * @return void
```
matt-thomason: This phpdoc is inaccurate

PR #500, opened 2020-01-07
Feature switch to manage slow roll out

roopakv:
# Description

This PR will allow slow roll out of bolt, controlling it via a feature switch.

#changelog Allows slow roll out of Bolt

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/FeatureSwitch/Definitions.php
+++ b/Helper/FeatureSwitch/Definitions.php
@@ -38,12 +38,24 @@ class Definitions {
      */
     const M2_SAMPLE_SWITCH_NAME = "M2_SAMPLE_SWITCH";
 
+    /**
+     * Whether bolt is enabled. This will be used for slow rollout / controlling bolt
+     * from the backend.
+     */
+    const M2_BOLT_ENABLED = "M2_BOLT_ENABLED";
+
     const DEFAULT_SWITCH_VALUES = array(
         self::M2_SAMPLE_SWITCH_NAME =>  array(
           self::NAME_KEY            => self::M2_SAMPLE_SWITCH_NAME,
           self::VAL_KEY             => true,
           self::DEFAULT_VAL_KEY     => false,
           self::ROLLOUT_KEY         => 0
-      )
+        ),
+        self::M2_BOLT_ENABLED =>  array(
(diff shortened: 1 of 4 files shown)


Helper/FeatureSwitch/Definitions.php
```diff
@@ -38,12 +38,24 @@ class Definitions {
...
           self::NAME_KEY            => self::M2_SAMPLE_SWITCH_NAME,
           self::VAL_KEY             => true,
           self::DEFAULT_VAL_KEY     => false,
           self::ROLLOUT_KEY         => 0
-      )
+        ),
+        self::M2_BOLT_ENABLED =>  array(
+          self::NAME_KEY            => self::M2_BOLT_ENABLED,
+          self::VAL_KEY             => true,
+          self::DEFAULT_VAL_KEY     => false,
```
daisy1754: shouldn't it true by default?

roopakv: the roll out value above should be true, this is the turned off value.

Helper/FeatureSwitch/Definitions.php
```diff
@@ -38,12 +38,24 @@ class Definitions {
...
+          self::ROLLOUT_KEY         => 100
```
matt-thomason: Shouldn't this be 0 by default? Otherwise, we always override with feature switches the value of the enabled/disabled button in the magento admin settings, right?

roopakv: no we dont want this to be 0 by default. we want bolt to always be enabled and we can choose to disable when needed.

the reason is: 
- we have ~100 merchants already live. We dont want to have to go add voerrides for each of them
- When a merchant is added we can choose to slow roll out for them if needed, and then we reduce this,. but by default we ship fully.

LMK if that makes sense

matt-thomason: @roopakv and I synced offline - I had a misunderstanding on rollout% - I interpreted this as a % chance of checking the status of the feature flag, whereas it's actually the state of the flag itself. @roopakv will update the phpdoc to be more explicit.

roopakv:
going to send a separate PR for the phpdoc update.

PR #511, opened 2020-01-13
Add support for custom checkboxes

vitaliyreznikov:
Fixes: https://app.asana.com/0/941920570700290/1155656238889067/f

\#changelog [Beta] Add support for custom checkboxes.

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/CheckboxesHandler.php
+++ b/Helper/CheckboxesHandler.php
@@ -19,7 +19,7 @@ namespace Bolt\Boltpay\Helper;
 
 use Magento\Framework\App\Helper\AbstractHelper;
 use Magento\Framework\App\Helper\Context;
-use Bolt\Boltpay\Helper\Order as OrderHelper;
+use Magento\Newsletter\Model\SubscriberFactory;
 
 /**
  * Boltpay web hook helper
@@ -29,40 +29,79 @@ use Bolt\Boltpay\Helper\Order as OrderHelper;
  */
 class CheckboxesHandler extends AbstractHelper
 {
+    const CATEGORY_NEWSLETTER = 'NEWSLETTER';
+    const COMMENT_PREFIX_TEXT = 'BOLTPAY INFO :: checkboxes';
+
     /**
-     * @var OrderHelper
+     * @var SubscriberFactory
      */
-    private $orderHelper;
+    private $subscriberFactory;
+
+    /**
+     * @var Bugsnag
+     */
+    private $bugsnag;
 
     /**
      * @param Context $context
      * @param OrderHelper $orderHelper
+     * @param Bugsnag $bugsnag
      */
     public function __construct(
         Context $context,
-        OrderHelper $orderHelper
+        Bugsnag $bugsnag,
+        SubscriberFactory $subscriberFactory
     )
     {
         parent::__construct($context);
-        $this->orderHelper = $orderHelper;
+        $this->subscriberFactory = $subscriberFactory;
+        $this->bugsnag = $bugsnag;
     }
 
-    public function handle($displayId, $checkboxes)
+    /**
+     * Handle checkboxes
+     *
+     * @param OrderModel $order
+     * @param array $checkboxes
+     */
+    public function handle($order, $checkboxes)
     {
         $comment = '';
         $needSubscribe = false;
         foreach ($checkboxes as $checkbox) {
-            if ($checkbox['category'] == 'NEWSLETTER' && $checkbox['value']) {
+            if ($checkbox['category'] == self::CATEGORY_NEWSLETTER && $checkbox['value']) {
                 $needSubscribe = true;
             }
             $comment .= '<br>' . $checkbox['text'] . ': ' . ($checkbox['value'] ? 'Yes' : 'No');
         }
         if ($comment) {
-            $commentPrefix = 'BOLTPAY INFO :: checkboxes';
-            $this->orderHelper->addCommentToStatusHistoryIfNotExists($displayId, $commentPrefix.$comment, $commentPrefix);
+            $order->addCommentToStatusHistory(self::COMMENT_PREFIX_TEXT.$comment);
+            $order->save();
         }
         if ($needSubscribe) {
-            $this->orderHelper->subscribeForNewsletter($displayId);
+            $this->subscribeForNewsletter($order);
+        }
+    }
+
+    /**
+     * Subscribe for newsletters
+     * - If order for logged in user subscribe by userId
+     * - If order for guest user subscribe by email
+     *
+     * @param OrderModel $order
+     */
+    public function subscribeForNewsletter($order) {
+        $customerId = $order->getCustomerId();
+        try {
+            if ($customerId) {
+                $this->subscriberFactory->create()->subscribeCustomerById($customerId);
+            } else {
+                $email = $order->getBillingAddress()->getEmail();
+                $this->subscriberFactory->create()->subscribe($email);
+            }
+        } catch (\Exception $e) {
+            // We are here if we are unable to send confirmation email, for example don't have transport
+            $this->bugsnag->notifyException($e);
         }
     }
 }
--- a/Helper/CheckboxesHandler.php
+++ b/Helper/CheckboxesHandler.php
@@ -31,6 +31,7 @@ class CheckboxesHandler extends AbstractHelper
 {
     const CATEGORY_NEWSLETTER = 'NEWSLETTER';
     const COMMENT_PREFIX_TEXT = 'BOLTPAY INFO :: checkboxes';
+    const FEATURE_SUBSCRIBE_TO_PLATFORM_NEWSLETTER = 'subscribe_to_platform_newsletter';
 
     /**
      * @var SubscriberFactory
@@ -69,7 +70,8 @@ class CheckboxesHandler extends AbstractHelper
         $comment = '';
         $needSubscribe = false;
         foreach ($checkboxes as $checkbox) {
-            if ($checkbox['category'] == self::CATEGORY_NEWSLETTER && $checkbox['value']) {
+            if ($checkbox['category'] == self::CATEGORY_NEWSLETTER && $checkbox['value']
+                && $checkbox['features'] && in_array(self::FEATURE_SUBSCRIBE_TO_PLATFORM_NEWSLETTER, $checkbox['features'])){
                 $needSubscribe = true;
             }
             $comment .= '<br>' . $checkbox['text'] . ': ' . ($checkbox['value'] ? 'Yes' : 'No');
@@ -79,7 +81,7 @@ class CheckboxesHandler extends AbstractHelper
             $order->save();
         }
         if ($needSubscribe) {
-            $this->subscribeForNewsletter($order);
+            $this->subscribeToNewsletter($order);
         }
     }
 
@@ -90,7 +92,7 @@ class CheckboxesHandler extends AbstractHelper
      *
      * @param OrderModel $order
      */
-    public function subscribeForNewsletter($order) {
+    public function subscribeToNewsletter($order) {
         $customerId = $order->getCustomerId();
         try {
             if ($customerId) {
--- a/Helper/CheckboxesHandler.php
+++ b/Helper/CheckboxesHandler.php
@@ -1,0 +1,68 @@
+<?php
+/**
+ * Bolt magento2 plugin
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ *
+ * @category   Bolt
+ * @package    Bolt_Boltpay
+ * @copyright  Copyright (c) 2018 Bolt Financial, Inc (https://www.bolt.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+namespace Bolt\Boltpay\Helper;
+
+use Magento\Framework\App\Helper\AbstractHelper;
+use Magento\Framework\App\Helper\Context;
+use Bolt\Boltpay\Helper\Order as OrderHelper;
+
+/**
+ * Boltpay web hook helper
+ *
+ * @SuppressWarnings(PHPMD.TooManyFields)
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
+ */
+class CheckboxesHandler extends AbstractHelper
+{
+    /**
+     * @var OrderHelper
+     */
+    private $orderHelper;
+
+    /**
+     * @param Context $context
+     * @param OrderHelper $orderHelper
+     */
+    public function __construct(
+        Context $context,
+        OrderHelper $orderHelper
+    )
+    {
+        parent::__construct($context);
+        $this->orderHelper = $orderHelper;
+    }
+
+    public function handle($displayId, $checkboxes)
+    {
+        $comment = '';
+        $needSubscribe = false;
+        foreach ($checkboxes as $checkbox) {
+            if ($checkbox['category'] == 'NEWSLETTER' && $checkbox['value']) {
+                $needSubscribe = true;
+            }
+            $comment .= '<br>' . $checkbox['text'] . ': ' . ($checkbox['value'] ? 'Yes' : 'No');
+        }
+        if ($comment) {
+            $commentPrefix = 'BOLTPAY INFO :: checkboxes';
+            $this->orderHelper->addCommentToStatusHistoryIfNotExists($displayId, $commentPrefix.$comment, $commentPrefix);
+        }
+        if ($needSubscribe) {
+            $this->orderHelper->subscribeForNewsletter($displayId);
+        }
+    }
+}
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -1984,12 +1984,16 @@ class Order extends AbstractHelper
             return;
         }
         $customerId = $order->getCustomerId();
-        if ($customerId) {
-            $this->subscriberFactory->create()->subscribeCustomerById($customerId);
-        } else {
-            $email = $order->getBillingAddress()->getEmail();
-            $this->subscriberFactory->create()->subscribe($email);
-
+        try {
+            if ($customerId) {
+                $this->subscriberFactory->create()->subscribeCustomerById($customerId);
+            } else {
+                $email = $order->getBillingAddress()->getEmail();
+                $this->subscriberFactory->create()->subscribe($email);
+            }
+        } catch (\Exception $e) {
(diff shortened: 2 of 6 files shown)


Model/Api/OrderManagement.php
```diff
@@ -189,6 +189,33 @@ public function manage(
         }
     }
 
+    /**
+     * Handle custom checkboxes if request has them
+     *
+     * @param string $displayId
+     */
+    private function handleCheckboxes($displayId) {
```
danac-gs: For extensibility / customization purpose we should at lest make this method public so we can write plugins (M2 interceptors) for it on the fly. This can be some sort of a dispatcher to methods that process specific checkbox categories. We should do the processing of every checkbox category out of this method. Maybe we should think of some naming convention so they can be automatically called if exist and we put them in a separate, extendable file.
@matt-thomason @njfaries @YevhenShBolt @johnmccombs1 thoughts?

matt-thomason: Great idea @danac-gs. I don't have a preference for any naming convention - open to suggestions.

Helper/Order.php
```diff
@@ -1850,4 +1860,52 @@ public function getOrderStoreIdByDisplayId($displayId)
...
+        if (!$order) {
+            return;
+        }
+        $customerId = $order->getCustomerId();
+        if ($customerId) {
+            $this->subscriberFactory->create()->subscribeCustomerById($customerId);
+        } else {
+            $email = $order->getBillingAddress()->getEmail();
+            $this->subscriberFactory->create()->subscribe($email);
+
```
YevhenShBolt: Extra row, please remove.

Model/Api/OrderManagement.php
YevhenShBolt: Also agree that method must be a public.

Model/Api/OrderManagement.php
```diff
@@ -189,6 +189,33 @@ public function manage(
...
+        $needSubscribe = false;
+        foreach ($request['checkboxes'] as $checkbox) {
+            if ($checkbox['category'] == 'NEWSLETTER') {
+                $needSubscribe = true;
+            }
+            $comment .= '<br>'.$checkbox['text'].': '.($checkbox['value'] ? 'Yes' : 'No');
+        }
+        if ($comment) {
+            $commentPrefix = 'BOLTPAY INFO :: checkboxes';
+            $this->orderHelper->addCommentToStatusHistoryIfNotExists($displayId, $commentPrefix.$comment, $commentPrefix);
```
var-bolt: I was looking at https://devdocs.magento.com/redoc/2.3/admin-rest-api.html#operation/salesOrderRepositoryV1SavePut. It looks like there is a field called `customer_note` what is the difference between that and order comments?

vitaliyreznikov: **customer_note** is a field for information that customers directly provide for us like "leave this on my porch". Opposite **order comments** for any information we want to save with order.
By default, only merchant staff see the order comment, but we can change it for the specific comment. **customer_note** is available for both staff and customers.
Am I correct @danac-gs

Model/Api/OrderManagement.php
vitaliyreznikov: Changed to public. @danac-gs any ideas about naming convention?

Helper/Order.php
vitaliyreznikov: Done

Helper/Order.php
```diff
@@ -1850,4 +1860,52 @@ public function getOrderStoreIdByDisplayId($displayId)
...
+        $order = $this->getExistingOrder($incrementId);
+        if (!$order) {
+            return;
+        }
+        $customerId = $order->getCustomerId();
+        if ($customerId) {
+            $this->subscriberFactory->create()->subscribeCustomerById($customerId);
+        } else {
+            $email = $order->getBillingAddress()->getEmail();
+            $this->subscriberFactory->create()->subscribe($email);
```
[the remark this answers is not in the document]
vitaliyreznikov: Good catch. Done.

danac-gs:
@vitaliyreznikov please fix the conflicts

Model/Api/OrderManagement.php
danac-gs: Actually, it seems like, although a bit confusing, that customer_note is currently only used when creating back office orders and the entered data is not visible to the customer, only to the merchant in the store backend. The preferred way of adding order comments is via addStatusHistory to sales_order_status_history table, the way we do it.

Model/Api/OrderManagement.php
danac-gs: handleCheckbox<Checkboxname>,  handleCheckboxNewsletter for instance.
And put the handlers in a separate class, eg. Helper\CheckboxHandlers.
That way we can override / extend that file in a separate, extension package (uploaded to app/code/Bolt/Bolt) when a customization is required (custom checkbox) without affecting the code plugin code.

Model/Api/OrderManagement.php
YevhenShBolt: i  agree with @danac-gs

Model/Api/OrderManagement.php
```diff
@@ -205,34 +232,39 @@ private function saveUpdateOrder($reference, $type, $display_id, $storeId)
...
-            $this->orderHelper->saveUpdateOrder(
-                $reference,
-                $storeId,
-                $this->request->getHeader(ConfigHelper::BOLT_TRACE_ID_HEADER),
-                $type
-            );
+            return;
+        }
+
+        $this->handleCheckboxes($display_id);
```
danac-gs: Perhaps we should handle checkboxes somewhere else, not in every hook. This way we can optimize things a bit by eliminating addCommentToStatusHistoryIfNotExists method and it's logic (that I am not very fond of).  \Bolt\Boltpay\Helper\Order::dispatchPostCheckoutEvents seems like a good candidate, it is run only once after order is approved.
@YevhenShBolt @matt-thomason @njfaries

vitaliyreznikov: @var-bolt I like @danac-gs proposal, but actually it means that we will look into checkboxes only in first, 'pending' hook.
If for some reason 'pending' hook doesn't contain checkboxes, but the next hook ('completed' ) contains we don't handle checkboxes.
Do you think it's a real issue or not?

Model/Api/OrderManagement.php
vitaliyreznikov: @danac-gs, For now, it seems logical to handle all checkboxes in one method, because we create only one comment with info of all checkboxes.
My proposal is just moving most part of handleCheckboxes into a separate helper class, like **handleCheckboxes($checkboxes)**.
If down the road we see that it's not enough for real integration tasks, we can improve it.

Model/Api/OrderManagement.php
var-bolt: I think it should be okay! But I will check with folks and get back on this comment

var-bolt:
@vitaliyreznikov Code as such looks good! But I am not familiar with the internals for Magento! So if the others ^ think it is good, it's a green light from my side

Model/Api/OrderManagement.php
danac-gs: Well, we can start with that. Then override that method or the whole helper class in a custom extension when needed.
What I was thinking of is a thin method that just loops through checkboxes and calls a handler for every one of them, passing it (all) the data available - the checkbox record from transaction, quote and order. This way, when dealing with custom checkboxes, we don't change the main handler, just add the handler for specific checkbox.
Adding order history comments regarding checkboxes can / should stay in the parent handler.
If we expect this custom checkboxes to be really custom then we should decouple their implementation handlers. If not, if subscribeForNewsletter would be the only one, then I got it all wrong (confused by the word custom) and we can leave it as is.

vitaliyreznikov: Yeah, I got what you offered, it's a good solution, but maybe it's better to wait for real cases. So just moving handleCheckboxes to separate helper for now.

danac-gs:
@vitaliyreznikov more conflicts

Helper/Order.php
```diff
@@ -1873,4 +1883,56 @@ public function getOrderStoreIdByDisplayId($displayId)
...
+
+    /**
+     * Subscribe for newsletters
+     * - Find order by displayId
+     * - If order for logged in user subscribe by userId
+     * - If order for guest user subscribe by email
+     *
+     * @param $displayId
+     */
+    public function subscribeForNewsletter($displayId) {
```
vitaliyreznikov: @danac-gs What do you think about this method? We can move it into new checkboxes helper, or leave it in this helper, in case we need to subscribe in case not related to checkboxes (very hypothetically).
What do you prefer?

danac-gs: @vitaliyreznikov I'd leave it here or more likely create a new helper class and put it there, \Bolt\Boltpay\Helper\Customer.
And change its parameter(s) from ($displayId) to either ($order) or even better ($customerId = null, $email = null) so we can (hypothetically) use it independently on the context it's been used.

danac-gs: @vitaliyreznikov after taking another look I think it's OK to leave this method here but pass it the order (get it from display_id before entering the method). We probably won't have another context of using this than the order.

Model/Api/OrderManagement.php
danac-gs: @vitaliyreznikov @matt-thomason @YevhenShBolt @njfaries 
 dispatchPostCheckoutEvents() is called at two places, in Controller\ReceivedUrlTrait::execute and Helper\Order::saveUpdateOrder and the full transaction info is not available in former one, so it is not a good canditate, you're right. There's this place just before Helper\Order::saveUpdateOrder:
```php
if ($order->getState() === OrderModel::STATE_PENDING_PAYMENT) {
    $this->resetOrderState($order);
}
```
This condition passes only once when order is approved and status changes from our pending_payment to magento default initial state - new / pending. We have all the data here ($transaction, $order) so we can pass it along to checkboxes handler. The handler gets all the data it needs, $transaction->checkboxes and $order, should just loop through checkboxes and either call specific handler for everyone of them (my proposal) or do it all in place (your solution). Moreover, $order->addStatusHistoryComment can be called there to record checkboxes, without the need of addCommentToStatusHistoryIfNotExists logic. As it is called one time only in Helper\Order::resetOrderState. Please consider this, it is much simpler than proposed solution. In a nutshell:
1. find a place in transaction lifetime (hooks) that executes only once after the order is approved (both bolt transaction data and magento order should be available here)
2. inject your method call there. It should be defined in an external class, Helper/CheckboxesHandler you already created.

vitaliyreznikov: Yeah, I didn't realize that I can find checkboxes data as `$transaction->checkboxes` Now I got it, thank you for detail explanation.

vitaliyreznikov: @danac-gs This is an example of `$transaction->checkboxes`
```
Array (    [0] => stdClass Object  (
            [external_id] => test_checkbox
	    [label] => test checkox
            [default_value] => 1
            [position] => 0
        )
)
```

It doesn't contain field **category** (like NEWSLETTER) so we should use data from webhook and we don't have it in **Helper\Order::saveUpdateOrder** right now. What do you think?

danac-gs: @vitaliyreznikov chances are external_id == category. If not this should be raised and probably fixed on another level, BX / checkout / core team, not sure. In my opinion, and from my experience all the relevant (checkboxes and other) data should be held in fetch transaction response. Lets at least try to fix the design before complicating the code on plugin level.
cc @var-bolt @matt-thomason @njfaries @ethanwayda22

danac-gs: @daisy1754 @nirajjayantbolt thoughts ^^^

vitaliyreznikov: Let's wait for more opinions, but from my point of view, although we found a sort of issue that should be fixed on the server side, it's still better to use primary information (webhook data) than the second one (data from the transaction).
It seems like a more scalable approach: if we need down the road to add additional checkboxes type or something else it's more simple doing webhook data.

If we want to avoid checking if comment exists we can pass all webhook data into `Helper\Order::saveUpdateOrder` It's not a very big deal and can be helpful in other situation in the future.

matt-thomason: @danac-gs @vitaliyreznikov @daisy1754 From the tech spec, it seems we were operating under the assumption that category would be explicitly provided. Here's a sample from the M2 spec:
```
"checkboxes": [
    {
      "checkbox_id": "subscription",
      "text": "Sign me up for the newsletter!",
      "value": true,
      "category": "newsletter",
      "features": [
        "subscribe_to_platform_newsletter"
      ]
    }
  ]
```

@vitaliyreznikov I also agree with you that relying on the webhook data is a stronger solution than the transaction data - Bolt should ideally be the source of truth since checkbox configs are coming from Bolt anyways, and it allows us more flexibility in the future.

@daisy1754 Can you clarify re: the category?

danac-gs: @matt-thomason @vitaliyreznikov @daisy1754 
So far the source of truth has always been the transaction data, ie. we always fetch the transaction info when get the hook because hook has not all the info needed. At this point we have already fetched the transaction. The question is, since there's a checkboxes section in transaction why it differs from the hook, is that intentional (why?) or an (easy to fix?) oversight. Anyway, I agree with this approach
> If we want to avoid checking if comment exists we can pass all webhook data into Helper\Order::saveUpdateOrder It's not a very big deal and can be helpful in other situation in the future.

Better pas one more variable and insert just one external call at the appropriate place then change / complicate already complex logic. In general, we should try to avoid altering the things that work / doing it as little as possible, unless the change is driven by code simplification or refactoring. Whenever we are about to add more complexity we should look back and check if there's another way. At least that's how I see it.

daisy1754: @var-bolt thought on why GET transaction response and payload in webhook is different?

daisy1754: re: wether we should rely on webhook or get transaction, I don't have a strong preference but I agree that I prefer to stick to one data source. I think once we complete refactoring of webhook [1] we can fully switch to webhook but I feel like that would be medium size project 
 [1] https://docs.google.com/document/d/159lDLJHjBPOCqLqPdLbabriS-OXEOX9lTIFcbsTOulY/edit?ts=5e33437c#

daisy1754: okay so couple of updates from convo in slack + me reading hail code:
1) difference between webhook and GET transaction is intended (I am personally not sure if I agree with this, but it's a longer discussion)
2) we'll refactor webhook payload and webhook will contain transaction view but that project is a few month away from productionize
3) external ID != category. if we want to access category we'd have to rely on webhook now

Given all + product's desire to include this in upcoming M2 release, I support relying on custom payload in hook here. we can re-discuss when (2) is ready to integrate.

@matt-thomason @danac-gs @vitaliyreznikov

var-bolt: @daisy1754 <3

Let's get this merged! Should we do 1 (or) 2 we can revisit this at a later point in time

vitaliyreznikov: @danac-gs about your comment in this thread from Feb 7th
```
if ($order->getState() === OrderModel::STATE_PENDING_PAYMENT) {
    $this->resetOrderState($order);
}
```
I've checked and at least on my test env. `$order->getState()` is `OrderModel::STATE_NEW` when we handle `pending ` hook and `OrderModel::STATE_PROCESSING` when we handle `payment `hook. So code inside if is never run.

Could you please check if we have any issues here?

danac-gs: @vitaliyreznikov the order is created in state / status pending_payment / pending_payment. \Bolt\Boltpay\Helper\Order::resetOrderState is the place where it's moved into new / pending.
However I just realized in pre-auth flow resetOrderState is called earlier, in success redirect call, \Bolt\Boltpay\Controller\ReceivedUrlTrait::execute. So the call from hook is skipped. Chances are that call is not needed and can be removed from there. I will take a look into it.

danac-gs: @vitaliyreznikov based on a quick test we can remove it from ReceivedUrlTrait, and rely on it being called in hooks. Will run more tests, but you can check as well if you have the time,

vitaliyreznikov: @danac-gs I also think your proposal should work, but if we implement it and remove **resetOrderState** from **ReceivedUrlTrait** it means that we set "new / pending" later that now and order will be sent to ERP system later. Do we need this change?
Anyway, this refactoring looks too big to be included in this PR.

If our goal is to handle checkboxes only once and do it in `Helper\Order::saveUpdateOrder` we can rely on bolt hook type (parameter $hookType) and handle checkboxes only if $hookType=='pending'.
And we can refactor it later, in separate PR. What do you think?

danac-gs: @vitaliyreznikov calling **resetOrderState** from hook is OK. It already works that way for non pre-auth orders. On the other hand, pending hook can potentially fail and we end up unnecessary doing this in loop. I'll be running tests over the weekend with resetOrderState removed from ReceivedUrlTrait and if all goes well create a separate pulls request. Early Monday morning our time at latest.

vitaliyreznikov: @danac-gs Ok, sounds good.

danac-gs: @vitaliyreznikov I tested this ^^ and it works. Will create a PR for it soon.

danac-gs: @vitaliyreznikov https://github.com/BoltApp/bolt-magento2/pull/554

Helper/CheckboxesHandler.php
```diff
@@ -0,0 +1,107 @@
...
+     *
+     * @param OrderModel $order
+     * @param array $checkboxes
+     */
+    public function handle($order, $checkboxes)
+    {
+        $comment = '';
+        $needSubscribe = false;
+        foreach ($checkboxes as $checkbox) {
+            if ($checkbox['category'] == self::CATEGORY_NEWSLETTER && $checkbox['value']) {
```
var-bolt:
```
  "checkboxes": [
    {
      "checkbox_id": "subscription",
      "text": "Sign me up for the newsletter!",
      "value": true,
      "category": "newsletter",
      "features": [
        "subscribe_to_platform_newsletter"
      ]
    }
  ]
```

To set `needSubscribe` to `true` we should also ensure that the `subscribe_to_platform_newsletter` feature is present in the features field.

Something like `$checkbox['features'] && in_array("subscribe_to_platform_newsletter", $checkbox['features'])`


Apart from this change, everything else looks good to me!

danac-gs:
@vitaliyreznikov I updated PR to point to master. Please check the conflicts.There had been lots of changes. Most important, the section where you call `$this->checkboxesHandler->handle($order, $hookPayload['checkboxes']);` has been moved to a different method.

Helper/CheckboxesHandler.php
vitaliyreznikov: Done

PR #534, opened 2020-02-05
Explicit async promise for bolt config

matt-thomason:
# Description
Async function and call await while calling BoltCheckout.configure.

Fixes: https://app.asana.com/0/941920570700290/1150898042943667/f (parent task for details)

\#changelog

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -640,7 +640,7 @@ ob_start();
         // Create Bolt order and configure BoltCheckout
         /////////////////////////////////////////////////////
         var createRequest = false;
-        var createOrder = function () {
+        var createOrder = async function () {
 
             // skip if there is already an ongoing request
             if (createRequest) return;
@@ -694,7 +694,7 @@ ob_start();
             }
             params = params.join('&');
 
-            var BC = BoltCheckout.configure(new Promise(function (resolve, reject) {
+            let promisedCart = await new Promise(function (resolve, reject) {
 
                 createRequest = true;
 
@@ -760,7 +760,9 @@ ob_start();
                         createRequest = false;
                     })
 
-            }), hints, callbacks);
+            });
+
+            var BC = await BoltCheckout.configure(promisedCart, hints, callbacks);
         };
         /////////////////////////////////////////////////////
 
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -640,7 +640,7 @@ ob_start();
         // Create Bolt order and configure BoltCheckout
         /////////////////////////////////////////////////////
         var createRequest = false;
-        var createOrder = async function () {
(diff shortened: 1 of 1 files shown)


view/frontend/templates/js/replacejs.phtml
```diff
@@ -760,7 +760,9 @@ ob_start();
                         createRequest = false;
                     })
 
-            }), hints, callbacks);
+            });
+
+            var BC = await BoltCheckout.configure(promisedCart, hints, callbacks);
```
roopakv: why var BC? it seems unusued.

view/frontend/templates/js/replacejs.phtml
```diff
@@ -640,7 +640,7 @@ ob_start();
         // Create Bolt order and configure BoltCheckout
         /////////////////////////////////////////////////////
         var createRequest = false;
-        var createOrder = function () {
+        var createOrder = async function () {
```
roopakv: do we have js async/await functionality in m2? i have never looked into it.

daisy1754:
I don't think async is supported in IE? https://caniuse.com/#feat=async-functions
you'd have to use good old `.then` and `.catch`. I believe Patric and Nathaniel will work on refactor this JS code which may include introducing transpiler, which would allow us to use modern feature like async

matt-thomason:
@daisy1754 Good catch on IE compatibility - I missed this.

view/frontend/templates/js/replacejs.phtml
matt-thomason: I'm not sure either. I carried it over from the old code - I didn't want to do a complete overhaul of what existed.

matt-thomason:
Synced with @daisy1754 - I missed that we have Promise because of polyfill from connect.js, but the same is not true for async. Rewrote to use .then

matt-thomason:
@roopakv There seems to be something wrong with this particular test, keeps running out of memory: `ci/circleci: unit-php72-magento23`

view/frontend/templates/js/replacejs.phtml
```diff
@@ -694,7 +694,7 @@ ob_start();
             }
             params = params.join('&');
 
-            var BC = BoltCheckout.configure(new Promise(function (resolve, reject) {
+            let cartPromise = new Promise(function (resolve, reject) {
```
daisy1754: Hmm I'm not sure if I understand motivation here. IMO The original code was what we want - we pass promise to `BoltCheckout.configure` so Bolt can properly wait for Promise even though user clicks checkout button before data is ready.

After change `BoltCheckout.configure` won't be called after Promise is resolved, if user clicks too early they would see error because Bolt was never initialized

matt-thomason: @danac-gs raised the same concern in the Asana ticket - he proposed a fix and I'll let him tackle it. Closing this ticket.

PR #536, opened 2020-02-05
[WIP] prevent checkout from opening without cart - undefined cart error fix

danac-gs:
# Description
We have found this error to occur most frequently for Windows 10 users on a slower internet connection. It is due to clicking the Bolt button before the cart object properly loads.

Fixes: https://app.asana.com/0/941920570700290/1149599154876665/f

\#changelog

# Type of change

# How Has This Been Tested?

# Checklist:


matt-thomason:
@roopakv Looks like that circle test is failing for all cases. Do you know why it might be running out of memory?

roopakv:
@matt-thomason no idea on circle, ill spend some time looking

view/frontend/templates/js/replacejs.phtml
```diff
@@ -588,6 +614,8 @@ ob_start();
                     }
                 }
 
+                if (!cart.orderToken) return false;
```
danac-gs: @daisy1754 this line fixes the entitled issue. However, when testing on a slow network I had a couple of cases that connect.js did not load on time. It is loaded asynchronously. So I added a fix for that as well in the same PR cause I thought this was kind of urgent.

daisy1754: can you open issue in engbugs with repro steps? I think it's better to handle this in connect.js side

said above, it may take a while to change connect.js - if we want to get change around waiting BoltCheckout let's do it as a separate PR

regarding this line, as Mike mentioned in asana I think intended way of doing is you pass cart promise in BoltCheckout.configure so we can wait. can you remind me why we cannot do that? (might be better to reply in asana so Mike can see your answer as well)

danac-gs: @daisy1754 we do pass cart as a promise when calling BoltCheckout.configure initially and on every cart change - in createOrder function. However, there are cases when only hints  change and we don't call crating of a new cart. We collect hints prefill from the front end - country, state, zip on the shopping cart page (there's support for the complete address on the checkout page that's not actively used). When hints change we call BoltCheckout.configure(cart, hints, callbacks) with the current state of the cart. If cart hasn't been fetched yet (slow network) we get the undefined cart error. This was the problem. It's fixed now.
cc @matt-thomason

daisy1754: sorry for delay - if you just want to update hint and not want to mess with cart you can pass null as argument

daisy1754: or you can hold reference to cart promise and keep reusing it

danac-gs: Thaks @daisy1754 that could help a lot.

danac-gs: Passing null is the cause of the issue. One of a few. connect.js loads and button is placed on the page before the javascript executes (all happen independently, asynchronously). The solution to this, at least from plugin perspective, is to prevent the checkout to pop up (callbacks.check returns false) until the javascript loads and reconfigures it. Trade off is a potential small number of dead clicks. Fixed the hints update by reusing the cart promise, great tip by Kazuki.

view/frontend/templates/js/connect.phtml
```diff
@@ -24,24 +24,14 @@ if ($block->shouldDisableBoltCheckout()) return;
...
-    async
-    data-publishable-key="<?php echo $checkoutKey; ?>">
-</script>
 <script
     id="bolt-connect"
     type="text/javascript"
     src="<?php echo $connectJsUrl; ?>"
     data-shopping-cart-id="magento2"
-    async
+    defer
```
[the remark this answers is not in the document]
danac-gs: @daisy1754 This is to postpone the execution of connect.js to after the page is fully parsed (having our javascript in place), preventing button rendering too early, thus lowering the chances of having dead clicks. Especially on slow networks. Even this way there is no noticeable delay in showing the button - it's available way before every (async) parts of the page load. It's a micro optimization I added after acknowledged how seriously dead clicks are treated. However, this may not be needed since the connect.js is now moved to the end of body. Anyway, it does no harm I think.

view/frontend/layout/default.xml
```diff
@@ -20,12 +20,15 @@
 <page xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:View/Layout/etc/page_configuration.xsd">
     <body>
         <referenceBlock name="head.additional">
-            <block class="Bolt\Boltpay\Block\Js" name="boltjs"    template="Bolt_Boltpay::js/boltjs.phtml"/>
+            <block class="Bolt\Boltpay\Block\Js" name="bolttrack" template="Bolt_Boltpay::js/track.phtml"/>
```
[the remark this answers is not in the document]
danac-gs: Then I can join them again in one pthml file and put them at the end of body, where connect is now moved to.

YevhenShBolt:
FYI:  from the 2.3.2 or 2.3.3 version Magento have specific functionality regarding all js files.
magento_config

It works well as in production mode and in development. i think that is more neat to use a js widgets or insert code through js files and require.js

danac-gs:
Update: initially place button disabled and enable it when javascript loads. Removes dead clicks.

PR #540, opened 2020-02-08
Product page checkout: add special codes for inventory errors

vitaliyreznikov:
Fixes: https://app.asana.com/0/0/1156147394609521/f

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -56,6 +56,8 @@ use Magento\Quote\Api\CartManagementInterface;
 use Bolt\Boltpay\Helper\Hook as HookHelper;
 use Magento\Framework\Webapi\Exception as WebapiException;
 use Magento\Customer\Api\CustomerRepositoryInterface as CustomerRepository;
+use Bolt\Boltpay\Exception\BoltException;
+use Bolt\Boltpay\Model\ErrorResponse as BoltErrorResponse;
 
 /**
  * Boltpay Cart helper
@@ -1832,6 +1834,7 @@ class Cart extends AbstractHelper
         //add item to quote
         $item = $request['items'][0];
         $product = $this->productRepository->getbyId($item['reference']);
+
         $options = json_decode($item['options'],true);
         if (isset($options['storeId']) && $options['storeId']) {
             $quote->setStoreId($options['storeId']);
@@ -1841,7 +1844,28 @@ class Cart extends AbstractHelper
         $options['qty'] = $item['quantity'];
         $options = new \Magento\Framework\DataObject($options);
 
-        $quote->addProduct($product, $options);
+        try {
+            $quote->addProduct($product, $options);
+        } catch (\Exception $e) {
+            $error_message = $e->getMessage();
+            if ($error_message == 'Product that you are trying to add is not available.') {
+                throw new BoltException(
+                    __($error_message),
+                    null,
+                    BoltErrorResponse::ERR_PPC_OUT_OF_STOCK
+                );
+            } else {
+                throw new BoltException(
+                    __('The requested qty is not available'),
+                    null,
+                    BoltErrorResponse::ERR_PPC_INVALID_QUANTITY
+                );
+            }
+        };
+        $storeId = @$item['options'];
+        if ($storeId) {
+            $quote->setStoreId($storeId);
+        }
 
         $quote->reserveOrderId();
 
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -1862,10 +1862,6 @@ class Cart extends AbstractHelper
                 );
             }
         };
-        $storeId = @$item['options'];
-        if ($storeId) {
-            $quote->setStoreId($storeId);
-        }
 
         $quote->reserveOrderId();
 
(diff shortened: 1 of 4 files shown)


Model/Api/OrderManagement.php
```diff
@@ -253,11 +260,24 @@ private function handleCartCreateApiCall()
...
+            $this->response->setHttpResponseCode(200);
+            $this->response->setBody(json_encode([
+                'status' => 'success',
+                'cart' => $cart,
+            ]));
+        } catch (\Exception $e) {
+            $error_message = $e->getMessage();
+            error_log($e->getCode());
+            error_log($e->getMessage());
+            if ($error_message == 'The requested qty is not available') {
```
danac-gs: I could not find this error message in Magento code base. The error handling in this file does not seem to be done right.

Model/Api/OrderManagement.php
```diff
@@ -253,11 +260,24 @@ private function handleCartCreateApiCall()
...
+            $cart = $this->cartHelper->createCartByRequest($request);
+            $this->response->setHttpResponseCode(200);
+            $this->response->setBody(json_encode([
+                'status' => 'success',
+                'cart' => $cart,
+            ]));
+        } catch (\Exception $e) {
+            $error_message = $e->getMessage();
+            error_log($e->getCode());
+            error_log($e->getMessage());
```
danac-gs: remove error_logs

Model/Api/OrderManagement.php
```diff
@@ -179,10 +179,17 @@ public function manage(
             $this->bugsnag->notifyException($e);
             $this->metricsClient->processMetric("webhooks.failure", 1, "webhooks.latency", $startTime);
             $this->response->setHttpResponseCode(422);
+            $error_code = $e->getCode();
+            if (($error_code != 6301 && $error_code != 6303)) {
```
danac-gs: We should have error codes defined as constants.
M1 reference https://github.com/BoltApp/bolt-magento1/blob/master/app/code/community/Bolt/Boltpay/Model/Productpage/Cart.php

Model/Api/OrderManagement.php
```diff
@@ -166,6 +167,14 @@ public function manage(
                 $this->saveUpdateOrder($reference, $type, $display_id, $storeId);
             }
             $this->metricsClient->processMetric("webhooks.success", 1, "webhooks.latency", $startTime);
+        } catch (BoltException $e) {
+            $this->bugsnag->notifyException($e);
+            $this->metricsClient->processMetric("webhooks.failure", 1, "webhooks.latency", $startTime);
+            $this->response->setHttpResponseCode(422);
+            $this->response->setBody(json_encode([
+                'status' => 'failure',
+                'error' => ['code' => $e->getCode(), 'message' => $e->getMessage()],
```
vitaliyreznikov: I've changed the error answer format.
With "old" format we have in DD:
image

With new one:

image

Helper/Cart.php
```diff
@@ -1833,8 +1835,24 @@ public function createCartByRequest($request)
         //add item to quote
         $item = $request['items'][0];
         $product = $this->productRepository->getbyId($item['reference']);
-        $quote->addProduct($product, $item['quantity']);
-
+        try {
+            $quote->addProduct($product, $item['quantity']);
+        } catch (\Exception $e) {
+            $error_message = $e->getMessage();
+            if ($error_message == 'Product that you are trying to add is not available.') {
```
danac-gs: We should not rely on message text but in this case I guess it's OK because there's a fallback on the second expression which is true in both cases. The first one seems more precise if the message matches. LGTM.

Model/Api/OrderManagement.php
```diff
@@ -166,6 +167,14 @@ public function manage(
                 $this->saveUpdateOrder($reference, $type, $display_id, $storeId);
             }
             $this->metricsClient->processMetric("webhooks.success", 1, "webhooks.latency", $startTime);
+        } catch (BoltException $e) {
+            $this->bugsnag->notifyException($e);
+            $this->metricsClient->processMetric("webhooks.failure", 1, "webhooks.latency", $startTime);
+            $this->response->setHttpResponseCode(422);
+            $this->response->setBody(json_encode([
+                'status' => 'failure',
```
danac-gs: Status is 'error' if other exceptions are thrown below, not 'failure'. Please check what's correct.

danac-gs:
@vitaliyreznikov please fix conflicts

Model/Api/OrderManagement.php
vitaliyreznikov: I am sure about "failure" for 'cart.create' because it works with bolt server, see screenshots above.

I don't know about the right error answer for webhooks, because bolt server accepts in the same way answers with "failure" and "error". Actually it seems like it doesn't parse answer.

Let's left "error" for webhooks right now. I've created asana to investigate it.
https://app.asana.com/0/941920570700290/1162655511300921

Model/Api/OrderManagement.php
vitaliyreznikov: Done

Model/Api/OrderManagement.php
vitaliyreznikov: Done

Helper/Cart.php
matt-thomason: I also really agree that this is a bad pattern. Preferable if we check the error code instead since that has less of a chance of changing over the message. I want this in 2.3.0 though and it's unlikely to break in one patch - I'll make a backlog task in Asana to strengthen this @vitaliyreznikov

PR #543, opened 2020-02-10
Don't empty cart after PPC checkout

vitaliyreznikov:
Fixes: https://app.asana.com/0/0/1155249819640874/f

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Plugin/ClearQuote.php
+++ b/Plugin/ClearQuote.php
@@ -50,7 +50,7 @@ class ClearQuote
 
     /**
      * @param CheckoutSession $subject
-     * @return CheckoutSession
+     * @return null Return null because method clearQuote have no arguments
      * @throws \Exception
      */
     public function beforeClearQuote(CheckoutSession $subject)
@@ -60,9 +60,9 @@ class ClearQuote
         $this->quote_to_restore = null;
         $current_quote_id = $subject->getQuote()->getId();
         $order_quote_id = $subject->getLastSuccessQuoteId();
-        if (!$current_quote_id || !$order_quote_id && $current_quote_id == $order_quote_id) {
+        if (!$current_quote_id || !$order_quote_id || $current_quote_id == $order_quote_id) {
             // In PPC checkout quote should be different then quote tied to order just created
-            return $subject;
+            return null;
         }
 
         // Although check above is enough, double check that we are in Bolt PPC process
@@ -70,12 +70,12 @@ class ClearQuote
         if (!$quote || $quote->getBoltParentQuoteId() != $order_quote_id) {
             // BoltParentQuoteId should be set (sign of Bolt)
             // and should be the same as quoteID (sign of PPC)
-            return $subject;
+            return null;
         }
 
         $this->quote_to_restore = $subject->getQuote();
 
-        return $subject;
+        return null;
     }
 
     /**
--- a/Plugin/ClearQuote.php
+++ b/Plugin/ClearQuote.php
@@ -36,7 +36,7 @@ class ClearQuote
     /**
      * @var \Magento\Quote\Model\Quote|null
      */
-    private $quote_to_restore;
+    private $quoteToRestore;
 
     /**
      * @param CartHelper $cartHelper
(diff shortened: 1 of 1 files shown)


Plugin/ClearQuote.php
```diff
@@ -25,14 +28,69 @@
...
+     * @throws \Exception
+     */
+    public function beforeClearQuote(CheckoutSession $subject)
+    {
+        // We don't want to clear quote
+        // in Product page checkout (PPC) flow
+        $this->quote_to_restore = null;
+        $current_quote_id = $subject->getQuote()->getId();
+        $order_quote_id = $subject->getLastSuccessQuoteId();
+        if (!$current_quote_id || !$order_quote_id && $current_quote_id == $order_quote_id) {
```
danac-gs: Not sure I understand this condition. Can you explain?

vitaliyreznikov: Uf, it's a type. Will fix.

Plugin/ClearQuote.php
```diff
@@ -25,14 +28,69 @@
...
+        $quote = $this->cartHelper->getQuoteById($order_quote_id);
+        if (!$quote || $quote->getBoltParentQuoteId() != $order_quote_id) {
+            // BoltParentQuoteId should be set (sign of Bolt)
+            // and should be the same as quoteID (sign of PPC)
+            return $subject;
+        }
+
+        $this->quote_to_restore = $subject->getQuote();
+
+        return $subject;
```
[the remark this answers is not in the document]
vitaliyreznikov: Thank you Daniel! It seems by documentation that we need to return "null" because we don't want to change arguments.
>  If the method does not change the argument for the observed method, it should return a null value.
What do you think?
https://devdocs.magento.com/guides/v2.3/extension-dev-guide/plugins.html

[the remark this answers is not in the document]
danac-gs: I agree.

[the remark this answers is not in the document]
vitaliyreznikov: Done

danac-gs:
@vitaliyreznikov the code looks good but I'd like to run a few test cases.

vitaliyreznikov:
@danac-gs Sorry I haven't mentioned this restriction.
Maybe we can leave it as it is for legacy (non pre-auth)?
What do you think?

danac-gs:
> @danac-gs Sorry I haven't mentioned this restriction.
> Maybe we can leave it as it is for legacy (non pre-auth)?
> What do you think?

I'm OK with that. ON the other hand, I'd like to see why it does not work.
Maybe we can do it in 2.4.0, if this is a must for 2.3.0 (which I don't believe it is).
cc @matt-thomason

YevhenShBolt:
@danac-gs , @matt-thomason @vitaliyreznikov 
I want to figure it out: Do we really need not to empty a shopping cart and redirect to the success page?
Each Magento website has several tracking systems that also track the success page.
I believe that each client should see the default success page with the order number, some additional product offers, banners, ads and etc.
Very often here is also a feedback form on the success page.

Plugin/ClearQuote.php
```diff
@@ -25,14 +28,69 @@
...
+     * @return null Return null because method clearQuote have no arguments
+     * @throws \Exception
+     */
+    public function beforeClearQuote(CheckoutSession $subject)
+    {
+        // We don't want to clear quote
+        // in Product page checkout (PPC) flow
+        $this->quote_to_restore = null;
+        $current_quote_id = $subject->getQuote()->getId();
+        $order_quote_id = $subject->getLastSuccessQuoteId();
```
YevhenShBolt: Can you use a proper variable naming?
https://devdocs.magento.com/guides/v2.3/coding-standards/code-standard-php.html
It is related to the
- `$this->quote_to_restore `
- `$current_quote_id`
- `$order_quote_id`

vitaliyreznikov:
@YevhenShBolt We want don't empty cart but do redirect on the success page. Do you think we can have any problem with a tracking system in this case?

YevhenShBolt:
@vitaliyreznikov  i think we can have some problems with tracking.
Also some websites on success page hide or remove mini-cart at all!

P.S.: i would be better to look at the following functionality:
Magento have `vendor/magento/module-instant-purchase/`, which implements an instant purchase for payments which can use this API (Braintre, Paypal and etc https://docs.magento.com/m2/ce/user_guide/sales/checkout-instant-purchase.html). 
If we are looking deeper, we can find that there is no redirect to the success page after placing an order, just success message for customer on the product page without redirect to the success page.

danac-gs:
> @danac-gs , @matt-thomason @vitaliyreznikov
> I want to figure it out: Do we really need not to empty a shopping cart and redirect to the success page?
> Each Magento website has several tracking systems that also track the success page.
> I believe that each client should see the default success page with the order number, some additional product offers, banners, ads and etc.
> Very often here is also a feedback form on the success page.

danac-gs:
> @danac-gs , @matt-thomason @vitaliyreznikov
> I want to figure it out: Do we really need not to empty a shopping cart and redirect to the success page?

We need to keep the shopping cart if it contains items. We need to redirect somewhere so our flow (pre-auth) works as is.

> Each Magento website has several tracking systems that also track the success page.

This quote is tracked theough Magento system all the way to the success page. Processing of it's success page should be no different that the "regular". It's the same page.

> I believe that each client should see the default success page with the order number, some additional product offers, banners, ads and etc.

This is exactly what happens here - the default success page.

> Very often here is also a feedback form on the success page.

Cool. So we redirect to the right page.

danac-gs:
> @vitaliyreznikov i think we can have some problems with tracking.

Maybe and maybe not. Let's deal with them when / if we face them. You're probably right, but it's hard to foresee every case and its impact. 

> Also some websites on success page hide or remove mini-cart at all!

When they move from the success page they would have their cart available.

> P.S.: i would be better to look at the following functionality:
> Magento have `vendor/magento/module-instant-purchase/`, which implements an instant purchase for payments which can use this API (Braintre, Paypal and etc https://docs.magento.com/m2/ce/user_guide/sales/checkout-instant-purchase.html).

This is OK but may require too much involvement and new stuff implemented. As well as moving from the default Magento implementation which works for logged in customers only with stored payment and uses vault.

> If we are looking deeper, we can find that there is no redirect to the success page after placing an order, just success message for customer on the product page without redirect to the success page.

We need the redirection for our flow to work. Pre-auth implementation (Bolt and Magento side) can not work without it.

vitaliyreznikov:
I agree with Daniel. @YevhenShBolt If you think we should find another approach for PPC please feel free to create new asana for discussing. It looks like it's better to not empty cart. The question like should we redirect to success page or not looks like more bigger that purpose of this PR.

Also, I created asana to handle non-pre-auth case https://app.asana.com/0/941920570700290/1162663341983417/f

This PR is ready for merge from my point of view.

Plugin/ClearQuote.php
vitaliyreznikov: Done, thank you

PR #567, opened 2020-02-21
Use parent quote id as a fallback to get the existing order

johnmccombs1:
# Description
This PR resolves the following case by using the parent quote id to get the existing order:

1. The customer clicks the Pay button in the Bolt modal
2. The create_order hook is sent to Magento and receives a time out error
3. There is a third party that updates the Magento increment id for the created order
4. Because of #2, there isn’t any successful response sent back to Bolt, so the Bolt increment id isn’t updated on Bolt
5. The pending hook is sent to Magento, it isn’t able to get the existing order by the Bolt increment id, so it tries to create the order and then it throws an exception
=> The expected behavior for #5 is that the pending hook is able to get the existing order to authorize and update the order.

Fixes: https://app.asana.com/0/564264490825835/1163120708856513

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -740,7 +740,7 @@ class Order extends AbstractHelper
         ///////////////////////////////////////////////////////////////
 
         // check if the order exists
-        $order = $this->getExistingOrder($incrementId);
+        $order = $this->getExistingOrder($incrementId, $parentQuoteId);
 
         // if not create the order
         if (!$order || !$order->getId()) {
@@ -1133,12 +1133,19 @@ class Order extends AbstractHelper
 
     /**
      * @param $orderIncrementId
-     * @return OrderModel|false
+     * @param null $parentQuoteId
+     * @return false|OrderInterface|OrderModel
      */
-    public function getExistingOrder($orderIncrementId)
+    public function getExistingOrder($orderIncrementId, $parentQuoteId = null)
     {
         /** @var OrderModel $order */
-        return $this->cartHelper->getOrderByIncrementId($orderIncrementId, true);
+        $order = $this->cartHelper->getOrderByIncrementId($orderIncrementId, true);
+
+        if ($parentQuoteId && (!$order || !$order->getId())) {
+            $order = $this->getOrderByQuoteId($parentQuoteId);
+        }
+
+        return $order;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


danac-gs:
> 1. The customer clicks the Pay button in the Bolt modal
> 2. The create_order hook is sent to Magento and receives a time out error
> 3. There is a third party that updates the Magento increment id for the created order
> 4. Because of #2, there isn’t any successful response sent back to Bolt, so the Bolt increment id isn’t updated on Bolt
> 5. The pending hook is sent to Magento, it isn’t able to get the existing order by the Bolt increment id, so it tries to create the order and then it throws an exception
>    => The expected behavior for #5 is that the pending hook is able to get the existing order to authorize and update the order.

2. I think this case should be taken care of differently at Bolt end. That is, never consider order was created on a timeout. Return an error and let customer try again. On the following, successful attempt, the existing order would be returned if previously created, or new order created and returned if not. Anything else could lead to issues.
3. In relation to 2., it is not clear how and why this may happen and what are all the possible cases and impacts. If the order was created by Bolt, the increment id was reserved in advance and passed on all the way to the order creation unchanged. Thus the line that updates the display id
```php
'display_id' => $createdOrder->getIncrementId() . ' / ' . $immutableQuoteId,
```
is redundant from plugin perspective, except that Bolt expects it and doesn't save the initial (from create order call) value to the transaction info / passes it along to hooks. Anyway, `$createdOrder->getIncrementId()` and `$quote->getReservedOrderId()` and the value passed initially in display_id should never differ. However, this can stay as it is for safeguarding reasons.
Moreover, if third party updates the Magento increment id for the created order, can that mean that the quote id is updated as well, ie. different quote used, or the same one (by id) but possibly changed in the meantime.
This changes can implicate that some other payment has been used for the order and that our order is a duplicate - should it be refunded and canceled / deleted?
4. As stated above, Bolt increment id should not be updated, it should be the same all the way.
5. We should at least have a payment type (Bolt or not) verification in place for this flow (for existing orders)

Shortly:
- Bolt side: on create order timeout do not consider order created, keep the display id if not updated
- Plugin side: add payment type verification when loading existing orders

cc @matt-thomason @nirajjayantbolt @daisy1754 @mccottry @vitaliyreznikov @YevhenShBolt

PR #571, opened 2020-03-02
Don't empty cart after successful PPC checkout (w/o preauth)

vitaliyreznikov:
We already did it for pre-auth, https://github.com/BoltApp/bolt-magento2/pull/543/, need to do the same for non-preauth flow.

Fixes: https://app.asana.com/0/941920570700290/1162663341983417/f

#changelog Don't empty cart after successful PPC checkout (w/o preauth)

# Type of change

# Checklist:

### DIFF
--- a/Controller/Order/Save.php
+++ b/Controller/Order/Save.php
@@ -103,13 +103,20 @@ class Save extends Action
     public function execute()
     {
         try {
+            $sessionQuote = $this->checkoutSession->getQuote();
             // get the transaction reference parameter
             $reference = $this->getRequest()->getParam('reference');
             // call order save and update
             list($quote, $order) = $this->orderHelper->saveUpdateOrder($reference);
-
-            // clear the session data
-            $this->replaceQuote($quote);
+            if ($sessionQuote && $sessionQuote->getId() != $quote->getId()) {
+                // If we had quote linked to session and it's not immutableQuote
+                // then we in product page, non-pre-auth checkout flow
+                // Session quote represents regular cart and we want to save it and restore after order creation
+                $this->replaceQuote($sessionQuote);
+            } else {
+                // clear the session data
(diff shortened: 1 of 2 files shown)


vitaliyreznikov:
@danac-gs I want to double-check if I thought about all cases.
https://github.com/BoltApp/bolt-magento2/blob/7cb86c96ad800cc60034d48f8136a8655eea7e92/Helper/Order.php#L791
In what cases this condition can be false?

danac-gs:
@vitaliyreznikov please check the conflicts

Controller/Order/Save.php
```diff
@@ -103,13 +103,20 @@ public function __construct(
     public function execute()
     {
         try {
+            $sessionQuote = $this->checkoutSession->getQuote();
             // get the transaction reference parameter
             $reference = $this->getRequest()->getParam('reference');
             // call order save and update
             list($quote, $order) = $this->orderHelper->saveUpdateOrder($reference);
```
vitaliyreznikov: @danac-gs Could you please help me when you have a chance?
Is it any chance that saveUpdateOrder method can return empty $quote?

I am asking because I see this check. We check if parentQuote is empty and few lines below we return it as $quote.
https://github.com/BoltApp/bolt-magento2/blob/7cb86c96ad800cc60034d48f8136a8655eea7e92/Helper/Order.php#L791

danac-gs: @vitaliyreznikov at some point all of the inactive quotes (our parent quote in this case) get deleted via cron job, by default after a month I think. So yes, an empty quote ("false" actually) can be returned from that call but not in your use case, calling order save right after the purchase.

vitaliyreznikov: Thank you, Daniel. So, what do you think about this PR? Just need fix/add tests?

PR #575, opened 2020-03-04
trigger a custom JS event when hints are set (button ready)

danac-gs:
# Description
Show the PPC button disabled and enable it when the required data (hints with encrypted customer id)  is loaded. Prevent dead clicks and errors. Do not hardcode the solution (eg. CSS selectors), make it work via Magento configuration.

Solution: 
1. trigger a custom JS event from the core plugin code when data is loaded (ppc-hints-set)
2. through Magento config add "ppc-disabled" CSS class and JavaScript that listens to the aforementioned event and removes "ppc-disabled" class.

Magento Bolt Pay configuration:

**Additional Checkout Button Class**:
```
ppc-disabled
```
**Global CSS**:
```css
.bolt-product-checkout-button.ppc-disabled {
    opacity: 0.5;
    cursor: default;
    pointer-events: none;
}
```
**Additional Javascript**:
```javascript
$(document).on('ppc-hints-set', function(){
    $('.bolt-product-checkout-button').removeClass('ppc-disabled');
});
```
Fixes: https://app.asana.com/0/941920570700290/1163069420923242/f

#changelog Load hints data via promise, trigger a custom JS event when PPC required data is available, 

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/view/frontend/templates/button_product_page.phtml
+++ b/view/frontend/templates/button_product_page.phtml
@@ -225,6 +225,7 @@ if (!$block->isSupportableType()) return;
                 }).always(function() {
                     setupProductPage();
                     $('#qty').on('change', setupProductPage);
+                    $(document).trigger('ppc-hints-set');
                 });
 
             // early button display, not active untill hints are received
--- a/view/frontend/templates/button_product_page.phtml
+++ b/view/frontend/templates/button_product_page.phtml
@@ -141,7 +141,6 @@ if (!$block->isSupportableType()) return;
                      * Magento customerData and authenticationPopup objects are
                      * used.
                      */
-
                     // check if login is required
                     if ( !customer().firstname && !isGuestCheckoutAllowed) {
                         // if authentication is required for checkout set a cookie
@@ -149,11 +148,6 @@ if (!$block->isSupportableType()) return;
                         authenticationPopup.showModal();
                         return false;
                     }
-
-                    if (!hints) {
-                        return false;
-                    }
-
                     // wait for validation module init
                     try {
                         $('#product_addtocart_form').validation('isValid');
@@ -164,7 +158,6 @@ if (!$block->isSupportableType()) return;
                     if ( ! $('#product_addtocart_form').validate().form()) {
                         return false;
                     }
-
                     return true;
                 },
 
@@ -181,6 +174,8 @@ if (!$block->isSupportableType()) return;
                 var quantity = Number($('#qty').val());
                 return quantity > 0 ? quantity : 1;
             };
+
+            var hints;
 
             var setupProductPage = function () {
                 var options = {};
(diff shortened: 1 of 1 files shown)


matt-thomason:
@danac-gs Does this mean every merchant who wants PPC needs to set this custom CSS and javascript? That seems really fragile. I know we are discouraging hard coding CSS and JS that manipulates the button, but this seems worse (making merchants write CSS and javascript that manipulates the button that we won't have good visibility on).

Did you talk with anyone from BX about how to slow-roll this button until hints are ready?
@daisy1754 @roopakv

danac-gs:
@matt-thomason there are 2 opposing things to deal with here:

1. button show latency - show button as early as possible
2. dead clicks - prevent clicks until the checkout configuration is valid

similar to how is's ben done for the default checkout

I am refactoring it now to put the HTTP request in a promise. Still I'd leave this event dispatch in the core plugin. On slow connections (and complaint from the merchant) this might be useful adding to configuration.

The Magento default buttons are also shown disabled until ready

Moreover, the script requires 3 magento modules (loaded asynchronously) and if they are not initialized on button click the check returns false (dead click) and the checkout does not open.
They are used to verify

1. if all the product options are selected
2. is guest checkout allowed (for guest session)

we need to know that before we can let the modal open

view/frontend/templates/button_product_page.phtml
```diff
@@ -207,31 +202,35 @@ if (!$block->isSupportableType()) return;
...
+                    $.get(settings.get_hints_url)
+                        .done(function(data) {
+                            hints = data.hints;
+                        })
+                        .fail(function() {
+                            hints = {prefill:{}};
+                        })
+                        .always(function() {
+                            resolve(hints);
+                            // disapatch event when required data is ready
```
vitaliyreznikov: typo in word dispatch

view/frontend/templates/button_product_page.phtml
```diff
@@ -207,31 +202,35 @@ if (!$block->isSupportableType()) return;
                         options: JSON.stringify(options)
                     }]
                 };
+
+                var hintsPromise = new Promise(function (resolve, reject) {
+                    if (hints) {
+                        resolve(hints);
+                        return;
+                    }
+                    $.get(settings.get_hints_url)
```
vitaliyreznikov: Do you think we can (in separate PR) find a way how to call the server only if a user logged in?

view/frontend/templates/button_product_page.phtml
```diff
@@ -207,31 +202,35 @@ if (!$block->isSupportableType()) return;
                         options: JSON.stringify(options)
                     }]
                 };
+
+                var hintsPromise = new Promise(function (resolve, reject) {
```
vitaliyreznikov: Maybe we have a little problem. If we hints ajax call works long enough (10 seconds), user can click product options, configureProductCheckout calls again, hints call again.
Bolt modal wait for last ajax call and can as result show timeout error.
Maybe it's better to create promise only once, not on each setupProductPage() call

matt-thomason: @vitaliyreznikov Can you create a backlog task in the M2 asana board to investigate the identified corner case?

view/frontend/templates/button_product_page.phtml
vitaliyreznikov: @danac-gs I'd like to create asana about it if you don't mind.

view/frontend/templates/button_product_page.phtml
vitaliyreznikov: @matt-thomason Sure. https://app.asana.com/0/941920570700290/1165280808501785/f

PR #594, opened 2020-03-17
Adds Minimum Order Amount support

dumega:
# Description
If we enable `Minimum Order Amount` in M2 (Admin dashboard -> Store -> Configuration -> Sales -> Sales -> Minimum Order Amount ) and set specific Minimum Amount, in the front shop, even the cart subtotal is less than the minimum amount, the customer can still open Bolt modal and process to payment, and eventually the customer will get an error for the restriction.

The `Minimum Order Amount` is a native setting in M2, so Bolt should support it. 

Then when the customer try to checkout with Bolt, we need to check the related restriction before opening Bolt modal

Fixes: https://app.asana.com/0/941920570700290/1165531437050969/f

#changelog Adds Minimum Order Amount support

# Type of change

# How Has This Been Tested?

# Checklist:


YevhenShBolt:
Yes, i worked on similar `Minimum amount` solution.
I did not add this solution to the core because it is not work for all merchants.

As result, the merchant decided to just add a custom popup with message before Bolt modal open.

dumega:
@danac-gs 

1. This PR already take the discounts into account, so when the customer try to apply the coupon via Bolt modal and the implement would fail if the cart amount become less than "Minimum Order Amount".

2. 

> The reason is taxes may or may not be included in minimum amount calculation. In case they are, we don't have them at this point so this may block potentially valid orders.

I call $quote->validateMinimumAmount() for validation, so in my opinion, the calculation would keep the same way as M2 core. I tested with "Include Tax to Amount"(Set to YES), the Bolt modal follow the same behavior as M2 checkout button.

So @danac-gs  could you show an example to block potentially valid orders? Thanks.

danac-gs:
@dumega what happens in this case:
1. minimum order amount is 50
2. tax should be taken into account
3. items subtotal is 49
4. at the time of Bolt Create Order call tax is still unknown, zero (order total is 49, less then allowed)
5. after entering shipping address tax should be 2 (order total 51, greater than minimum, order valid)
Do we get to step 5?

dumega:
@danac-gs 

I disable Bolt and follow your steps to check how the Magento process by default, and although the customer enter the address for shipping&tax estimation in the cart page and the tax&cart total get updated, the Magento checkout button is still disabled, so the customer only can process if the subtotal is larger than restriction amount.


image


image


image


So with PR, Bolt still follow what Magento does for the Minimum Order Amount, and does not block  valid order.

danac-gs:
@dumega I have replicated the example that you sent, as well as looked into Magento implementation and my conclusion is that there's a **bug** or inconsistency **in Magento** with this **that we should not repeat**. Let me try to explain with an example. First, please make sure **Include Tax to Amount** is set to **Yes** in **Minimum Order Amount** Magento configuration. Second, create a customer account and set its shipping address in a way that it should add taxes to orders. Try your example cart with the logged in customer. You'll see that the button is now enabled and the order is valid. What happens is, and I consider it a Magento bug or more likely a tech debt, for anonymous customer, although the shipping estimation is done, there's no address data yet assigned to the quote. There's an empty DB quote address record, not yet populated. So basically, in your example, second screenshot, Magento prevents potentially valid order for anonymous user. Minimum amount validation should be postponed until the address is determined, or done based on shipping estimation address. From our perspective, it is done in pre-auth order creation call, after shipping address and method are entered and tax calculated. I strongly disagree with this PR, it adds no value, actually it can do just the opposite. I see leaving it as is as our advantage, especially in some special cases with tight margins, eg. A/B testing against other methods. This is the second time this was introduced, previously by @YevhenShBolt some time ago. Pushed back.
cc @matt-thomason @njfaries @mccottry @johnmccombs1 @roopakv @daisy1754 @vitaliyreznikov @nirajjayantbolt

mccottry:
@danac-gs My thoughts is that we should match the behavior of the official Magento release.  Whenever it allows an order, we should.   When it does not, we should not.  I do believe we should see value in enforcing a minimum order value if a merchant has gone as far to take the step of enabling this config.

As @dumega states,
> The Minimum Order Amount is a native setting in M2, so Bolt should support it.

danac-gs:
> @danac-gs My thoughts is that we should match the behavior of the official Magento release. Whenever it allows an order, we should. When it does not, we should not. I do believe we should see value in enforcing a minimum order value if a merchant has gone as far to take the step of enabling this config.
> 
> As @dumega states,
> 
> > The Minimum Order Amount is a native setting in M2, so Bolt should support it.

@mccottry we do support minimum order amount already, in addition we bypass the Magento bug that blocks potentially valid orders. It's explained it the thread (non bold).

mccottry:
> @mccottry we do support minimum order amount already, in addition we bypass the Magento bug that blocks potentially valid orders. It's explained it the thread (non bold).

Ah, I see.  I missed the "**Include Tax to Amount** is set to **Yes**" Since pre-auth takes care of this check, and Magento will create the order if submitted, but on the front-end, will block the order, then I agree with @danac-gs, Bolt is adding more value as-is, the feature is already supported, and this PR only serves as a minor benefit at best by slightly improving the user experience.

njfaries:
I think reflecting M2 native behaviour in our plugin makes sense, but as @danac-gs pointed out there's not really a good way to do this at the plugin level without potentially causing problems for valid orders. If we want to implement this behaviour it would be more of a question of implementing it in the core Bolt app, not here. This also is something that becomes a question for the product team.

cc: @ellenatbolt @danielatbolt

dumega:
Well, I shall go to close this PR, after further investigation, as @danac-gs point out, there could be potential problems. Also I think the Bolt pre-auth already takes care of this check, so there is no further action for the core Bolt app now.

PR #600, opened 2020-03-18
Make GetFeatureSwitch call when merchant key is changed

vitaliyreznikov:
We should make GetFeatureSwitch call each time when plugin is installed / updated, but now we don't do it when plugin is installed: at this moment merchant key is empty and API call fails.
**Solution:**
Call GetFeatureSwitch  each time when merchant key is changed

Fixes: https://app.asana.com/0/941920570700290/1162717349359250/f

#changelog Make GetFeatureSwitch call when merchant key is changed

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Plugin/CheckSettingsUpdate.php
+++ b/Plugin/CheckSettingsUpdate.php
@@ -41,7 +41,7 @@ class CheckSettingsUpdate
     private $configHelper;
 
     /*
-     * @var Manager
+     * @var FeatureSwitchManager
      */
     protected $fsManager;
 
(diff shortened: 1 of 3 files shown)


Plugin/CheckSettingsUpdate.php
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * @param CheckoutSession $subject
+     * @return CheckoutSession
+     */
+    public function afterSave(MagentoConfig $subject, $result)
+    {
+        $apiKey = $this->configHelper->getApiKey();
+        if ($apiKey && $apiKey != $this->oldApiKey) {
```
danac-gs: This does not work if the key gets removed. Should it?

vitaliyreznikov: In new key is empty we shouldn't try to do API call: it's useless without API key, we just receive an error.

PR #609, opened 2020-03-21
[Order management] ENP-25 show bolt account button

vitaliyreznikov:
add <div class="bolt-track-order"></div> and include core script account.js to show bolt account button

Fixes: [ENP-25]

#changelog Show bolt account button (order management)

Project summary:
Frontend changes:
add &lt;div class="bolt-track-order"&gt;&lt;/div&gt;
include core script account.js
As a result "account" button is shown
2. When a user clicks on the "account" button, they see iFrame with a prompt to enter an email.
3. Bolt server calls the new plugin endpoint to know if there is magento2 user with this email.
If yes, a user is redirected to M2 login page
If no, a user fills in the phone number, enters SMS code, and sees a page with all orders from this store.
4. When merchant staff add track number into order plugin, send it to bolt so that a user can see it in account iFrame
See https://boltpay.atlassian.net/wiki/spaces/EN/pages/735019129/Shopper+Dashboard+-+Order+Management

# Type of change

# How Has This Been Tested?

# Checklist:

[ENP-25]: https://boltpay.atlassian.net/browse/ENP-25

### DIFF
--- a/Helper/Config.php
+++ b/Helper/Config.php
@@ -122,6 +122,11 @@ class Config extends AbstractHelper
      */
     const XML_PATH_PRODUCT_PAGE_CHECKOUT = 'payment/boltpay/product_page_checkout';
 
+    /**
+     * Enable Bolt order management
+     */
+    const XML_PATH_PRODUCT_ORDER_MANAGEMENT = 'payment/boltpay/order_management';
+
 
     /**
      * Prefetch shipping
@@ -670,6 +675,22 @@ class Config extends AbstractHelper
     {
         return $this->getScopeConfig()->isSetFlag(
             self::XML_PATH_PRODUCT_PAGE_CHECKOUT,
+            ScopeInterface::SCOPE_STORE,
+            $store
+        );
+    }
+
+    /**
+     * Get Order management flag from config
+     *
+     * @param int|string|Store $store
+     *
+     * @return  boolean
+     */
+    public function isOrderManagementEnabled($store = null)
+    {
+        return $this->getScopeConfig()->isSetFlag(
+            self::XML_PATH_PRODUCT_ORDER_MANAGEMENT,
             ScopeInterface::SCOPE_STORE,
             $store
         );
--- a/etc/config.xml
+++ b/etc/config.xml
@@ -31,6 +31,7 @@
                 <custom_cdn></custom_cdn>
                 <debug>0</debug>
                 <product_page_checkout>0</product_page_checkout>
+                <order_management>0</order_management>
                 <allowspecific>0</allowspecific>
                 <order_status>new</order_status>
                 <currency>USD</currency>
--- a/view/frontend/layout/default.xml
+++ b/view/frontend/layout/default.xml
@@ -24,6 +24,14 @@
             <block class="Bolt\Boltpay\Block\Js" name="boltcss"   template="Bolt_Boltpay::css/boltcss.phtml"/>
             <block class="Bolt\Boltpay\Block\Js" name="replacejs" template="Bolt_Boltpay::js/replacejs.phtml"/>
         </referenceBlock>
+        <!-- for blank theme-->
+        <referenceBlock name="top.links">
+            <block class="Bolt\Boltpay\Block\Js" name="boltaccount_blank" template="Bolt_Boltpay::boltaccount.phtml" before="-"/>
+        </referenceBlock>
+        <!-- for luna theme-->
+        <referenceBlock name="header.links">
+            <block class="Bolt\Boltpay\Block\Js" name="boltaccount_luna" template="Bolt_Boltpay::boltaccount.phtml" before="-"/>
+        </referenceBlock>
         <referenceBlock name="after.body.start">
             <block class="Bolt\Boltpay\Block\Js" name="bolt_popup" template="Bolt_Boltpay::popup.phtml" />
         </referenceBlock>
--- a/view/frontend/templates/boltaccount.phtml
+++ b/view/frontend/templates/boltaccount.phtml
@@ -1,0 +1,57 @@
+<?php
+/**
+ * Bolt magento2 plugin
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ *
+ * @category   Bolt
+ * @package    Bolt_Boltpay
+ * @copyright  Copyright (c) 2018 Bolt Financial, Inc (https://www.bolt.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+/**
+ * Bolt Account button
+ *
+ * @var $block \Bolt\Boltpay\Block\Js
+ */
+if ($block->shouldDisableBoltCheckout()) return;
+if (!$block->isOrderManagementEnabled()) return;
+if ($block->isBlockAlreadyShown('account')) return;
+$accountJsUrl   = $block->getAccountJsUrl();
+$checkoutKey  = $block->getCheckoutKey();
+?>
+<div class="bolt-account-login"></div>
+<script
+        id="bolt-account"
+        type="text/javascript"
+        src="<?php echo $accountJsUrl; ?>"
+        data-shopping-cart-id="magento2"
+        async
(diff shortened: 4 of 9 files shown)


Helper/Config.php
```diff
@@ -654,6 +659,22 @@ public function getProductPageCheckoutFlag($store = null)
...
     }
 
+    /**
+     * Get Order management flag from config
+     *
+     * @param int|string|Store $store
+     *
+     * @return  boolean
+     */
+    public function isOrderManagementEnabled($store = null)
```
vitaliyreznikov: If we enable new feature via plugin settings and disable it by default, should we add feature switches as well?
+: We can disable remotely if necessary
-: It can be confusing for merchant staff: if we disable feature remotely they couldn't enable it via plugin settings.

@roopakv @matt-thomason @akulkapoor ,

view/frontend/templates/boltaccount.phtml
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @var $block \Bolt\Boltpay\Block\Js
+ */
+if ($block->shouldDisableBoltCheckout()) return;
+if (!$block->isOrderManagementEnabled()) return;
+if ($block->isBlockAlreadyShown('account')) return;
+$accountJsUrl   = $block->getAccountJsUrl();
+$checkoutKey  = $block->getCheckoutKey();
+?>
+<div class="bolt-account-login"></div>
```
vitaliyreznikov: It works for desktop, but doesn't for mobile. Will discuss with BX team on Monday

danac-gs:
@vitaliyreznikov Please can you add more context to the PR description. Could find nothing explanatory in Asana or Jira.

Helper/Config.php
matt-thomason: I think we should make a feature switch. We'll only disable it remotely in emergency cases - something is wrong on our end and we want to disable it without going through the CSM or requiring a plugin hotfix first.
@roopakv

view/frontend/templates/boltaccount.phtml
matt-thomason: Any follow-up here from BX team?

Helper/Config.php
danac-gs: @matt-thomason @vitaliyreznikov since this is a new feature and probably not every merchant would want it enabled, especially the old ones, if there's no store side configuration then the feature switch should be OFF by default.

vitaliyreznikov: @danac-gs We have store side configuration (in this PR) and it is disabled by default.

vitaliyreznikov: @matt-thomason Added

view/frontend/layout/default.xml
```diff
@@ -24,6 +24,14 @@
...
             <block class="Bolt\Boltpay\Block\Js" name="replacejs" template="Bolt_Boltpay::js/replacejs.phtml"/>
         </referenceBlock>
+        <!-- for blank theme-->
+        <referenceBlock name="top.links">
+            <block class="Bolt\Boltpay\Block\Js" name="boltaccount_blank" template="Bolt_Boltpay::boltaccount.phtml" before="-"/>
+        </referenceBlock>
+        <!-- for luna theme-->
+        <referenceBlock name="header.links">
+            <block class="Bolt\Boltpay\Block\Js" name="boltaccount_luna" template="Bolt_Boltpay::boltaccount.phtml" before="-"/>
+        </referenceBlock>
```
danac-gs, line 34: Will this work every time, for every theme? Is there a better way to uniformly insert the button? Theme agnostic one.
Perhaps something similar to how we are injecting button in the minicart, JS + CSS selectors. It is not elegant, but it works. On the other hand, we may want to move over from that sort of implementation. Not sure. Just an idea to discuss. Even better, to do it Magento way, perhaps we can conditionally / programmatically set the referenceBlock and inject our block in it. cc @matt-thomason @njfaries @johnmccombs1

view/frontend/templates/boltaccount.phtml
vitaliyreznikov: I am sorry Matt, I didn't mention that your comment from this thread. I've figured out what wrong was with mobile, it was specifically M2 issue. I've fixed it, now it works for mobile as well.

view/frontend/layout/default.xml
danac-gs: @vitaliyreznikov After taking a second look I think you should probably leave it like this. You add it to both top.links and header.links, and one of the blocks should exist in (most of) all the themes. If it does not we could deal with it in the integration when we get there. If there is a case with both blocks present and two Bolt account buttons added we could hide one with CSS until we find a better solution.

view/frontend/templates/boltaccount.phtml
```diff
@@ -0,0 +1,96 @@
...
+
+    .header.panel > .header.links > li.customer-welcome > span > button {
+        margin-top: 14px;
+    }
+
+    .header.panel {
+        padding-top: 0px !important;
+        padding-bottom: 0px !important;
+    }
+</style>
```
danac-gs, line 96: We should probably have no hardcoded CSS. This can all go into GlobalCSS config field as addition to the default value.

view/frontend/layout/default.xml
vitaliyreznikov: Thank you Daniel, it's important thoughts. 

It's very logical for other themes developers save this block names "header.links" and "top.links". But of course, we can't be sure. This project is in early beta. I propose to leave code as it is and looks at how it works. If we handle with a theme that has another block name, we can just add it here.

Injection button via JS has two disadvantages:
- we need to add one more field into our settings, but we already have a lot
- we need to inject a button only after M2 reformat menu to the mobile view and we don't have any events to know when it happens. Probably we can find solution but right now I don't have it.

Long story short, now we support two themes and probably we are good with other themes. We can go ahead and maybe create a new task to investigate how we can handle with different themes.

view/frontend/templates/boltaccount.phtml
vitaliyreznikov: 1. We shouldn't apply these rules if feature is enabled.
2. If we add it into default value it doesn't affect to existing merchant

For (1) we can add an additional CSS class to .header.panel and add this class to rules. But what about (2).
What's your opinion @danac-gs

danac-gs: @vitaliyreznikov correct, the (default) configuration fields won't update on plugin update. Perhaps we should ask merchants to update configuration manually in such cases. @matt-thomason @njfaries

vitaliyreznikov: I propose to copy settings to GlobalCSS defaults but in this place check if GlobalCSS has any rules related to the feature and include CSS here if not. How it sounds @danac-gs?

danac-gs: @vitaliyreznikov Personally, I don't have nothing against it, but recently we had a request to remove all hardcoded CSS from the plugin code. Moreover, assuming _.header.panel_ dimensions are the same in every theme / device can be wrong.

view/frontend/layout/default.xml
matt-thomason: A task in M2P to investigate other themes is a good idea @vitaliyreznikov . Please title it is `[Spike]` at the beginning of the title.

view/frontend/templates/boltaccount.phtml
matt-thomason: @danac-gs is correct here, we should not have any hardcoded CSS in the plugin code, especially relating to bolt-created classes or items.
@vitaliyreznikov perhaps we can take a simpler approach, and include these items to copy into GlobalCSS as part of our onboarding documents, both public and private facing?

matt-thomason:
Also, please resolve conflicts @vitaliyreznikov

view/frontend/layout/default.xml
vitaliyreznikov: @matt-thomason Task is created https://boltpay.atlassian.net/browse/M2P-33

view/frontend/templates/boltaccount.phtml
vitaliyreznikov: Done. Created task about documentation - https://boltpay.atlassian.net/browse/M2P-34

etc/config.xml
```diff
@@ -31,6 +31,7 @@
                 <custom_cdn></custom_cdn>
                 <debug>0</debug>
                 <product_page_checkout>0</product_page_checkout>
+                <order_management>0</order_management>
```
danac-gs, line 34: @vitaliyreznikov it is safe to add the CSS below in _global_css_ section

PR #613, opened 2020-03-24
Fix Bug - undefined store id result in the failure of  backoffice order creation 

dumega:
# Description
The window object may be reset during backoffice creation for some reasons, as a result, the variable settings fail to keep in sync with window.boltConfig, so we use Proxy object to make sure settings always point to the updated window.boltConfig.

Fixes: https://app.asana.com/0/941920570700290/1167452929959226/f

#changelog Fix Bug - undefined store id result in the failure of  backoffice order creation 

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/view/adminhtml/templates/boltpay/js/replacejs.phtml
+++ b/view/adminhtml/templates/boltpay/js/replacejs.phtml
@@ -307,8 +307,18 @@ require([
         }
     };
 
+    var boltConfigHandler = {
+        get: function(target, property) {
+            return target['boltConfig'][property];
+        },
+        set: function(target, property, value, receiver) {
+            target['boltConfig'][property] = value;
+            return true;
+        },
+    };
+
     // The configuration parameters passed from the php block
-    var settings = window.boltConfig,
+    var settings = new Proxy(window, boltConfigHandler),
     // TODO: make `with-cards` option backend configurable
         bolt_button_css_class    = 'bolt-checkout-button with-cards',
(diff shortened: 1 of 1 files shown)


view/adminhtml/templates/boltpay/js/replacejs.phtml
```diff
@@ -307,8 +307,21 @@ require([
...
+        },
+        set: function(target, property, value, receiver) {
+            target['boltConfig'][property] = value;
+            return true;
+        },
+    };
+
     // The configuration parameters passed from the php block
-    var settings = window.boltConfig,
+    var settings = new Proxy(window, boltConfigHandler),
```
danac-gs: @dumega This is a great solution. Unfortunately, it does not work in IE.
You could just replace
```javascript
window.boltConfig.storeId = data.storeId;
window.boltConfig.is_pre_auth = data.isPreAuth;
```
with
```javascript
settings.storeId = data.storeId;
settings.is_pre_auth = data.isPreAuth;
```
in `createOrder` function below.

dumega: @danac-gs Good catch~!!! My idea is to force to keep the two object reference in sync, but since the Proxy does not support IE, I am afraid there is not simple way to do it.

Then I think we have to maintain the variable `settings`  locally, so just treat it as a copy of window.boltConfig, and apply the fix you suggest.

PR #619, opened 2020-03-27
Remove the outdated rounding code

johnmccombs1:
# Description
This code is causing additional rounding problems for a merchant 

Fixes: 
https://app.asana.com/0/1125081140214268/1168686308499785
https://app.asana.com/0/1125081140214268/1164552839717137

#changelog Remove the outdated rounding code

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -1386,11 +1386,6 @@ class Cart extends AbstractHelper
             $taxAmount = 0;
         }
 
-        // include potential rounding difference and reset $diff accumulator
-        $cart['items'][0]['total_amount'] += round($diff);
-        $totalAmount += round($diff);
-        $diff = 0;
-
         // add discount data
         list ($discounts, $totalAmount, $diff) = $this->collectDiscounts (
             $totalAmount, $diff, $paymentOnly


Helper/Cart.php
```diff
@@ -1386,9 +1386,6 @@ public function getCartData($paymentOnly, $placeOrderPayload, $immutableQuote =
             $taxAmount = 0;
         }
 
-        // include potential rounding difference and reset $diff accumulator
-        $cart['items'][0]['total_amount'] += round($diff);
-        $totalAmount += round($diff);
         $diff = 0;
```
danac-gs: This line should go as well. We apply this rounding diff below in this method. Applying it twice seems redundant and could probably cause errors. So all 3 lines above should be removed.

johnmccombs1: @dan-bolt I've updated the code

danac-gs:
@johnmccombs1 @matt-thomason @njfaries  @daisy1754 @OleksiiOnyshchenko 
The rounding issue that merchant had here is a very rare one and hard to replicate and foresee. I believe this is a good fix, applied to merchant store already, and John and I were discussing it. He was cautious about the regression and my standpoint was to go on with it let QA do extensive testing. So, I suggest paying special attention when doing acceptance testing and spending some time on it with diverse carts. For example, cart items should have various decimal number prices, discounts should be applied, especially percentage type discounts, different ones, many combinations of these should be taken into account. One should just spend a couple of hours making random orders with complex carts. And if the case that does not work is found we should have the exact cart specification for debugging purposes.

daisy1754:
okay your call - still don't understand why test can be added though

danac-gs:
> okay your call - still don't understand why test can be added though

We should add unit test, of course, but can not be sure to catch this case in them. It is a functional / integration issue that pops up later, in order creation.

PR #620, opened 2020-03-27
ENG-375 Added log retrieval to debug endpoint

njfaries:
# Description
Wrote a log retrieval helper for the debug endpoint. This takes in both the path and number of lines to retrieve from the log and returns these lines in array format (easier to read as the json output of the endpoint). No work done for taking in parameters from the POST body but since the method is parameterized expanding the capabilities of the endpoint should not be difficult in future.

Fixes: https://boltpay.atlassian.net/browse/ENG-375?atlOrigin=eyJpIjoiZDczYjliNmI0Y2I1NDkzOWJiNzEwYzMxMTA2YjQ3MmIiLCJwIjoiaiJ9

#changelog Add log retrieval to debug endpoint

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Model/Api/Data/DebugInfo.php
+++ b/Model/Api/Data/DebugInfo.php
@@ -39,6 +39,11 @@ class DebugInfo
 	 */
 	private $otherPluginVersions;
 
+    /**
+     * @var string
+     */
+    private $logInfo;
+
 	/**
 	 * @return string
 	 */
--- a/Model/Api/Data/DebugInfo.php
+++ b/Model/Api/Data/DebugInfo.php
@@ -138,6 +138,12 @@ class DebugInfo implements \JsonSerializable
 	 */
 	public function jsonSerialize()
 	{
-		return get_object_vars($this);
+		return [
+			'php_version' => $this->phpVersion,
+			'platform_version' => $this->platformVersion,
+			'bolt_config_settings' => $this->boltConfigSettings,
+			'other_plugin_versions' => $this->otherPluginVersions
+		];
+
 	}
 }
--- a/Model/Api/Data/DebugInfo.php
+++ b/Model/Api/Data/DebugInfo.php
@@ -108,4 +108,15 @@ class DebugInfo
 		$this->otherPluginVersions = $otherPluginVersions;
 		return $this;
 	}
+
+    /**
+     * @param string $logInfo
+     * @return $this
+     */
+
+	public function setLogInfo($logInfo)
+    {
+        $this->logInfo = $logInfo;
+        return $this;
+    }
 }
--- a/Model/Api/Data/DebugInfo.php
+++ b/Model/Api/Data/DebugInfo.php
@@ -115,6 +115,14 @@ class DebugInfo
 	}
 
     /**
+     * @return string
+     */
+	public function getLogInfo()
+    {
+        return $this->logInfo;
+    }
+
+    /**
      * @param string $logInfo
      * @return $this
      */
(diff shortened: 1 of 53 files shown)


njfaries:
Wasn't sure about viability of using shell commands in php. Would it ever happen that our plugin would be running on a windows server (if such a thing exists)? I agree that it's much simpler if it's a safe assumption to make.

njfaries:
Update: I've been informed the above is a silly question and now with the tail approach it works but broke all my unit tests. The linter is also complaining...

njfaries:
The linter didn't like it when I used backticks without quotes. If I added the quotes, for whatever reason the tail didn't work properly (I only got the "file not found" message). I tried getting around it by just using shell_exec instead but looks like that's forbidden.

@danac-gs I couldn't find any examples of shell commands being invoked in our plugin (searched for "shell_exec" and for backticks), how do you suggest I proceed?

cc: @matt-thomason as well

danac-gs:
@njfaries instead of backtick operator you can use shell_exec. They are identical.

njfaries:
shell_exec is forbidden by the linter

danac-gs:
> shell_exec is forbidden by the linter

Can we bypass it?

matt-thomason:
I think `shell_exec` here is a little too aggressive. I'm in favour of going back to an `fopen` approach, either through your previous code or through a utility such as https://packagist.org/packages/icyapril/tail.

`tail` command is certainly simpler, but it opens up liability for security issues, we need to do a lot more validation on input parameters, and it won't work on filesystems where we (or PHP) have limited permissions.

njfaries:
I've reverted the changes

Test/Unit/Model/Api/DebugTest.php
```diff
@@ -200,7 +213,11 @@ public function debug_successful()
 						'name' => 'plugin3',
 						'version' => '3.0.0'
 					]
-				]
+				],
+                'logInfo' => [
```
Jiyuan-Bolt: Seems like you don't have the latest code. We now use the snake case for field names. Or let's change it to logs for simplicity. Check this PR for reference: https://github.com/BoltApp/bolt-magento2/pull/621

njfaries: Ah yes, this was originally written friday. Will update.

Model/Api/Data/DebugInfo.php
```diff
@@ -118,8 +142,9 @@ public function jsonSerialize()
 			'php_version' => $this->phpVersion,
 			'platform_version' => $this->platformVersion,
 			'bolt_config_settings' => $this->boltConfigSettings,
-			'other_plugin_versions' => $this->otherPluginVersions
+			'other_plugin_versions' => $this->otherPluginVersions,
+            'log_info' => $this->logInfo
```
Jiyuan-Bolt: Let's call it logs.

Jiyuan-Bolt:
Please address @matt-thomason and @danac-gs' comments also

njfaries:
> Have you tested this with a log of typical size on a Magento store? How long have you seen logs get in customer's stores before they overflow into another store, and is there a performance impact from tailing from those logs?

The default is to just tail the last 100 lines. Not exactly sure the typical size of store logs, but we're always sticking to the amount of lines specified (if #lines > existing lines, filesize dictates end). 100 is just a rough guess for v1, not sure in practice how long you'd need to have useful info. If you're just triggering something and then pinging the endpoint you'd probably only need like 10 or so.

matt-thomason:
To be clear - I'm not worried about what we're sending through the endpoint, I'm worried about scanning through like a 1GB size file to get to the end to grab 10 lines.

njfaries:
> To be clear - I'm not worried about what we're sending through the endpoint, I'm worried about scanning through like a 1GB size file to get to the end to grab 10 lines.

Oh I see. If it is a concern, what would the solution be? Is there a better way than fseek()?

njfaries:
For the record, the actual implementation of `tail` uses `lseek`, which is the linux system command equivalent of `fseek` http://git.savannah.gnu.org/cgit/coreutils.git/tree/src/tail.c

PR #651, opened 2020-04-21
Add catalog price rule support for backend orders

mccottry:
# Description
Items which have special prices set are not picked up by Bolt in the Admin checkout

Fixes: https://boltpay.atlassian.net/browse/M2P-44

#changelog Add catalog price rule support for backend orders

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -18,6 +18,7 @@
 namespace Bolt\Boltpay\Helper;
 
 use Bolt\Boltpay\Model\Response;
+use Magento\Framework\Registry;
 use Magento\Framework\Session\SessionManagerInterface as CheckoutSession;
 use Magento\Catalog\Model\ProductRepository;
 use Bolt\Boltpay\Helper\Api as ApiHelper;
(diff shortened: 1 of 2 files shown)


Helper/Cart.php
```diff
@@ -1246,6 +1255,25 @@ public function getCartData($paymentOnly, $placeOrderPayload, $immutableQuote =
         }
 
         $this->setLastImmutableQuote($immutableQuote);
+        if ($this->isBackendSession()) {
```
njfaries, line 1258: To make sure I understand this correctly, the problem was that even when in admin the registered rule_data was using the wrong customer_group_id and thus pulling the wrong price?

mccottry: The issue was that the Bolt custom order creation path missed calling this critical method `Magento/Sales/Model/AdminOrder/Create::initRuleData`

https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Sales/Model/AdminOrder/Create.php#L395-L415

As a result, no rule data was ever set in the registry to override the default values (including `customer_group_id`) that are otherwise taken from the product directly or not recognized at all by the observer that is in charge of setting the final product price in admin orders, (i.e `ProcessAdminFinalPriceObserver`).

https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/CatalogRule/Observer/ProcessAdminFinalPriceObserver.php#L82-L98

PR #660, opened 2020-04-24
Add bolt checkout type to quote

danac-gs:
# Description
Replace **bolt_is_backend_order** quote table column with **bolt_checkout_type**
so we can use it to cover more cases, PPC being one of them.
Needed for shipping and tax split to work.

Fixes: https://app.asana.com/0/941920570700290/1172688451409921/f

#changelog Add bolt checkout type to quote

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -72,6 +72,10 @@ class Cart extends AbstractHelper
     const BOLT_ORDER_TAG = 'Bolt_Order';
     const BOLT_ORDER_CACHE_LIFETIME = 3600; // one hour
 
+    const BOLT_CHECKOUT_TYPE_MULTISTEP = 1;
+    const BOLT_CHECKOUT_TYPE_PPC = 2;
+    const BOLT_CHECKOUT_TYPE_BACKOFFICE = 3;
+
     /** @var CacheInterface */
     private $cache;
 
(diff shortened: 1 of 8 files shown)


Helper/Cart.php
```diff
@@ -1023,7 +1027,9 @@ protected function createImmutableQuote($quote)
     {
         // assign origin session type flag to the parent quote
         // to be replicated to the immutable quote with the other data
-        $quote->setBoltIsBackendOrder( (int) $this->isBackendSession() );
+        if ($this->isBackendSession()) {
+            $quote->setBoltCheckoutType(self::BOLT_CHECKOUT_TYPE_BACKOFFICE);
+        }
```
njfaries, line 1032: What happens if bolt checkout type is left unset?

danac-gs: @njfaries The default value of the field in DB is 1, corresponding to BOLT_CHECKOUT_TYPE_MULTISTEP.

Setup/UpgradeSchema.php
```diff
@@ -56,14 +56,20 @@ public function upgrade(SchemaSetupInterface $setup, ModuleContextInterface $con
             ]
         );
 
+        $setup->getConnection()->dropColumn(
+            $setup->getTable('quote'),
```
matt-thomason, line 60: We will lose the backend order type for old orders on upgrade if we do it this way, right?
I think a safer approach is to write a migration function that'll run if upgrading from bolt version <=2.6.0.

danac-gs: @matt-thomason this field is not propagated to the order table, it exists in quote table only and its sole purpose was defining the right redirect url, storefront vs admin. This functionality is completely preserved with the new field.

matt-thomason: That's a good point, thanks 👀

PR #668, opened 2020-05-05
WIP Working m2 for floating checkout button

pschutz93:
# Description
First attempt at introducing floating checkout button to M2 code base. The floating checkout button. The button appears and enables initiation of checkout from any product or category page when there are one or more items in the cart.

Screen Shot 2020-05-04 at 9 20 03 PM

The button is:
1. Hidden when no items in cart.
2. When an item is added to cart the Buy Now button appears. Minicart animates.
3. When navigating to another product or category page the Buy Now button appears.

To setup the button call `BoltCheckout.configure` with `{floatingButtonOption: "show_cart"}` or `{floatingButtonOption: "show_cart_on_hover"}` as the last argument depending on whether an item was just added or has been there.

To call when an item is added to cart, I set a listener on "ajax:addToCart". I call createOrder with `dataChanged` set to true which triggers configure with {floatingButtonOption: "show_cart"}`.

The tricky part is determining whether an item was just added to cart, has been there, or if there are no items in the cart. I check to see if there are now items in the cart by modifying `monitorDataChange` to watch for cart counter going from empty to a number. When this occurs we call `createOrder` call which triggers `BoltCheckout.configure(cart, hints, callbacks, {floatingButtonOption: "show_cart_on_hover"})`.

Another tricky part is not calling configure from `monitorDataChange` when we already called from the "ajax:addToCart" listener. We do this by checking what changed in `monitorDataChange` to only capture the cart being updated on page load (transitioning from empty string to a number).

Fixes: (link Asana Task)

#changelog WIP Working m2 for floating checkout button

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Block/BlockTrait.php
+++ b/Block/BlockTrait.php
@@ -129,6 +129,10 @@ trait BlockTrait
         return array_unique(array_merge(Config::$defaultPageWhitelist, $values));
     }
 
+    protected function getFloatingButtonWhitelist() {
+        return array_unique(Config::$floatingButtonWhitelist);
+    }
+
     /**
      * Check if the client IP is restricted -
      * there is an IP whitelist and the client IP is not on the list.
--- a/Block/Js.php
+++ b/Block/Js.php
@@ -392,11 +392,25 @@ class Js extends Template
         return in_array($currentPage, $this->getPageWhitelist());
     }
 
+    public function isOnPageFromFloatingButtonWhitelist() {
+        $currentPage = $this->getRequest()->getFullActionName();
+        return in_array($currentPage, $this->getFloatingButtonWhitelist());
+    }
+
+    public function getPageActionName() {
+        return $this->getRequest()->getFullActionName();
+    }
+
+
     /**
      * Return true if bolt on minicart is enabled
      */
     public function isMinicartEnabled() {
         return $this->configHelper->getMinicartSupport();
+    }
+
+    public function isFloatingButtonEnabled() {
+        return true;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


danac-gs:
@pschutz93 can you add the link to Asana / Jira task to the description? I'd like to add some more general comments that would have better visibility there. Anyway, I'll write it here and you can add it to the task.

I've been investigating the idea of removing the DOM monitoring via Mutation Observers (`monitorDataChange, monitorAttributesChange...`) and use the native Magento events for monitoring the cart, customer and quote data changes.

Mutation Observers do the job but are really just a workaround with a lot of unnecessary overhead and cumbersome code to support it. We might leave them only to support arbitrary button injection / replacement, used on the default minicart page and in some special cases.

With events we could just set up listeners to call our handlers which would in turn call creteOrder, configureHints or prefetchShipping. It would simplify / reduce the `replacejs.phtml` code a lot. 

Perhaps this should be done before other changes take place. But it can be done independently and joined later as well. I never had the chance to implement it but would love to. Feel free to create a task and assign it to me.

@matt-thomason @njfaries thoughts?

matt-thomason:
@danac-gs I've been thinking about this a lot as well. Embracing native events will also allow us to write simpler customizations, as well as solve a lot of the issues from [M2P-31], yes. That being said, I think it's fine to enact these changes here and retro-fit them.

I'll bring it up in a chat with Ellen and Usman today - we can probably break down the linked JIRA ticket into two parts, one to replace complex observer logic with events, and a second part to set up better testing/linting for it.

pschutz93:
@usmanismail-bolt This PR is old and closed. I rolled into new PR that we merged. https://github.com/BoltApp/bolt-magento2/pull/705 No need to do anything with this one.

PR #705, opened 2020-05-26
Always Present Checkout button.

pschutz93:
# Description
Introducing the always present checkout button into the M2 code base. The button appears and enables initiation of checkout from any white listed when there are one or more items in the cart. The feature exists behind a feature switch. We respect the Bolt Everywhere flag for minicart and appears on whitelisted pages (the same as minicart). We likely will disable minicart for merchants who use this button but the code should still support minicart and always present checkout button on the same page.

image

The button is:

Hidden when no items in cart.
When an item is added to cart the Buy Now button appears. Minicart animates.
When navigating to another product or category page the Buy Now button appears.
To setup the button call BoltCheckout.configure with {floatingButtonOption: "show_cart"} or {floatingButtonOption: "show_cart_on_hover"} as the last argument depending on whether an item was just added or has been there.

To call when an item is added to cart, I set a listener on "ajax:addToCart". I call createOrder with dataChanged set to true which triggers configure with {floatingButtonOption: "show_cart"}`.

The tricky part is determining whether an item was just added to cart, has been there, or if there are no items in the cart. I check to see if there are now items in the cart by modifying monitorDataChange to watch for cart counter going from empty to a number. When this occurs we call createOrder call which triggers BoltCheckout.configure(cart, hints, callbacks, {floatingButtonOption: "show_cart_on_hover"}).

Another tricky part is not calling configure from monitorDataChange when we already called from the "ajax:addToCart" listener. We do this by checking what changed in monitorDataChange to only capture the cart being updated on page load (transitioning from empty string to a number).

See Product Spec here: https://docs.google.com/document/d/1EoTBhNeRPAj-RupFczAaLHakYgqonYMJvAjMFIm-EFk/edit#heading=h.dfj49q9kev35

See Technical Spec here: 
https://docs.google.com/document/d/1wP962zqqvyBPWQx7F3yXA48a8BtX1lN1C5Ni0hVqncY/edit

Fulfills:
https://app.asana.com/0/1146941972725303/1165704788287983/f
https://app.asana.com/0/1146941972725303/1176822385483791/f
https://app.asana.com/0/1146941972725303/1176822385483793/f

#changelog Always Present Checkout button.
Adds support for Always Present Checkout button. Adds feature switch for button.

# Type of change

- [X] New feature (change which adds functionality)

# How Has This Been Tested?

- [X] Successfully tested locally (or docker image)

# Checklist:

- [X] My code follows the style guidelines of this project.
- [X] I have performed a self-review of my own code.
- [X] I have commented my code, particularly in hard-to-understand areas.
- [X] New and existing unit tests pass locally with my changes.
	- Added unit tests for feature switch. JS changes not super unit testable.
- [X] I have added my Jira ticket link and provided a changelog message.

### DIFF
--- a/Block/Js.php
+++ b/Block/Js.php
@@ -226,6 +226,7 @@ class Js extends Template
             'is_pre_auth'              => $this->getIsPreAuth(),
             'default_error_message'    => $this->getBoltPopupErrorMessage(),
             'button_css_styles'        => $this->getButtonCssStyles(),
+            'always_present_checkout' => $this->featureSwitches->isAlwaysPresentCheckoutEnabled(),
         ]);
     }
 
--- a/Helper/Config.php
+++ b/Helper/Config.php
@@ -1714,15 +1714,15 @@ class Config extends AbstractHelper
 
     /**
      * Get config value for whether or not always present checkout is enabled
-     * @param int|string|Store $store
+     * @param int|string|Store $storeId
      * @return boolean
      */
-    public function isAlwaysPresentCheckoutEnabled($store = null)
+    public function isAlwaysPresentCheckoutEnabled($storeId = null)
     {
         return $this->getScopeConfig()->isSetFlag(
             self::XML_PATH_ALWAYS_PRESENT_CHECKOUT,
             ScopeInterface::SCOPE_STORE,
-            $store
+            $storeId
         );
     }
 
--- a/Helper/FeatureSwitch/Definitions.php
+++ b/Helper/FeatureSwitch/Definitions.php
@@ -91,6 +91,8 @@ class Definitions {
      */
     const M2_MERCHANT_METRICS = "M2_MERCHANT_METRICS";
 
+    const M2_ALWAYS_PRESENT_CHECKOUT = "M2_ALWAYS_PRESENT_CHECKOUT";
+
     const DEFAULT_SWITCH_VALUES = array(
         self::M2_SAMPLE_SWITCH_NAME =>  array(
           self::NAME_KEY            => self::M2_SAMPLE_SWITCH_NAME,
@@ -157,6 +159,12 @@ class Definitions {
             self::VAL_KEY             => true,
             self::DEFAULT_VAL_KEY     => false,
             self::ROLLOUT_KEY         => 100
+        ),
+        self::M2_ALWAYS_PRESENT_CHECKOUT => array(
+            self::NAME_KEY => self::M2_ALWAYS_PRESENT_CHECKOUT,
+            self::VAL_KEY => true,
+            self::DEFAULT_VAL_KEY => false,
+            self::ROLLOUT_KEY => 0,
         )
     );
 }
--- a/etc/adminhtml/system.xml
+++ b/etc/adminhtml/system.xml
@@ -95,6 +95,11 @@
                         <config_path>payment/boltpay/order_management</config_path>
                         <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
                     </field>
+                    <field id="always_present_checkout" translate="label" type="select" sortOrder="83" showInDefault="0" showInWebsite="0" showInStore="0">
+                        <label>(BETA) Enable Always Present Checkout Button</label>
+                        <config_path>payment/boltpay/always_present_checkout</config_path>
+                        <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                    </field>
                     <field id="button_color" translate="label" type="text" sortOrder="87" showInDefault="1" showInWebsite="1" showInStore="0">
                         <label>Primary Color</label>
                         <config_path>payment/boltpay/button_color</config_path>
--- a/etc/adminhtml/system.xml
+++ b/etc/adminhtml/system.xml
@@ -95,7 +95,7 @@
                         <config_path>payment/boltpay/order_management</config_path>
                         <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
                     </field>
-                    <field id="always_present_checkout" translate="label" type="select" sortOrder="83" showInDefault="0" showInWebsite="0" showInStore="0">
+                    <field id="always_present_checkout" translate="label" type="select" sortOrder="83" showInDefault="1" showInWebsite="1" showInStore="0">
                         <label>(BETA) Enable Always Present Checkout Button</label>
                         <config_path>payment/boltpay/always_present_checkout</config_path>
                         <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -45,7 +45,6 @@ function($argName) {
 ob_start();
 ?>
 <script type="text/javascript">
-
     // Store the configuration parameters passed from the php block
     // in the global object. Used in this file and on the payment page
     // in payment method renderer, vendor/boltpay/bolt-magento2/view/frontend/web/js/view/payment/method-renderer/boltpay.js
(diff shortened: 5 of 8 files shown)


Block/Js.php
```diff
@@ -226,6 +226,7 @@ public function getSettings()
             'is_pre_auth'              => $this->getIsPreAuth(),
             'default_error_message'    => $this->getBoltPopupErrorMessage(),
             'button_css_styles'        => $this->getButtonCssStyles(),
+            'always_present_checkout' => $this->featureSwitches->isAlwaysPresentCheckoutEnabled(),
```
matt-thomason, line 229: You've used feature switches here which is good, but something like this should also be controllable by the merchant in their plugin settings. Look at `pay_by_link_url` a few lines above for an example. Can you add that behaviour as well so it's toggle-able in M2?

matt-thomason:
There also appear to be conflicts - please do a rebase @pschutz93 :)

Block/Js.php
pschutz93: To do this do I just add a config path and call something like `$this->getCustomURLValueOrDefault(self::ALWAYS_PRESENT_CHECKOUT_CONFIG_PATH, false);` and define `ALWAYS_PRESENT_CHECKOUT_CONFIG_PATH = "payments/boltpay/always_present_checkout"`? Is there anywhere to define a toggle-able switch that shows up in setings panel?

matt-thomason: Yes, in `etc/config.xml`.

matt-thomason: You also need it in `etc/adminhtml/system.xml`.

pschutz93: Fixed up

Helper/Config.php
```diff
@@ -1706,6 +1712,20 @@ public function isStorePickupFeatureEnabled($storeId = null)
         );
     }
 
+    /**
+     * Get config value for whether or not always present checkout is enabled
+     * @param int|string|Store $store
+     * @return boolean
+     */
+    public function isAlwaysPresentCheckoutEnabled($store = null)
```
matt-thomason, line 1720: Our convention here is `$storeId`

etc/adminhtml/system.xml
```diff
@@ -95,6 +95,11 @@
                         <config_path>payment/boltpay/order_management</config_path>
                         <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
                     </field>
+                    <field id="always_present_checkout" translate="label" type="select" sortOrder="83" showInDefault="0" showInWebsite="0" showInStore="0">
+                        <label>(BETA) Enable Always Present Checkout Button</label>
```
matt-thomason, line 99: @njfaries An idle thought here - maybe we should consider having a section for developer options/beta features, similar to what we have in the admin dash. Then our beta features aren't in the middle of the regular config settings.

Out of scope for this PR but it had me thinking.

view/frontend/templates/js/replacejs.phtml
```diff
@@ -806,8 +805,12 @@ ob_start();
                         createRequest = false;
                     })
             });
+            // Get number of items in cart from counter element
+            var numItems = $(".counter-number").text();
 
-            var BC = BoltCheckout.configure(cart, hintBarrier.promise, callbacks);
+            // Call config with floatingButtonMode if always present checkout enabled and items were or were not just added to cart (dataChanged)
+            var config = boltConfig.always_present_checkout && (numItems && numItems !=="0" || dataChanged)? {floatingButtonMode: dataChanged ? "show_cart" : "show_cart_on_hover" } : {};
+            var BC = BoltCheckout.configure(cart, hintBarrier.promise, callbacks, config);
```
vitaliyreznikov, line 813: We have three configure call in this file. Could you please add your code to two others as well? It's possible that numItems still 0 at this moment because data is still loading. Later we re-configure bolt, but from another place and don't calculate config.
I think we can just copy-paste the code. We are planning to refactor this JS code soon.

etc/adminhtml/system.xml
```diff
@@ -95,6 +95,11 @@
                         <config_path>payment/boltpay/order_management</config_path>
                         <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
                     </field>
+                    <field id="always_present_checkout" translate="label" type="select" sortOrder="83" showInDefault="0" showInWebsite="0" showInStore="0">
```
vitaliyreznikov, line 98: Do you want to hide this setting?

view/frontend/templates/js/replacejs.phtml
```diff
@@ -859,6 +862,9 @@ ob_start();
             checkout_buttons = true;
             processButtons();
         });
+        if (boltConfig.always_present_checkout) {
+            processButtons();
+        }
```
vitaliyreznikov, line 867: when we have in cart both simple and configurable product and click “add to cart” (I do it for the same product that is already on cart)
When we check .block-minicart .subtitle it related to two different DOM nodes: subtitle of minicart and subtitle of configurable product
Function monitorDataChange thinks this was update and call createOrder() again and again.

If we delete lines 865-867 then after few attempts this reloading is finished.
I propose to remove them and completely fix the issue later when we refactor this JS file.

Helper/FeatureSwitch/Definitions.php
```diff
@@ -168,6 +173,12 @@ class Definitions {
             self::VAL_KEY             => true,
             self::DEFAULT_VAL_KEY     => false,
             self::ROLLOUT_KEY         => 0
+        ),
+        self::M2_ALWAYS_PRESENT_CHECKOUT => array(
+            self::NAME_KEY => self::M2_ALWAYS_PRESENT_CHECKOUT,
+            self::VAL_KEY => true,
+            self::DEFAULT_VAL_KEY => false,
+            self::ROLLOUT_KEY => 0,
```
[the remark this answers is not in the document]
pschutz93: I want to default rollout to off and require both config and bolt feature switch to be enabled. This is a brand new feature so it makes sense to have disabled by default even for merchants that turn on. We also at this time don't want merchants to be able to enable always present checkout on their own.

etc/adminhtml/system.xml
pschutz93: I want to show it. This way a merchant who we enable through feature switch can also turn it off if they want.

view/frontend/templates/js/replacejs.phtml
pschutz93: Updated. I only want to call with "show_cart" when a new item is added (when ajax:addToCart event is fired) so for other two places I call with only "show_cart_on_hover" and remove dataChanged param.

view/frontend/templates/js/replacejs.phtml
pschutz93: Removed

view/frontend/templates/js/replacejs.phtml
pschutz93: Actually we can't call everywhere. The issue is that we only want to call with config when a cart with items is created (which to my knowledge happens only in `createOrder`. If I call with config later `BC.configure` is called with a null cart. This breaks always present button as it depends on having cart to show. We only want to call it when passing a cart that is not null.

pschutz93: I also changed monitorChagnes function quite a bit which should make fix edge cases where number hasn't updated uet.

vitaliyreznikov: On my test site when I am on product page:
1) The first configure call happens on line 813. numItem is "0" at this moment, we don't show always present button
2) The second call happens on line 916. Cart is set and we can check $(".counter-number").text() and show button if we copy the same code before line 916.

So for now on my test site button is shown if I add product to cart, but isn't shown if reload the page.

etc/adminhtml/system.xml
vitaliyreznikov: I see. Then we need set parameters to:
**showInDefault="1" showInWebsite="1" showInStore="0"**

PR #716, opened 2020-05-30
Add observer on order creation event to injest non-Bolt orders

nickfiacco-bolt:
# Description
This observer will trigger on all completed orders, and conditionally send non-Bolt order information to the Bolt backend. This is part of the non-Bolt order management project, which will let shoppers see all of the purchases in the Bolt shopper dashboard regardless of how they paid.

Tested locally with my Magento 2 devbox-- Bolt orders were ignored and non-Bolt order information was successfully sent to the Hail backend and stored in the DB.

#changelog Add observer on order creation event to injest non-Bolt orders
Add observer for all completed orders to ingest non-Bolt order information.

### DIFF
--- a/Observer/NonBoltOrderObserver.php
+++ b/Observer/NonBoltOrderObserver.php
@@ -221,6 +221,9 @@ class NonBoltOrderObserver implements ObserverInterface
             if (count($names) > 1) {
                 $address->setFirstName($names[0]);
                 $address->setLastName($names[1]);
+            } else if (count($names) > 2) {
+                $address->setFirstName(array_shift($names));
+                $address->setLastName(join(" ", $names));
             } else {
                 $address->setLastName($name);
             }
--- a/Observer/NonBoltOrderObserver.php
+++ b/Observer/NonBoltOrderObserver.php
@@ -10,6 +10,7 @@ use Bolt\Boltpay\Helper\Log as LogHelper;
 use Bolt\Boltpay\Helper\MetricsClient;
 use Bolt\Boltpay\Model\Payment;
 use Bolt\Boltpay\Model\Response;
+use Error;
 use Exception;
 use Magento\Customer\Api\Data\CustomerInterface;
 use Magento\Framework\DataObjectFactory;
@@ -133,6 +134,11 @@ class NonBoltOrderObserver implements ObserverInterface
         } catch (Exception $exception) {
             $this->metricsClient->processCountMetric("non_bolt_order_creation.failure", 1);
             $this->bugsnag->notifyException($exception);
+            return;
+        } catch (Error $error) {
+            // catch errors so failures here don't prevent orders from being created
+            $this->metricsClient->processCountMetric("non_bolt_order_creation.failure", 1);
+            $this->bugsnag->notifyException($error);
             return;
         }
 
--- a/Observer/NonBoltOrderObserver.php
+++ b/Observer/NonBoltOrderObserver.php
@@ -218,10 +218,7 @@ class NonBoltOrderObserver implements ObserverInterface
         if ($address->getLastName() == null) {
             $name = $address->getFirstName();
             $names = preg_split("/[\s]+/", $name);
-            if (count($names) > 1) {
-                $address->setFirstName($names[0]);
-                $address->setLastName($names[1]);
-            } else if (count($names) > 2) {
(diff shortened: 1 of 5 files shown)


Observer/NonBoltOrderObserver.php
```diff
@@ -0,0 +1,201 @@
...
+     * Call Bolt Create Non-Bolt Order API
+     *
+     * @param array $cart
+     * @param CustomerInterface $customer
+     * @param int $storeId
+     * @return Response|int
+     * @throws LocalizedException
+     * @throws Zend_Http_Client_Exception
+     */
+    protected function createNonBoltOrder($cart, $customer, $storeId)
```
danac-gs, line 158: Does the `$customer` always have associated email and phone data? First and last names also. What happens in case of a guest checkout? As a fallback, if needed, this data can be collected from the quote and / or quoteAddress. Please investigate.

nickfiacco-bolt: In all of my testing, the `$customer` had the necessary information. If it comes up, I can always add that later.

Observer/NonBoltOrderObserver.php
```diff
@@ -0,0 +1,229 @@
...
+        return $this->apiHelper->sendRequest($request);
+    }
+
+    /**
+     * Get phone number from the cart if it exists
+     *
+     * @param array $cart
+     * @return string
+     */
+    protected function getPhone($cart)
```
njfaries, line 194: I believe this whole function could be replaced with `return $cart['shipments'][0]['shipping_address']['phone'] ?? null;` In fact, since you're returning null anyways for it not being set you might be able to skip the coalescing altogether.

That's a php7 thing, so it may break on older instances, in which case `isset()` might be the ticket.

nickfiacco-bolt: I originally had that, but got an error when the key was missing

njfaries: interesting. I wonder if it's the `[0]` that breaks it

nickfiacco-bolt: Added a test-- it's because if the `'shipments'` key isn't even set, then it will break. I.e. it will work if `$cart['shipments']==null`, but not if `$cart` doesn't have the key `shipments`

Observer/NonBoltOrderObserver.php
```diff
@@ -0,0 +1,229 @@
...
+     * @param $quote
+     */
+    protected function handleMissingName($quote) {
+        $address = $quote->isVirtual() ? $quote->getBillingAddress() : $quote->getShippingAddress();
+        if ($address->getLastName() == null) {
+            $name = $address->getFirstName();
+            $names = preg_split("/[\s]+/", $name);
+            if (count($names) > 1) {
+                $address->setFirstName($names[0]);
+                $address->setLastName($names[1]);
```
danac-gs, line 223: Taking into account  cases when `count($names) > 2` you could set the first name to the first element and the last name to everything that's left, something like:
```php
$address->setFirstName(array_shift($names));
$address->setLastName(join(" ", $names));
```

nickfiacco-bolt: ah good idea!

Observer/NonBoltOrderObserver.php
```diff
@@ -0,0 +1,232 @@
...
+            } else if (count($names) > 2) {
```
danac-gs, line 224: If the previous condition, `if (count($names) > 1)`, is fulfilled you never get here.
The new block below should replace the old block above cause it covers both (all) cases.

Observer/NonBoltOrderObserver.php
```diff
@@ -0,0 +1,232 @@
...
+                $address->setFirstName(array_shift($names));
+                $address->setLastName(join(" ", $names));
+            } else {
+                $address->setLastName($name);
+            }
```
danac-gs, line 229:
```suggestion
           if (count($names) > 1) {
                $address->setFirstName(array_shift($names));
                $address->setLastName(join(" ", $names));
            } else {
                $address->setLastName($name);
            }
```

Observer/NonBoltOrderObserver.php
danac-gs: see below

PR #718, opened 2020-06-01
[MX-504] Added ES version check for IE11 compatibility

karkevin:
# Description
- bandaid fix for ES version compatibility for IE11 users
- added git pre commit check using eslint and husky
- No clean way of checking for ES version in `.phtml` files using composer, so added some npm dev dependencies. These can be removed once refactoring to move JS files out of M2 into source is accomplished. Dependencies can then be accessed from source. 

- A more sustainable change would be to use a transpiler (such as babel and webpack) to convert es6+ syntax into es5 during the build stage. This solution can be further looked into after the JS refactor, and assets are loaded into M2 from source.

Fixes: [MX-504](https://boltpay.atlassian.net/browse/MX-504)

#changelog [MX-504] Added ES version check for IE11 compatibility
- added pre commit hooks and linting instructions
- modified JS from es6 to es5 syntax in `boltaccount.phtml`

# Type of change

- [X] New feature (change which adds functionality)

# How Has This Been Tested?

- [X] Successfully tested locally (or docker image)

# Checklist:

### DIFF
--- a/view/frontend/templates/boltaccount.phtml
+++ b/view/frontend/templates/boltaccount.phtml
@@ -51,7 +51,7 @@ $checkoutKey = $block->getCheckoutKey();
                 // When magento constructs mobile menu it copies elements instead of moving
                 // We need to wait for this coping, remove first bolt-account-login div
                 // And only ofter that insert account.js script
-                let timerId = setTimeout(function boltAccountLookup() {
+                var timerId = setTimeout(function boltAccountLookup() {
                     var $account_div = $("div.bolt-account-login");
                     if ($account_div.length==2) {
                         $account_div.eq(0).remove();
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,4 @@
+# ignore the test folder
+Test
+
+.gitignore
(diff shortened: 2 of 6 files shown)


view/frontend/templates/boltaccount.phtml
```diff
@@ -51,7 +51,7 @@ $checkoutKey = $block->getCheckoutKey();
                 // When magento constructs mobile menu it copies elements instead of moving
                 // We need to wait for this coping, remove first bolt-account-login div
                 // And only ofter that insert account.js script
-                let timerId = setTimeout(function boltAccountLookup() {
+                var timerId = setTimeout(function boltAccountLookup() {
```
njfaries, line 54: This seems unrelated?

njfaries: nvm missed bit in description

karkevin: @njfaries `let` is a keyword present in es6, which is not supported in es5 for defining variables. Unsure if it causes any problems with IE, but the change is required in order to fulfill the pre-commit hook.

karkevin: This shouldn't be an issue regardless, since `timerId` is only defined in the block and not called globally.

PR #723, opened 2020-06-05
M2P-85 Create an offline Magento invoice and move order status to processing after placing an order using a saved card in the back-office

johnmccombs1:
# Description
Currently, after the admin creates an order using a saved card in the back-office, the order is created and stuck in the pending status as the Bolt recharged transaction doesn’t have webhooks.

This PR creates an offline Magento invoice for the recharged order and move its status to processing.

Fixes: https://boltpay.atlassian.net/browse/M2P-85

#changelog Create an offline Magento invoice and move order status to processing after placing an order using a saved card in the back-office

# Type of change

# How Has This Been Tested?

# Checklist:


johnmccombs1:
@danac-gs  because the recharged transaction is a direct payment. Bolt doesn’t authorize and capture it. It just goes to complete status right away after its creation. It’s just like an auto-capture transaction without the hooks being sent to Magento.

Plugin/OrderPlugin.php
```diff
@@ -48,6 +63,11 @@ public function beforeSetState(Order $subject, $state)
         if (!$subject->getPayment() || $subject->getPayment()->getMethod() != \Bolt\Boltpay\Model\Payment::METHOD_CODE) {
             return [$state];
         }
+
+        if ($subject->getIsRechargedOrder()){
```
vitaliyreznikov, line 67: Could you add a comment while we need this?

johnmccombs1: @vitaliyreznikov I've updated

Plugin/OrderPlugin.php
```diff
@@ -69,6 +90,12 @@ public function beforeSetStatus(Order $subject, $status)
         if (!$subject->getPayment() || $subject->getPayment()->getMethod() != \Bolt\Boltpay\Model\Payment::METHOD_CODE) {
             return [$status];
         }
+
+        if ($subject->getIsRechargedOrder()) {
+            // Check and set the recharged order state to processing
```
vitaliyreznikov, line 95: Could you add a comment while we need set status to processing in this case? What happens if we don't do this?

PR #725, opened 2020-06-08
Reward points need to be displayed in the minicart [EN-188]

mccottry:
# Description
This PR adds support for Magento EE Reward Points to minicart.

Fixes: https://boltpay.atlassian.net/browse/EN-188

#changelog Reward points need to be displayed in the minicart [EN-188]

# Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Plugin/Magento/Rewards/Controller/Cart/RemoveActionPlugin.php
+++ b/Plugin/Magento/Rewards/Controller/Cart/RemoveActionPlugin.php
@@ -36,7 +36,7 @@ class RemoveActionPlugin
     }
 
     /**
-     * The original behavior, intended for the shopping cart page is to trigger
+     * The original behavior intended for the shopping cart page is to trigger
      * the browser to reload by sending a "Location" header. This is ok
      * from the shopping cart page, but we do not want this from the mini-cart.
      * To improve UX we remove the "Location" header so that the page does not
--- a/Plugin/Magento/Rewards/Controller/Cart/RemoveActionPlugin.php
+++ b/Plugin/Magento/Rewards/Controller/Cart/RemoveActionPlugin.php
@@ -1,0 +1,61 @@
+<?php
+/**
+ * Bolt magento2 plugin
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ *
+ * @category   Bolt
+ * @package    Bolt_Boltpay
+ * @copyright  Copyright (c) 2017-2020 Bolt Financial, Inc (https://www.bolt.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+namespace Bolt\Boltpay\Plugin\Magento\Rewards\Controller\Cart;
+
+/**
+ * Plugin for {@see \Magento\Reward\Controller\Cart\Remove} used to add AJAX support
+ */
+class RemoveActionPlugin
+{
+    /**
+     * @var \Bolt\Boltpay\Helper\Config Bolt configuration helper instance
+     */
+    private $configHelper;
+
+    /**
+     * @param \Bolt\Boltpay\Helper\Config $configHelper Bolt configuration helper instance
+     */
+    public function __construct(\Bolt\Boltpay\Helper\Config $configHelper)
+    {
+        $this->configHelper = $configHelper;
+    }
+
+    /**
+     * The original behavior, intended for the shopping cart page is to trigger
+     * the browser to reload by sending a "Location" header. This is ok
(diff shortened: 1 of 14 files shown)


view/frontend/templates/checkout/minicart_addons.phtml
```diff
@@ -0,0 +1,76 @@
...
+                            );
+                        }
+                    });
+                });
+                <?php endif; ?>
+            });
+        }
+
+        if (typeof window.checkoutConfig == 'undefined') {
+            $.post('<?php echo $block->getUrl('boltpay/cart/checkoutconfig'); ?>').success(function (result) {
```
vitaliyreznikov, line 61: Could we use get instead of post? Post triggers all magento2 sections reload.

Plugin/Magento/Rewards/Controller/Cart/RemoveActionPlugin.php
```diff
@@ -0,0 +1,56 @@
...
+     * Adds support for AJAX calls by removing redirect from response for AJAX requests
+     *
+     * @see \Magento\Reward\Controller\Cart\Remove::execute
+     *
+     * @param \Magento\Reward\Controller\Cart\Remove        $subject the frontend controller wrapped by this plugin
+     * @param void|\Magento\Framework\App\ResponseInterface $result original result of the wrapped method
+     *
+     * @return void|\Magento\Framework\App\ResponseInterface original result of the wrapped method
+     */
+    public function afterExecute(\Magento\Reward\Controller\Cart\Remove $subject, $result)
```
vitaliyreznikov, line 48: Why we need this?

view/frontend/templates/checkout/minicart_addons.phtml
mccottry: @vitaliyreznikov `POST` was used to prevent JQuery from caching the request.  But, yes, we can and will use `GET` with the extra Jquery parameter `cacheable:false` to achieve this goal.   

This will slightly optimize the code because not ALL sections reload with `POST` but rather ONLY the messages section; And you are right, this minor load is not needed for this particular request.

Plugin/Magento/Rewards/Controller/Cart/RemoveActionPlugin.php
mccottry: I'll expand my comment in the code.  Which was 

> Adds support for AJAX calls by removing redirect from response for AJAX requests

To expand on this, what we are doing is greatly improving the UX experience.  We are reusing components which are designed for the Shopping Cart page.  The design on that page for reward points removal is to send a location header to force the page to reload.  This is ok from the shopping cart page, but we do not want this from the mini-cart.  We use ajax for a smoother user experience and remove the location header so that the page does not randomly reload the user's page.  This is consistent with other mini-cart edits.

mccottry: Comment updated.

view/frontend/templates/checkout/minicart_addons.phtml
```diff
@@ -0,0 +1,82 @@
...
+                            );
+                        }
+                    });
+                });
+                <?php endif; ?>
+            });
+        }
+
+        if (typeof window.checkoutConfig == 'undefined') {
+            $.ajax({
```
danac-gs, line 61: NOTE: it seems that jQuery `$.get` is configured in Magento to not cache the responses. Anyway, this is the safe way.

mccottry: thanks.  That's great to know.

mccottry: btw, @danac-gs , can you post some documentation or a reference that confirms this?  I'm having trouble finding this.

mccottry: Nevermind.  I found confirmation:

view/frontend/templates/checkout/minicart_addons.phtml
vitaliyreznikov: I don't think this necessary, but magento2 code has
```
    $.ajaxSetup({
        cache: false
    });
```

So looks like we can skip GET parameters, just to be short.
But this is up to you

Plugin/Magento/Rewards/Controller/Cart/RemoveActionPlugin.php
vitaliyreznikov: Got it. thank you. So looks like we change UX experience for M2 module.
Maybe it makes sense to add this into code comments.

view/frontend/templates/checkout/minicart_addons.phtml
mccottry: yep. I discovered this and commented on this yesterday in the comments below. 
https://github.com/BoltApp/bolt-magento2/pull/725#discussion_r437519342

Let's opt for the tried and true (@danac-gs calls it safer) approach that makes sure that if` `$.ajaxSetup` was called from elsewhere with conflicting configs, it does not affect our code.

PR #737, opened 2020-06-12
Send payment method in Bolt request and set transaction reference

nickfiacco-bolt:
Also add more robust methods to pull email, first name, and last name
from the Cart/Customer objects.

#changelog Send payment method in Bolt request and set transaction reference
Set payment method in request

### DIFF
--- a/Observer/NonBoltOrderObserver.php
+++ b/Observer/NonBoltOrderObserver.php
@@ -109,7 +109,8 @@ class NonBoltOrderObserver implements ObserverInterface
             }
 
             $payment = $order->getPayment();
-            if ($payment && $payment->getMethod() == Payment::METHOD_CODE) {
+            $paymentMethod = $payment ? $payment->getMethod() : "unknown";
+            if ($paymentMethod == Payment::METHOD_CODE) {
                 // ignore Bolt orders
                 return;
             }
@@ -131,11 +132,15 @@ class NonBoltOrderObserver implements ObserverInterface
             $cart = $this->cartHelper->buildCartFromQuote($quote, $quote, $items, null, true, false);
             $customer = $quote->getCustomer();
             $storeId = $order->getStoreId();
-            $result = $this->createNonBoltOrder($cart, $customer, $storeId);
-            if ($result != 200) {
+            $result = $this->createNonBoltOrder($cart, $customer, $storeId, $paymentMethod);
+            $response = $result->getResponse();
+            if (empty($response)) {
                 $this->metricsClient->processCountMetric("non_bolt_order_creation.failure", 1);
                 return;
             }
+
+            $payment->setAdditionalInformation("transaction_reference", $response->reference);
+            $payment->save();
         } catch (Exception $exception) {
             $this->metricsClient->processCountMetric("non_bolt_order_creation.failure", 1);
             $this->bugsnag->notifyException($exception);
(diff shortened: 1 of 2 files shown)


Observer/NonBoltOrderObserver.php
```diff
@@ -208,6 +217,84 @@ protected function getPhone($cart)
...
+     * @param CustomerInterface $customer
+     * @return string
+     */
+    protected function getEmail($cart, $customer)
+    {
+        if (!is_null($customer->getEmail())) {
+            return $customer->getEmail();
+        }
+
+        if (!@$cart['shipments']) {
```
vitaliyreznikov, line 231: I think the most native PHP way for this
```
if (isset($cart['shipments'][0]['shipping_address'])) {
 return $cart['shipments'][0]['shipping_address']['email'];
}
```
This code works ok even if $cart['shipments'] undefined

Observer/NonBoltOrderObserver.php
```diff
@@ -131,11 +132,15 @@ public function execute(Observer $observer)
...
-            $result = $this->createNonBoltOrder($cart, $customer, $storeId);
-            if ($result != 200) {
+            $result = $this->createNonBoltOrder($cart, $customer, $storeId, $paymentMethod);
+            $response = $result->getResponse();
+            if (empty($response)) {
                 $this->metricsClient->processCountMetric("non_bolt_order_creation.failure", 1);
                 return;
             }
+
+            $payment->setAdditionalInformation("transaction_reference", $response->reference);
```
danac-gs, line 142: Why are we setting "transaction_reference" to a payment instance that is not related to Bolt?
We could overwrite something essential for the particular method (payment gateway) to work correctly.

Observer/NonBoltOrderObserver.php
nickfiacco-bolt: Are you sure? I just tried this and received an error:
```
Notice: Undefined index: shipments in /Applications/MAMP/htdocs/magento2/vendor/boltpay/bolt-magento2/Observer/NonBoltOrderObserver.php:206.
```

Observer/NonBoltOrderObserver.php
nickfiacco-bolt: Good point-- I will use `bolt_transaction_reference` so as to avoid overwriting anything.

We need the Bolt transaction reference to be set in order to ingest tracking information, which we want to do for all orders:

https://github.com/BoltApp/bolt-magento2/blob/master/Observer/TrackingSaveObserver.php#L119

Observer/NonBoltOrderObserver.php
nickfiacco-bolt: wait never mind I wrote the test wrong

Observer/NonBoltOrderObserver.php
```diff
@@ -197,15 +206,65 @@ protected function getPhone($cart)
             return null;
         }
 
-        if (count($cart['shipments']) < 1) {
-            return null;
+        if (isset($cart['shipments'][0]['shipping_address'])) {
```
danac-gs, line 209: this condition covers the previous one so that one can be removed

nickfiacco-bolt: oops meant to include that as well

PR #741, opened 2020-06-13
M2P-82 Further JS refactoring

vitaliyreznikov:
In this PR:
- Remove legacy code we added for support multistep&paymentonly on the same page (we don't use this option for years)
- Mark code that we need only for payment only (wrap up this code into if statement). I intentionally don't format indents for PR readability. Will format in separate PR
- Remove unnecessary DOM listener - we no need to listen to "sum" changes anymore.

Fixes: [M2P-82]

#changelog M2P-82 Further JS refactoring

# Type of change

- [x] Refactoring

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -615,7 +615,7 @@ ob_start();
             // do not run the same request twice
             if (prefetch_list.indexOf(params) == -1) {
                 prefetch_list.push(params);
-                $.post(settings.shipping_prefetch_url, params);
+                $.get(settings.shipping_prefetch_url, params);
             }
         };
 
@@ -794,9 +794,10 @@ ob_start();
                             BC.open();
                             allowAutoOpen = false;
                         }
+
+                        // prefetch Shipping and Tax for multi-step checkout
+                        prefetchShipping();
                     });
-            // prefetch Shipping and Tax for multi-step checkout
-            prefetchShipping();
 
             if (!wasConfigureCalled) {
                 // temporary disable always present checkout button for refactoring reason
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -420,7 +420,7 @@ ob_start();
         };
         //////////////////////////////////////////////
 
-        // flag that repesents should the checkout auto-open
+        // flag that represents should the checkout auto-open
         var initiateCheckout = settings.initiate_checkout && getInitiateCheckoutCookie();
         unsetInitiateCheckoutCookie();
 
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -587,8 +587,6 @@ ob_start();
             }
         };
 
-        if (getCheckoutType()=="payment") {
-
         /////////////////////////////////////////////////////
         // Call Shipping and Tax prefetch & cache endpoint
         /////////////////////////////////////////////////////
@@ -620,7 +618,6 @@ ob_start();
                 $.post(settings.shipping_prefetch_url, params);
             }
         };
-        }
 
         /////////////////////////////////////////////////////
         // Create Bolt order and configure BoltCheckout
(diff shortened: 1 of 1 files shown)


view/frontend/templates/js/replacejs.phtml
```diff
@@ -587,7 +587,7 @@ ob_start();
             }
         };
 
-        ////////////////////////////////////
+        if (getCheckoutType()=="payment") {
```
danac-gs, line 590: Pre-fetch should be called on multi step checkout, not payment only.

view/frontend/templates/js/replacejs.phtml
```diff
@@ -1017,6 +994,7 @@ ob_start();
         };
         ///////////////////////////////////////////////////////////
 
+        if (getCheckoutType()=="payment") {
```
danac-gs, line 997: Same here.

view/frontend/templates/js/replacejs.phtml
vitaliyreznikov: How I can emulate this mode on my local site?

view/frontend/templates/js/replacejs.phtml
vitaliyreznikov: The same question. How can I emulate this?

view/frontend/templates/js/replacejs.phtml
danac-gs: @vitaliyreznikov do we support payment only checkout? Was it ever used? I thought we should remove it.

vitaliyreznikov: I asked @Abhinay-bolt and @usmanismail-bolt and as I understand we decided to save payment only for now.

About pre-fetch. Do I understand correctly we use it only to save shipping calculation in cache and get it from cache on shipping&tax call?

view/frontend/templates/js/replacejs.phtml
vitaliyreznikov: Do I understand correctly? In multistep we want to use only three fields for hints: country, State/Province and Zip code?

vitaliyreznikov: image

danac-gs: @vitaliyreznikov those 3 fields are enough to get the shipping pre-fetch / estimation. In many cases the actual street address does not make a difference. Even the state is not mandatory, not all countries have states or regions. So yes, we call prefetch with whatever data we have available and save the results in cache. However, when prefetch is triggered it does a few more calculations if possible, based on the following
1. address saved to quote, if any
2. logged in customer default address, if any
3. address determined by geolocation service (https://ipstack.com/), if configured.

vitaliyreznikov: Ok, so let me don't change it for now. Updated.

view/frontend/templates/js/replacejs.phtml
vitaliyreznikov: Fixed

view/frontend/templates/js/replacejs.phtml
```diff
@@ -800,6 +794,8 @@ ob_start();
                             allowAutoOpen = false;
                         }
                     });
+            // prefetch Shipping and Tax for multi-step checkout
+            prefetchShipping();
```
mccottry, line 798: This makes prefetch happen unnecessarily often.
consider bundling this with with the `boltcart` customer section, or
create a new `bolthints` customer section have it invalidate on `/V1/carts/`*/estimate-shipping-methods` and `/V1/guest-carts/*/estimate-shipping-methods`

vitaliyreznikov: Good catch Leon! I fixed this by changing "POST" to "GET" in prefetchShipping function. We no need to invalidate any sections when we do this prefetch call.

And this is an interesting idea to handle prefetchShipping using sections down the road, let me think about it.

vitaliyreznikov:
@mccottry Are we ok to merge this one?

view/frontend/templates/js/replacejs.phtml
mccottry: sounds good.

PR #756, opened 2020-06-17
Send non-Bolt tracking information to Bolt backend

nickfiacco-bolt:
#changelog Send non-Bolt tracking information to Bolt backend
Send non-Bolt tracking information to Bolt backend

### DIFF
--- a/Observer/TrackingSaveObserver.php
+++ b/Observer/TrackingSaveObserver.php
@@ -112,12 +112,12 @@ class TrackingSaveObserver implements ObserverInterface
             $order = $shipment->getOrder();
             $payment = $order->getPayment();
 
-            // If this is not a bolt payment, ignore it
-            if (!$payment || $payment->getMethod() != \Bolt\Boltpay\Model\Payment::METHOD_CODE) {
-                return;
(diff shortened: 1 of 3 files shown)


Observer/TrackingSaveObserver.php
```diff
@@ -112,13 +112,13 @@ public function execute(Observer $observer)
             $order = $shipment->getOrder();
             $payment = $order->getPayment();
 
-            // If this is not a bolt payment, ignore it
-            if (!$payment || $payment->getMethod() != \Bolt\Boltpay\Model\Payment::METHOD_CODE) {
-                return;
+            $isNonBoltOrder = !$payment || $payment->getMethod() != \Bolt\Boltpay\Model\Payment::METHOD_CODE;
+            if ($isNonBoltOrder) {
+                $transactionReference = $payment->getAdditionalInformation('bolt_transaction_reference');
```
vitaliyreznikov, line 117: Do you think it's possible that we don't have bolt_transaction_reference? For example, order was created before we updated plugin to the last version or bolt was temporary disabled.

nickfiacco-bolt: That’s a good point, I’ll handle this case

PR #758, opened 2020-06-19
M2P-49 Fix bug - Unable to complete order with virtual item after applying a discount

dumega:
# Description
When trying to apply coupon to virtual qoute on Bolt modal, it would fail, cause the billing address of $immutableQuote is always empty, and the cart data returned from Cart Helper function is empty as a result, this PR is to fix this bug.

Fixes: https://boltpay.atlassian.net/browse/M2P-49

#changelog Fix bug - Unable to complete order with virtual item after applying a discount

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -1304,7 +1304,7 @@ class Cart extends AbstractHelper
      * @return array
      * @throws \Exception
      */
-    public function getCartData($paymentOnly, $placeOrderPayload, $immutableQuote = null)
+    public function getCartData($paymentOnly, $placeOrderPayload, $immutableQuote = null, $requireBillingAddress = true)
     {
 
         // If the immutable quote is passed (i.e. discount code validation, bugsnag report generation)
@@ -1368,7 +1368,7 @@ class Cart extends AbstractHelper
         }
         $immutableQuote->collectTotals();
 
-        $cart = $this->buildCartFromQuote($quote, $immutableQuote, $items, $placeOrderPayload, $paymentOnly);
+        $cart = $this->buildCartFromQuote($quote, $immutableQuote, $items, $placeOrderPayload, $paymentOnly, $requireBillingAddress);
         if ($cart == []) {
             // empty cart implies there was an error
             return $cart;
--- a/Model/Api/DiscountCodeValidation.php
+++ b/Model/Api/DiscountCodeValidation.php
@@ -404,26 +404,16 @@ class DiscountCodeValidation implements DiscountCodeValidationInterface
             }
 
             // Set the shipment if request payload has that info.
-            if (isset($request->cart->shipments[0]->reference)) {
+            if ( $immutableQuote->isVirtual() && isset($request->cart->billing_address) ) {
+                $billingAddress = $immutableQuote->getBillingAddress();
+                $addressData = $this->prepareAddressData($request->cart->billing_address);
+                
+                $billingAddress->setShouldIgnoreValidation(true)
+                               ->addData($addressData)
+                               ->save();
+            } elseif ( isset($request->cart->shipments[0]->reference) ) {
                 $shippingAddress = $immutableQuote->getShippingAddress();
-                $address = $request->cart->shipments[0]->shipping_address;
-                $address = $this->cartHelper->handleSpecialAddressCases($address);
-                $region = $this->regionModel->loadByName(@$address->region, @$address->country_code);
-                $addressData = [
-                            'firstname'    => @$address->first_name,
-                            'lastname'     => @$address->last_name,
-                            'street'       => trim(@$address->street_address1 . "\n" . @$address->street_address2),
-                            'city'         => @$address->locality,
-                            'country_id'   => @$address->country_code,
-                            'region'       => @$address->region,
-                            'postcode'     => @$address->postal_code,
-                            'telephone'    => @$address->phone_number,
-                            'region_id'    => $region ? $region->getId() : null,
-                            'company'      => @$address->company,
-                        ];
-                if ($this->cartHelper->validateEmail(@$address->email_address)) {
-                    $addressData['email'] = $address->email_address;
-                }
+                $addressData = $this->prepareAddressData($request->cart->shipments[0]->shipping_address);
 
                 $shippingAddress->setShouldIgnoreValidation(true);
                 $shippingAddress->addData($addressData);
@@ -800,7 +790,7 @@ class DiscountCodeValidation implements DiscountCodeValidationInterface
     {
         $request = $this->getRequestContent();
         $is_has_shipment = isset($request->cart->shipments[0]->reference);
-        $cart = $this->cartHelper->getCartData($is_has_shipment, null, $quote);
+        $cart = $this->cartHelper->getCartData($is_has_shipment, null, $quote, false);
         return [
             'total_amount' => $cart['total_amount'],
             'tax_amount'   => $cart['tax_amount'],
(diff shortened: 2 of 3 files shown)


Model/Api/DiscountCodeValidation.php
```diff
@@ -1024,4 +1014,32 @@ protected function getParentQuoteDiscountResult($couponCode, $coupon, $parentQuo
...
+     * @param object $address
+     *
+     * @return array
+     */
+    private function prepareAddressData($address)
+    {
+        $address = $this->cartHelper->handleSpecialAddressCases($address);
+        $region = $this->regionModel->loadByName(@$address->region, @$address->country_code);
+        $addressData = [
+                    'firstname'    => @$address->first_name,
```
vitaliyreznikov, line 1028: @matt-thomason Do I understand correctly we want to stop use "@" ? Do we have any doc about this?

Model/Api/DiscountCodeValidation.php
```diff
@@ -1024,4 +1014,32 @@ protected function getParentQuoteDiscountResult($couponCode, $coupon, $parentQuo
             'discount_type'   => $this->convertToBoltDiscountType($rule->getSimpleAction()),
         ];
     }
+    
+    /**
+     * @param object $address
+     *
+     * @return array
+     */
+    private function prepareAddressData($address)
```
vitaliyreznikov, line 1023: @dumega Could you add a comment to this method?

Helper/Cart.php
```diff
@@ -1304,7 +1304,7 @@ protected function getCalculationAddress($quote)
      * @return array
      * @throws \Exception
      */
-    public function getCartData($paymentOnly, $placeOrderPayload, $immutableQuote = null)
+    public function getCartData($paymentOnly, $placeOrderPayload, $immutableQuote = null, $requireBillingAddress = true)
```
vitaliyreznikov, line 1307: Can you update phpdoc?

danac-gs, line 1307: @dumega Taking into account the complexity of the problem and it's root case on Bolt end - not calling shipping and tax for virtual carts and inconsistency in processing guest vs. logged in user - discussed in our private slack channel, and having in mind that this PR only solves an edge case, leaving the main problem open, I'd like to suggest one simplification here that you may or may not consider.

From Bolt perspective there's only one address, `shipping == billing`. Although billing address is not present in every request type data, it is the same. So, I'll remove logic around `$requireBillingAddress` (everywhere) and use shipping address as billing one everywhere billing is missing. I've checked how `$requireBillingAddress` is used in the code and it does not make much sense.

Model/Api/DiscountCodeValidation.php
```diff
@@ -800,7 +790,7 @@ private function getCartTotals($quote)
     {
         $request = $this->getRequestContent();
         $is_has_shipment = isset($request->cart->shipments[0]->reference);
```
danac-gs, line 792: Note: shipments is populated for the user logged in Bolt checkout (shipping and tax is called every time in the process of generating checkout data) and also for the guest users after they apply the discount in the modal (shipping and tax gets called by Bolt as a side effect of discount code validation). No matter if the cart is virtual or not. However shipping and tax endpoint is not called for guest user virtual carts in general (no coupon applied). This is inconsistency in Bolt functionality that causes issues and should be fixed on Bolt end.

dumega:
@danac-gs I think it is a plugin issue, not related to Bolt side, also it is not a edge case, but a common issue inside the plugin.

Let me show you an example, https://m2.nvip.eu/index.php/test-virtual.html this is a virtual product in my dev site, and for debugging, it can trigger the shipping&tax endpoint (the shipping and tax endpoint can be called for both guest user and logged-in user), so in the "Payment" step when we apply coupon on Bolt modal, the api request data of discounts.code.apply contains full billing address and shipments info. In this way, I think we can simulate the proper implement of applying coupon via Bolt modal.

And as we can see, the api hook still fails due to error that the billing address is empty (https://github.com/BoltApp/bolt-magento2/blob/master/Helper/Cart.php#L1478). So my point is, even Bolt fixes shipping and tax endpoint for virtual cart of guest user, this issue would still exist, that's why I says this is a common plugin issue. It is totally a different issue from the one  we discuss on slack (https://boltpay.slack.com/archives/CEEAATB2Q/p1592880744154800)

Although there's actually only one address in Bolt side, but the api hook send billing address and shipping info separately, and in the plugin side, we should always use billding address for checkout, that seems to be more logical in M2, so I think it is reasonable to check $requireBillingAddress for virtual cart.

danac-gs:
@dumega The "virtual" product, https://m2.nvip.eu/index.php/test-virtual.html,  does not seem to be configured correctly. The type of the item is "unknown". Should be "digital".
Moreover, I think using shipping address as a fallback for billing one (only if billing is missing, of course) and removing `$requireBillingAddress` is more clear and more logical solution. First, we should prefer removing code instead of adding it, thus keeping the overall complexity lower if possible. Then, it complies with Bolt logic of having one address only. At least one should always be present in the request from Bolt. If there's no address (no billing nor shipping) then and error should happen.

dumega:
@danac-gs The misconfigured product is intended, cause only using it, we can properly reproduce the issue with Bolt PPC, cause due to another issue on Bolt side (https://boltpay.atlassian.net/browse/EN-238), the shipping&tax endpoint wont be triggered for virtual cart.  And when applying coupon in Payment step of Bolt modal, the cart can be treated as virtual cart in M2 normally.

Well, my proposal is to wait until the virtual cart tax issue is fixed on Bolt side, then we can look into this issue again.

vitaliyreznikov:
@dumega Why we want to postpone this until the virtual item tax issue is resolved?
Let's say use applies a discount on the first step of guest checkout. We still have the issue in this case, correct?

dumega:
@vitaliyreznikov Yes, in my opinion, this is a plugin issue, whatever step you apply the discount, the error would happen. But @danac-gs  insist that the root cause is inconsistency in Bolt functionality and should be fixed on Bolt end. Then we have to wait until the virtual item tax issue is resolved and see if the issue still exist or not.

danac-gs:
> @vitaliyreznikov Yes, in my opinion, this is a plugin issue, whatever step you apply the discount, the error would happen. But @danac-gs insist that the root cause is inconsistency in Bolt functionality and should be fixed on Bolt end. Then we have to wait until the virtual item tax issue is resolved and see if the issue still exist or not.

@vitaliyreznikov @dumega Virtual carts do not work. Yet. It's a known issue and it's on Bolt end. They might work in some really edge cases, for instance when the discount is applied at some (last?) step due to the circumstance that shipping and tax is also called in the process (at the end of) discount validation step. There's no point in fixing this now and making plugin work with virtual carts in this edge case only, before the main issue is fixed. After it's fixed, this probably won'e be an issue anymore.

vitaliyreznikov:
@danac-gs What if we have a cart with virtual cart and apply a discount on the first step:
1) do we have an issue now?
2) can it be fixed on bolt side?

danac-gs:
> @danac-gs What if we have a cart with virtual cart and apply a discount on the first step:
> 
> 1. do we have an issue now?
> 2. can it be fixed on bolt side?

@vitaliyreznikov Just to be sure we are all on the same page. All the above discussion refers only to virtual carts, a cart with ALL items virtual. @dumega to confirm
I don't think the aforementioned case, or any other case, including the one from the task, is relevant until the `virtual cart collect tax` issue is fixed on Bolt side.
Back to your question. With the update / fix on Bolt side, all the carts, virtual or physical, would (should) behave the same, with no issue of this type.
My point, again, is that we can not consider and present (to merchants) this as a solution because it is a fix for only a small portion of virtual carts, the ones that have discounts applied (at some or any checkout step, does't matter). On the other hand, this would be no issue with Bolt side fix applied. It is misleading for everyone - engineers, engineer management, customer support, merchants.

vitaliyreznikov:
Let's close this for now, taking in account @danac-gs opinion

dumega:
@vitaliyreznikov When does the team fix the virtual cart collect tax issue on Bolt side?

vitaliyreznikov:
**Reopening:**
By the contract, we shouldn’t update cart (quote) on shipping&tax call. We need just provide information.
For example, for now we are working on static shipping project, when we miss shipping call because we know shipping price.
So long-term solution: don’t save billing address in quote
short-term: make sure that we are ok even if billing address is empty

adentong:
This has been open for half year now... I'm just going to close this.

PR #759, opened 2020-06-19
M2P-105 Fix issue with magento sections

vitaliyreznikov:
I added a bug in version 2.8.0.
I call customerData.invalidate with string argument instead of array.
This is the same call with argument ["b","o","l" ... ]

As result, magento tries to get a new value for section "b". Backend returns error and no sections updated.

**Who potentially affected:**
* Merchant on 2.8.0 or 2.8.1
* User did the purchase
* Use tries to pay with new cart
* User still has the old cookies at this moment  (cookies save for one day)
* User doesn't only add the product to cart but apply a discount or something similar (when user add a product to cart all magento sections are invalidated and magento doesn't look to list for invalidation, so this is a change to get correct bolt cart)

**Result**
The plugin doesn't receive a  correct bolt cart and call configure with cart {}. We expected to see an error on bolt modal but for some reason checkout is possible - with empty cart. This another bug (but not important one). I created task for them - https://boltpay.atlassian.net/browse/EN-268

**Who is really affected**
Although possible some users have this issue and sections don't updated for them it should not be a real issue: in version 2.8.0 and 2.8.1 we don't use magento sections for main checkout process. This will be fixed in 2.9.0

**Solution**
1. Fix the bug
2. Check if user has bad cookies and delete them - just in case. We should delete this check once we don't have any merchant on 2.8.0 or 2.8.1 I added asana task to check this https://app.asana.com/0/946297550934606/1181184977634457

Fixes: (link Jira ticket)

#changelog M2P-105 Fix issue with magento sections

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -483,7 +483,7 @@ if (!$block->isSaveCartInSections()) { ?>
                      * return void
                      */
                     var processSuccess = function (data) {
-                        customerData.invalidate('bolthints')
+                        customerData.invalidate(['bolthints']);
                         try {
                             if (typeof data !== 'undefined') {
                                 callbacks.success_url = data.success_url;
@@ -1270,6 +1270,13 @@ if (!$block->isSaveCartInSections()) { ?>
         'mage/cookies',
         'domReady!'
     ], function ($, authenticationPopup, customerData) {
+        var sectionDataIds = $.cookieStorage.get('section_data_ids');
+        if (sectionDataIds && sectionDataIds.b  !== undefined) {
+            $.cookieStorage.set('section_data_ids','');
+            storageInvalidation = $.initNamespaceStorage('mage-cache-storage-section-invalidation').localStorage;
+            storageInvalidation.removeAll();
+        }
+
         ////////////////////////////////////////////////////
 
         // initialize customer and cart observables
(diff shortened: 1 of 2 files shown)


view/frontend/templates/js/replacejs.phtml
```diff
@@ -1270,6 +1270,13 @@ if (!$block->isSaveCartInSections()) { ?>
         'mage/cookies',
         'domReady!'
     ], function ($, authenticationPopup, customerData) {
+        var sectionDataIds = $.cookieStorage.get('section_data_ids');
+        if (sectionDataIds && sectionDataIds.b  !== undefined) {
+            $.cookieStorage.set('section_data_ids','');
+            storageInvalidation = $.initNamespaceStorage('mage-cache-storage-section-invalidation').localStorage;
+            storageInvalidation.removeAll();
+        }
+
```
danac-gs, line 1279: @vitaliyreznikov I was curious what happens when you don't include this part. Is it really needed after the store is updated and caches cleared? Full page cache included. Did not get a chance to test. Did you?

vitaliyreznikov: Yes, I tested very carefully. Yes, we need it if caches cleared on server side, but user has bad cookies.
Few users can be affected, only who made purchase and continue actively using site (cookies are deleted after 1 day of inactivity), but I offered to include this check just in case.

PR #764, opened 2020-06-23
M2P-104 Call configure() with promises when we know cart will be updating soon

vitaliyreznikov:
If user updated carts (add discount / change quanity / remove item) and click bolt checkout right after that user can see checkout with old cart.
It happens because of magento didn't update cart yet.
In this PR we are fixing it by calling configure() with promises in all cases when we know cart will be updated in a next few seconds.

If user clicks bolt button before promises are resolved user just waits for resolve. 

Fixes: [M2P-104]

#changelog M2P-104 Call configure() with promises when we know cart will be updating soon

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -1878,7 +1878,6 @@ if (!$block->isSaveCartInSections()) { ?>
                         if (waitingForResolvingPromises && !boltConfig.always_present_checkout) {
                             cartBarrier.resolve(cart);
                             hintBarrier.resolve(hints);
-                            waitingForResolvingPromises = false;
                         } else {
                             if (boltConfig.always_present_checkout) {
                                 boltCheckoutConfig = (cart.orderToken && cart.orderToken !== "") ? {floatingButtonMode: newItemAddedToCart ? "show_cart" : "show_cart_on_hover" } : {};
@@ -1887,6 +1886,7 @@ if (!$block->isSaveCartInSections()) { ?>
                             // reconfigure bolt checkout with new values
                             BC = BoltCheckout.configure(cart, hints, callbacks, boltCheckoutConfig);
                         }
+                        waitingForResolvingPromises = false;
                         oldBoltCartValue = JSON.stringify(cart);
 
                         // open the checkout if auto-open flag is set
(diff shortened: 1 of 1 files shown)


vitaliyreznikov:
Hey Leon. Three points where we don't see eye to eye:

> adding over 1000 lines of code
We added lines only for safety reasons to have the ability to use old code. We remove them right after we launch release

>you've reported that the speed of the code has been reduced by over 2x. 
I haven't reported that speed of the code has been reduced by 2x.
Speed was increased in most cases because we call bolt server much less often.
In one specific case speed was reduced

> Additionally, the code complexity is increasing.
Now we have one entry point - magento section updates - so code became to be more logical.

view/frontend/templates/js/replacejs.phtml
```diff
@@ -1695,9 +1695,37 @@ if (!$block->isSaveCartInSections()) { ?>
...
+        var callConfigureWithPromises = function() {
+            cartBarrier = boltBarrier();
+            hintBarrier = boltBarrier();
+            BC = BoltCheckout.configure(cartBarrier.promise, hintBarrier.promise, callbacks);
+            waitingForResolvingPromises = true;
+            countConfigureCalls++;
+            var localCallNumber = countConfigureCalls;
+            // invalidate bolt cart if we didn't resolve promises for some reasons
+            setTimeout(function () {
+                if (waitingForResolvingPromises && (localCallNumber === countConfigureCalls)) {
```
njfaries, line 1714: Not understanding the exact purpose of the second comparison here. @vitaliyreznikov can you explain?

matt-thomason:
@mccottry I agree regarding the speed and aloofness we have packed things into this release. After discussions in slack though, we've decided to try and roll forward with this particular approach: https://boltpay.slack.com/archives/CQ864R9JQ/p1592925633397900

We've been discussing in parallel over slack and in other areas - after we figure out this release we'll run a post-mortem and make sure we implement controls to ensure situations like this don't happen in the future.

view/frontend/templates/js/replacejs.phtml
vitaliyreznikov: If, for example we call this function twice with an interval of 3 seconds we don't want the first call invalidates cart because the second call still works and we are waiting for the result.

PR #766, opened 2020-06-25
M2P-116 Update bolt cart when magento cart is updated

vitaliyreznikov:
# Description
After recent refactoring, we get bolt cart from server in the same request as Magento cart.
This led to the Magento interface works a little slowly: it takes a couple of additional seconds between user clicks "add to cart" and Magento shows updates is finished.

By this PR we trigger bolt cart updating when we get a new Magento cart from server, so we stop affecting to M2 interface.
It will be almost the same for users: after recent refactoring, we call configures() with promises pretty much after user clicks add to cart button (or opposite remove something).
So if a user clicks the bolt button when we haven't received a new cart yet they just wait additional couple of seconds and see the correct cart.

Fixes: [M2P-116]

#changelog M2P-116 Update bolt cart when magento cart is updated

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:


view/frontend/templates/js/replacejs.phtml
```diff
@@ -1714,16 +1714,50 @@ if (!$block->isSaveCartInSections()) { ?>
...
+            if (!waitingForResolvingPromises) {
+                callConfigureWithPromises();
+            }
+        }
+
+        var magentoCart;
+        var boltCartDataID;
+
+        var invalidateBoltCartIfOutdated = function() {
+            setTimeout(function () {
```
vitaliyreznikov, line 1731: We need to run this code after all code from the JS stack executed. Is timeout 1 is the proper way to do it? @daisy1754

view/frontend/templates/js/replacejs.phtml
```diff
@@ -1714,16 +1717,48 @@ if (!$block->isSaveCartInSections()) { ?>
                 if (waitingForResolvingPromises && (localCallNumber === countConfigureCalls)) {
                     customerData.reload(['boltcart'], true);
                 }
-            }, 5000);
+            }, 10000);
```
vitaliyreznikov, line 1720: On my [very slow] test site + with local bolt sometimes it takes 6 seconds. Anyway
- in the worst case [when user click add to cart and we check how many seconds need to update bolt cart] we have pretty much the same latency but with better UI (promises)
- in general case [user goes from one page to another] we win because we don't update bolt cart anymore, get it from local storage.
- we have this check just in case, it never should work

view/frontend/templates/js/replacejs.phtml
```diff
@@ -1714,16 +1717,48 @@ if (!$block->isSaveCartInSections()) { ?>
...
+        var invalidateBoltCartIfOutdated = function() {
+            setTimeout(function () {
+                if (magentoCart === undefined || magentoCart.data_id === undefined || boltCartDataID === undefined) {
+                    return;
+                }
+                if (magentoCart.data_id > boltCartDataID) {
+                    invalidateBoltCart();
+                    return;
+                }
+                if (magentoCart.data_id == boltCartDataID) {
```
vitaliyreznikov, line 1740: It doesn't look like a possible real case, when magento cart and bolt cart have the same timestamp but bolt cart generated before and it is outdated.
But just in case I decided to add this check.

vitaliyreznikov:
@danac-gs Function `a()` calls `b()` and then `c()`. I need to call `d()` from `b()` but want it to be executed after `c()`.
Do you know how to do this without setTimeout? I believe shoud be more native way.

danac-gs:
> @danac-gs Function `a()` calls `b()` and then `c()`. I need to call `d()` from `b()` but want it to be executed after `c()`.
> Do you know how to do this without setTimeout? I believe shoud be more native way.

You can try to set a control variable at the end of c() - option 2, or as a result of c() - option 1, and then use whenDefined in b() to call d() when the control variable is set. Something like this:
```javascript
function a() {
    b();
    // option 1
    window.controlVariable = c();
}

function b() {
    ...
    whenDefined(window, 'controlVariable', d)
}

function c() {
    ...
    // option 2
    window.controlVariable = c();
}

function d() {
}
```

vitaliyreznikov:
I'm sorry Daniel. I didn't explain all flow.
When magento2 receive new sections from server it calls all subscribers
We have two
customerData.get('cart').subscribe <---  b
customerData.get('boltcart').subscribe <--- c

we need call d() after b() and c() but we don't know if a() calls only b() or only c() or both b() and c()

danac-gs:
Looks like a design problem. We should observe only Magento 'cart' changes and update cart on Bolt end accordingly. Something that should have simplified things turned out to complicate them quite a bit. I don't have the answer for your last question off the top of my head. Perhaps you can use more than one control variable and nest whenDefined. If that makes sense.

vitaliyreznikov:
I think this is an important case. If all data become outdated magento received cart and bolt_cart in one request. If we don't recognize this situation, we updated bolt_cart one more time without needs.
setTimeout works fine, but I'll try to find another solution tomorrow.

view/frontend/templates/js/replacejs.phtml
daisy1754: what do you mean by "all code from the JS stack"?

vitaliyreznikov:
Thank you Kazuki and Daniel.
Looks like setTimeout with zero delay is a valid way for what we need: add our code at the end of JS **event loop** (I mistakenly **called** it stack yesterday).
Some links:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop
https://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-sometimes-useful

added some changes:
1) change delay from 1 to 0
2) implemented Daniel's proposal to listen to only magento cart updates
2) added comment.

Please review @danac-gs

view/frontend/templates/js/replacejs.phtml
vitaliyreznikov: I needed to say **event loop**. We need to add code in the end of JS event loop

PR #771, opened 2020-06-28
M2P-118 Clone items of source quote to destination quote properly

dumega:
# Description
Now when we clone quote data from source to destination,  we use the method `merge` of Quote model to clone the the items and some other data. In this `merge` method, it mainly does three things :
1. Clone the items
2. Init shipping and billing address if quote is new
3. Set coupon code if exists

After investigation, I found that:
1. Clone the items
The method `merge` of Quote model would compare quote items with the product id and options, if these property data are same then it just merge the two items. But we face cases that the cart could be created by ERP, and the items of cart could have exactly same product id and options, only the unit price are different, M2 still treats them as different quote items and have unique quote item ids.  When we replicate quote data, what we need to do is 1:1 clone, obviously the method `merge` can not implement the clone in the proper way, the quote items get merged, not cloned.

2. Init shipping and billing address if quote is new
It seems to be duplicate, cause we have specific code to do that
```
$destination->getBillingAddress()->setShouldIgnoreValidation(true);
$this->transferData($source->getBillingAddress(), $destination->getBillingAddress());

$destination->getShippingAddress()->setShouldIgnoreValidation(true);
$this->transferData($source->getShippingAddress(), $destination->getShippingAddress());
```

3. Set coupon code if exists
It also seems to be duplicate, cause we have specific code to do that
```
$this->transferData($source, $destination, false);
```

So to solve the issue of cloning quote items, we need to remove the method `merge` and do 1:1 clone on items.

Fixes: https://boltpay.atlassian.net/browse/M2P-118

#changelog Clone items of source quote to destination quote properly

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -1015,7 +1015,17 @@ class Cart extends AbstractHelper
 
         $destination->removeAllItems();
 
-        $destination->merge($source);
+        foreach ($source->getAllVisibleItems() as $item) {
+            $newItem = clone $item;
+            $destination->addItem($newItem);
+            if ($item->getHasChildren()) {
+                foreach ($item->getChildren() as $child) {
+                    $newChild = clone $child;
+                    $newChild->setParentItem($newItem);
+                    $destination->addItem($newChild);
+                }
+            }
+        }
 
         $destination->getBillingAddress()->setShouldIgnoreValidation(true);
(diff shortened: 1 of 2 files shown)


dumega:
@danac-gs Yes, I tested with coupons and gift cards

PR #795, opened 2020-07-15
M2P-147 Implement a feature switch to toggle the logic to pre-fill address from Bolt for the logged-in customer

johnmccombs1:
# Description
There is a merchant that doesn’t want to use the logic pre-fill address from Bolt for the logged-in customer. 
So I created this PR to implement a feature switch to toggle that logic

Fixes: 
https://boltpay.atlassian.net/browse/M2P-147
https://boltpay.atlassian.net/browse/EN-395

#changelog M2P-147 Implement a feature switch to toggle the logic to pre-fill address from Bolt for the logged-in customer

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -60,6 +60,7 @@ use Magento\Customer\Api\CustomerRepositoryInterface as CustomerRepository;
 use Bolt\Boltpay\Exception\BoltException;
 use Bolt\Boltpay\Model\ErrorResponse as BoltErrorResponse;
 use Bolt\Boltpay\Helper\MetricsClient;
+use Bolt\Boltpay\Helper\FeatureSwitch\Decider as DeciderHelper;
 
 /**
  * Boltpay Cart helper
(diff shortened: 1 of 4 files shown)


johnmccombs1:
@danac-gs Yes, that is correct. If the feature switch is enabled, it would prevent logging in the existing Bolt accounts and force every customer to confirm their address first like this http://prntscr.com/t4ipjg

njfaries:
@johnmccombs1 isn't it the other way? As in, it looks like if the switch is enabled prefilling is able to happen, and if disabled then prefilling is stopped. So this switch would be on by default except this merchant (and others with similar needs) could disable? Just want to be clear because that's how it looks and what the naming seems to imply.

vitaliyreznikov:
`isPrefillAddressFromBoltForLoggedInCustomerEnabled` looks like we can enable or disable this as we want.
I think we need to always have this functional enable and only if we really need - disable it.
Please consider renaming feature switch into negative name, like
`ifShouldDisablePrefillAddressForLoggedInCustomer`
This feature switch can be disabled by default (rollout 0%) and we can enable it for the specific merchant.

johnmccombs1:
@njfaries  I've just updated the method name so it looks more specific about the feature. By default, the feature switch is off and we allow the prefilling address for the logged-in customer. When we enable the feature switch, it won't allow the prefilling address for the logged-in customer.


@vitaliyreznikov  I've updated the code. Thanks for the suggestion

PR #796, opened 2020-07-15
M2P-140 Listen if 3p modules update quote total

vitaliyreznikov:
If a third party plugin doesn't update Magento cart after discount applying we don't update bolt cart as well.
Solution: invalidate bolt cart if quote total was updated and we don't know a reason.
We don't want to invalidate bolt date if quote total updated when page is loaded, or after on of magento event (add to cart, etc)

Fixes: [M2P-140]

#changelog Fix an issue with some third party discounts modules

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -2292,14 +2292,14 @@ if (!$block->isSaveCartInSections()) { ?>
         'Magento_Checkout/js/model/quote',
         'domReady!'
         ], function ($, customerData, quote) {
+            // If quote total was changed through a method we didn't create,
+            // we should reload the bolt cart since totals are probably now out of sync.
             quote.totals.subscribe(function (newValue) {
                 if (expectCartRendering) {
                     expectCartRendering = false;
                     return;
                 }
 
-                // If we don't expect cart rendering possible it caused by any third party plugin (for example, 3p coupon is applied)
-                // Bolt cart can be changes, it worth to invalidate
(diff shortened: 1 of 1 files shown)


view/frontend/templates/js/replacejs.phtml
```diff
@@ -2279,6 +2284,30 @@ if (!$block->isSaveCartInSections()) { ?>
...
+        'Magento_Checkout/js/model/quote',
+        'domReady!'
+        ], function ($, customerData, quote) {
+            quote.totals.subscribe(function (newValue) {
+                if (expectCartRendering) {
+                    expectCartRendering = false;
+                    return;
+                }
+
+                // If we don't expect cart rendering possible it caused by any third party plugin (for example, 3p coupon is applied)
```
matt-thomason, line 2301: Can we bubble up this comment to the top of this function and explain a bit more in detail what the whole thing is doing, something like "If quote total was changed through a method we didn't create, we should reload the bolt cart since totals are probably now out of sync"

vitaliyreznikov:
> Do we need to do this for adminhtml as well?
I think not. In the legacy workflow, we listen to cart total changes in DOM and make a new AJAX call if cart total is changed.
DOM change is the next step after the Magento event that I propose to listen to in this PR.

view/frontend/templates/js/replacejs.phtml
vitaliyreznikov: Done

PR #798, opened 2020-07-16
M2P-137 General plugin support for Amasty GiftCard 2.0.0 or later

dumega:
# Description
Since v2.0.0, the Amasty team completely overhauled Amasty GiftCard module, as a result, Bolt M2 plugin does not work with the latest version of Amasty GiftCard. So we need to update the related code to support.

Fixes: https://boltpay.atlassian.net/browse/M2P-137

#changelog General plugin support for Amasty GiftCard 2.0.0 or later

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Model/ThirdPartyModuleFactory.php
+++ b/Model/ThirdPartyModuleFactory.php
@@ -77,7 +77,17 @@ class ThirdPartyModuleFactory
      */
     public function isExists()
     {
-        return class_exists($this->className);
+        ///////////////////////////////////////////////////////////////
+        // Due to a known bug https://github.com/magento/magento2/pull/21435,
+        // the autoloader throws an exception on class_exists.
+        // Add try-catch block to avoid uncaught exceptions during autoloading.
+        // Return false instead if any uncaught exceptions.
+        ///////////////////////////////////////////////////////////////
+        try {
+            return class_exists($this->className);
+        } catch (\Exception $e) {
+            return false;
+        }
     }
 
     /**
--- a/Model/ThirdPartyModuleFactory.php
+++ b/Model/ThirdPartyModuleFactory.php
@@ -70,6 +70,15 @@ class ThirdPartyModuleFactory
     {
         return $this->_moduleManager->isEnabled($this->moduleName);
     }
+    
+    /**
+     * Check whether the class exists
+     * @return bool
+     */
+    public function isExists()
+    {
+        return class_exists($this->className);
+    }
 
     /**
      * @param array $data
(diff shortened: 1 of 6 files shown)


vitaliyreznikov:
@matt-thomason In this PR @dumega added two types of support for different versions of a plugin, legacy, and new flow.
I think it's important to document it: which version needs legacy flow and which a new flow.

Down the road we can remove legacy flow if we know that all merchants updated the plugin.

What's the best way to document? I think we can add comments into etc/di.xml or create a new file in the repo with all plugins and versions.

matt-thomason:
@vitaliyreznikov This is a good point. I think comments in `di.xml` are a good place for us, and then we should also probably document this in Confluence somewhere - maybe a top-level page in Merchant Experience > Magento 2 page.

dumega:
@matt-thomason @vitaliyreznikov I have added comments about version info in the di.xml

danac-gs:
@dumega was this solution thoroughly tested with the old Amasty module?

dumega:
@danac-gs Yes, Tested with v1.18.15 and v2.1.3

dumega:
@vitaliyreznikov Do you have Amasty GiftCard installed properly in your dev site? It looks like you have files from latest version of Amasty GiftCard on the server but old version number / module name in config.

I do not see any similar error during the testing, or when switch between different versions.

danac-gs:
> @vitaliyreznikov Do you have Amasty GiftCard installed properly in your dev site? It looks like you have files from latest version of Amasty GiftCard on the server but old version number / module name in config.
> 
> I do not see any similar error during the testing, or when switch between different versions.

@dumega @vitaliyreznikov the plugin installation / compilation (code generation) with the new code addition should work without the Amasty GiftCard module installed (old or new). Please verify or confirm you already did.

dumega:
@danac-gs @vitaliyreznikov I setup a fresh M2 site and install this specific branch (so there is not Amasty GiftCard installed), everything works fine

dumega:
@vitaliyreznikov I see, I tested with M2 2.3.5-p1, that's why it does not happen to me. Well, I will manage to fix it.

dumega:
@vitaliyreznikov Fixed.

Model/ThirdPartyModuleFactory.php
```diff
@@ -77,7 +77,17 @@ public function isAvailable()
      */
     public function isExists()
```
danac-gs, line 78: @dumega please remind me why do we need isExists.

dumega: Since the Amasty team completely overhauled Amasty GiftCard module, module is enabled does not mean the specific class exists, instead of comparing the version, I think checking the class existence is more reliable.

PR #806, opened 2020-07-21
M2P-155 Bugfix - Unable to remove Amasty GiftCard

dumega:
# Description
```
Steps to reproduce: 

1. Purchase a gift card (generated by Amasty GiftCard)
2. Add a simple product to cart
3. Go to cart page
4. Open Bolt modal and apply the gift card in the "Delivery" or "Payment" step
5. Keep the Bolt modal open and directly reload the cart page
6. Delete the applied gift card by clicking “Remove” button of Amasty GiftCard 
7. Open the Bolt modal

Expected behavior:
The Bolt modal does not have any applied gift card

Actual behavior:
The Bolt modal still has the previous gift card
```

Root cause:
When loading the cart page, we save the current quote without giftcard into cache, also an associated cart is created on Bolt side. Then we apply giftcard through Bolt modal, the Bolt cart get updated while the cache stay the same. So when we remove the giftcard from cart page, the plugin just retrieve the cache and wont refresh the cart on Bolt side, it result in a mismatch.

Fixes: https://boltpay.atlassian.net/browse/M2P-155

#changelog Bugfix - Unable to remove Amasty GiftCard

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -487,6 +487,18 @@ class Cart extends AbstractHelper
         }
         return $unserialize ? unserialize($cached) : $cached;
     }
+    
+    /**
+     * Remove data from Magento cache
+     *
+     * @param string $identifier
+     */
+    public function removeFromCache($identifier)
+    {
+        if (!empty($identifier)) {
+            $this->cache->remove($identifier);
+        }
+    }
 
     /**
      * Save data to Magento cache
@@ -2216,5 +2228,43 @@ class Cart extends AbstractHelper
         } finally {
             return $result;
         }
+    }
+    
+    /**
+     * Get the hash of the quote to be used as cache identifier
+     *
+     * @param bool   $paymentOnly              flag that represents the type of checkout
+     * @param string $placeOrderPayload        additional data collected from the (one page checkout) page,
+     *                                         i.e. billing address to be saved with the order
+     * @param Quote $quote
+     * @param null|int    $storeId             The ID of the Magento store
+     *
+     * @return string
+     */
+    public function getQuoteCacheIdentifier($paymentOnly, $placeOrderPayload, $quote, $storeId = null)
+    {
+        // If storeId was missed through request, then try to get it from the session quote.
+        if ($storeId === null) {
+            $storeId = $this->getSessionQuoteStoreId();
+        }
+
+        // Try fetching data from cache
+        if (!$this->isBoltOrderCachingEnabled($storeId)) {
+            return '';
+        }
+        
+        //Get cart data
+        $cart = $this->getCartData($paymentOnly, $placeOrderPayload, $quote);
+
+        if (!$cart) {
+            return '';
+        }
+
+        if ($this->doesOrderExist($cart, $quote)) {
+            $this->deactivateSessionQuote($quote);
+            return '';
+        }
+
+        return $this->getCartCacheIdentifier($cart);
     }
 }
--- a/Model/Api/DiscountCodeValidation.php
+++ b/Model/Api/DiscountCodeValidation.php
@@ -402,6 +402,8 @@ class DiscountCodeValidation implements DiscountCodeValidationInterface
 
                 return false;
             }
+            
+            $quoteCacheIdentifier = $this->getQuoteCacheIdentifier($immutableQuote, $storeId, $request);
 
             // Set the shipment if request payload has that info.
             if (isset($request->cart->shipments[0]->reference)) {
@@ -442,7 +444,7 @@ class DiscountCodeValidation implements DiscountCodeValidationInterface
                     $result = $this->applyingCouponCode($couponCode, $coupon, $immutableQuote, $parentQuote);
                 }
             } elseif ($giftCard && $giftCard->getId()) {
-                $result = $this->applyingGiftCardCode($couponCode, $giftCard, $immutableQuote, $parentQuote);
+                $result = $this->applyingGiftCardCode($couponCode, $giftCard, $immutableQuote, $parentQuote, $quoteCacheIdentifier);
             } else {
                 throw new WebApiException(__('Something happened with current code.'));
             }
@@ -682,7 +684,7 @@ class DiscountCodeValidation implements DiscountCodeValidationInterface
      * @return array
      * @throws \Exception
      */
-    private function applyingGiftCardCode($code, $giftCard, $immutableQuote, $parentQuote)
+    private function applyingGiftCardCode($code, $giftCard, $immutableQuote, $parentQuote, $quoteCacheIdentifier)
     {
         try {
             if ($giftCard instanceof \Amasty\GiftCard\Model\Account) {
@@ -692,7 +694,10 @@ class DiscountCodeValidation implements DiscountCodeValidationInterface
                 $this->discountHelper->removeAmastyGiftCard($giftCard->getCodeId(), $parentQuote);
                 // Apply Amasty Gift Card to the parent quote
                 $giftAmount = $this->discountHelper->applyAmastyGiftCard($code, $giftCard, $parentQuote);
-                // Reset and apply Amasty Gift Cards to the immutable quote
+                // Clear the quote data from Magento cache, so when the customer remove the gift card from cart page,
+                // it can force to refresh the cart in Bolt side.
+                $this->clearQuoteCache($quoteCacheIdentifier);
+                // Reset and apply Amasty Gift Cards to the immutable quote                
                 $this->discountHelper->cloneAmastyGiftCards($parentQuote->getId(), $immutableQuote->getId());
             } elseif ($giftCard instanceof \Unirgy\Giftcert\Model\Cert) {
                 /** @var \Unirgy\Giftcert\Helper\Data $unirgyHelper */
@@ -806,6 +811,31 @@ class DiscountCodeValidation implements DiscountCodeValidationInterface
             'tax_amount'   => $cart['tax_amount'],
             'discounts'    => $cart['discounts'],
         ];
+    }
+    
+    /**
+     * Remove data from Magento cache
+     *
+     * @param string $quoteCacheIdentifier
+     */
+    private function clearQuoteCache($quoteCacheIdentifier)
+    {
+        $this->cartHelper->removeFromCache($quoteCacheIdentifier);
+    }
+    
+    /**
+     * Get the hash of the quote to be used as cache identifier
+     *
+     * @param Quote       $quote
+     * @param null|int    $storeId
+     * @param Object      $request
+     *
+     * @return string
+     */
+    private function getQuoteCacheIdentifier($quote, $storeId, $request)
+    {
+        $is_has_shipment = isset($request->cart->shipments[0]->reference);
+        return $this->cartHelper->getQuoteCacheIdentifier($is_has_shipment, null, $quote, $storeId);
     }
 
     /**
--- a/Test/Unit/Helper/CartTest.php
+++ b/Test/Unit/Helper/CartTest.php
@@ -5780,4 +5780,128 @@ ORDER
 
         $this->assertEquals($expected, $currentMock->calculateCartAndHints());
         }
+        
+        /**
+        * @test
+        * that getQuoteCacheIdentifier always return empty string when caching is disabled
+        *
+        * @covers ::getQuoteCacheIdentifier
+        *
+        */
+        public function getQuoteCacheIdentifier_withCachingDisabled_returnEmptyString()
+        {
+            $this->currentMock = $this->getCurrentMock(
+                [
+                    'getSessionQuoteStoreId',
+                    'isBoltOrderCachingEnabled',
+                ]
+            );
+
+            $this->currentMock->expects(static::once())->method('getSessionQuoteStoreId')->willReturn(self::STORE_ID);
+            $this->currentMock->expects(static::once())->method('isBoltOrderCachingEnabled')->with(self::STORE_ID)
+                ->willReturn(false);
+
+            $result = $this->currentMock->getQuoteCacheIdentifier(false, '', $this->quoteMock);
+            static::assertEquals('', $result);
+        }
+
+        /**
+        * @test
+        * that getQuoteCacheIdentifier always return empty string if cart data is empty
+        *
+        * @covers ::getQuoteCacheIdentifier
+        *
+        */
+        public function getQuoteCacheIdentifier_withEmptyCartData_returnEmptyString()
+        {
+            $this->currentMock = $this->getCurrentMock(
+                [
+                    'getSessionQuoteStoreId',
+                    'isBoltOrderCachingEnabled',
+                    'getCartData',
+                ]
+            );
+
+            $this->currentMock->expects(static::once())->method('getSessionQuoteStoreId')->willReturn(self::STORE_ID);
+            $this->currentMock->expects(static::once())->method('isBoltOrderCachingEnabled')->with(self::STORE_ID)
+                ->willReturn(true);
+            $this->currentMock->expects(static::once())->method('getCartData')->with(false, '', $this->quoteMock)
+                ->willReturn([]);
+
+            $result = $this->currentMock->getQuoteCacheIdentifier(false, '', $this->quoteMock);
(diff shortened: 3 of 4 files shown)


dumega:
> @dumega Could we use a simpler approach - just adding something unique related to the Amasty GiftCard applied to quote (if existing, used gc balance for instance) to the cart cache identifier, checking it every time? Perhaps that way it would get solved by itself without the use of much more additional code.
> @vitaliyreznikov @matt-thomason @njfaries

@danac-gs The GiftCard is applied on Bolt modal, not on the cart page, so the cart cache identifier does not change for the same quote (cause the quote data does not change). So I am afraid your suggestion wont work.

danac-gs:
> @danac-gs The GiftCard is applied on Bolt modal, not on the cart page, so the cart cache identifier does not change for the same quote (cause the quote data does not change). So I am afraid your suggestion wont work.

@dumega 

4. Open Bolt modal and apply the gift card in the "Delivery" or "Payment" step
_Here a call to a store endpoint is made, and some (identifier, quote related / Amasty gc) data is changed and the change stored in DB at store end_
5. Keep the Bolt modal open and directly reload the cart page
_The new cache identifier is generated (taking into account the quote state change in the previous step) and Bolt cart is updated_
6. Delete the applied gift card by clicking “Remove” button of Amasty GiftCard
_I assume the problem is here. We don't get a notification on the front end that the cart (totals) had been changed and that create cart (BoltCheckout configure with a promise) should be called again_

**This looks like another side effect of using sections the way we do, and we are still fixing the consequences instead of dealing with the root cause (for a month or so already)**

dumega:
@danac-gs 

First lets check how we generate the cache identifier

https://github.com/BoltApp/bolt-magento2/blob/c039effea085bdeb0ba172bf7fa3bb4b0837244c/Helper/Cart.php#L581

So it always calculate the cache identifier by the cart content data, then even there are two quote with different id but same content, the cache identifier would be the same.

Then referring to https://github.com/BoltApp/bolt-magento2/blob/master/Helper/Cart.php#L725 , the plugin just return the cached Bolt order insteading calling Bolt Ceate Order API.


**4. Open Bolt modal and apply the gift card in the "Delivery" or "Payment" step**
_Here a call to a store endpoint is made, and some (identifier, quote related / Amasty gc) data is changed and the change stored in DB at store end_

Bolt order at Bolt server side also get updated cause discounts.code.apply returns discount info and new totals. **The problem is here**, the Bolt order saved in the cache at store end does not get updated, so there is mismatch.

**5. Keep the Bolt modal open and directly reload the cart page**
_The new cache identifier is generated (taking into account the quote state change in the previous step) and Bolt cart is updated_

Although a new cache identifier would be generated, the old cache identifier still exists, and the new cache identifier would associate with another Bolt order which works as expected.

**6. Delete the applied gift card by clicking “Remove” button of Amasty GiftCard**
_I assume the problem is here. We don't get a notification on the front end that the cart (totals) had been changed and that create cart (BoltCheckout configure with a promise) should be called again_

After removing GiftCard, the cart content data is as same as before, so it retrieve Bolt order from cache and never call the Bolt Ceate Order API.

danac-gs:
https://github.com/BoltApp/bolt-magento2/blob/c039effea085bdeb0ba172bf7fa3bb4b0837244c/Helper/Cart.php#L581-L596

This method can be extended to include Amasty GiftCard / quote related addition to the cache identifier.
As it was done for custom address fields and gift message.

dumega:
> https://github.com/BoltApp/bolt-magento2/blob/c039effea085bdeb0ba172bf7fa3bb4b0837244c/Helper/Cart.php#L581-L596
> 
> This method can be extended to include Amasty GiftCard / quote related addition to the cache identifier.
> As it was done for custom address fields and gift message.

@danac-gs There is nothing to do with cache identifier
1. Initially on the cart page we do not have GiftCard applied, so the cache identifier does not have any info with GiftCard.
2. On the Bolt modal, when we add GiftCard, the cache identifier is not updated (actually it should not get updated)
3. Then after removing the GiftCard on the cart page, the cart changes back to the previous state which has same cache identifier as before.

Besides that, if we apply the GiftCard on the cart page, the cart does have Amasty GiftCard info in the discount property, do you mean we need to add some more extra info?

dumega:
> @dumega If I want to test this with last plugin version I should cherry-pick this one as well, correct?
> #798

@vitaliyreznikov yes, if you want to test with Amastry GiftCard v2.0.0 and later, you need to have the PR #798 included.

danac-gs:
@dumega  I think we should do 2 things (only):

1. add extra info, quote / immutable quote / Amasty Gift Card related, to the cache identifier
2. identify and fix why the totals change (adding / removing gift cards) does not trigger cart update on the front end (shopping cart page, replacejs.phtml)

dumega:
@danac-gs 
For the two things you suggest,
`1.  add extra info, quote / immutable quote / Amasty Gift Card related, to the cache identifier`
Again, I believe it is irreverent to this issue. And we can consider to add those extra info in another PR

`2. identify and fix why the totals change (adding / removing gift cards) does not trigger cart update on the front end (shopping cart page, replacejs.phtml)`
In the front page, the ajax get triggered to get Bolt order when removing the GiftCard, but due to the existence of cache, it returns the cache instead of calling api. And this PR is mainly to fix this issue.

danac-gs:
> @danac-gs
> For the two things you suggest,
> `1. add extra info, quote / immutable quote / Amasty Gift Card related, to the cache identifier`
> Again, I believe it is irreverent to this issue. And we can consider to add those extra info in another PR
> 
> `2. identify and fix why the totals change (adding / removing gift cards) does not trigger cart update on the front end (shopping cart page, replacejs.phtml)`
> In the front page, the ajax get triggered to get Bolt order when removing the GiftCard, but due to the existence of cache, it returns the cache instead of calling api. And this PR is mainly to fix this issue.

@dumega As I see it, if we do 1., the cache identifier wont match in 2. and the api would be called, generating and caching a new cart, instead returning an old one from cache

dumega:
@danac-gs Well, the rule of the cache identifier is unique identifier for same cart, right? 
The two cart
1. The one before opening the Bolt modal (step 3. Go to cart page)
2. The one after removing the gift card (step 6. Delete the applied gift card by clicking “Remove” button of Amasty GiftCard )

Do you think they are same or not?  In my opinion, they are same, even we add more extra info, but they are still same cart. Then why/how do we return different cache identifier for same cart?

danac-gs:
> @danac-gs Well, the rule of the cache identifier is unique identifier for same cart, right?
> The two cart
> 
> 1. The one before opening the Bolt modal (step 3. Go to cart page)
> 2. The one after removing the gift card (step 6. Delete the applied gift card by clicking “Remove” button of Amasty GiftCard )
> 
> Do you think they are same or not? In my opinion, they are same, even we add more extra info, but they are still same cart. Then why/how do we return different cache identifier for same cart?

@dumega Currently they are treated the same because there's nothing to distinguish between them in the cache identifier. And they really are not the same, one has a gift cart applied and the other one does not. If we add some gift card related field to the cache identifier then they won't be treated as the same cart anymore and everything should work. I can't know if it 100% would, but it seems far simpler then this pull request so it is worth trying imho. No adding to or changing the existing plugin logic, just adding another part to the identifier (which, I admit, might be a challenge to get right).

dumega:
> > @danac-gs Well, the rule of the cache identifier is unique identifier for same cart, right?
> > The two cart
> > 
> > 1. The one before opening the Bolt modal (step 3. Go to cart page)
> > 2. The one after removing the gift card (step 6. Delete the applied gift card by clicking “Remove” button of Amasty GiftCard )
> > 
> > Do you think they are same or not? In my opinion, they are same, even we add more extra info, but they are still same cart. Then why/how do we return different cache identifier for same cart?
> 
> @dumega Currently they are treated the same because there's nothing to distinguish between them in the cache identifier. And they really are not the same, one has a gift cart applied and the other one does not. If we add some gift card related field to the cache identifier then they won't be treated as the same cart anymore and everything should work. I can't know if it 100% would, but it seems far simpler then this pull request so it is worth trying imho. No adding to or changing the existing plugin logic, just adding another part to the identifier (which, I admit, might be a challenge to get right).

@danac-gs If one of them has gift card applied, they would definitely be different, but none of them have gift card, that's why I says they are same. 

Do you mind running some tests on your side to see how it works?

danac-gs:
@dumega I am probably missing something, but I still think something's not right in the process and that it should be done simpler. I will try to reiterate the flow to see if I'm right or not.
1. Purchase a gift card (generated by Amasty GiftCard)
* I am assuming this is done in another order, so we have it available to use in this one
2. Add a simple product to cart
3. Go to cart page
* no cart cached yet
* the plugin collects the cart data
* the cart data has no discounts property populated
* sends it to bolt
* receives the bolt order (cart data)
* generates the cache identifier and stores the bolt order response in cache 
4. Open Bolt modal and apply the gift card in the "Delivery" or "Payment" step
* the gc code is sent to bolt
* the verify discount endpoint is called (bolt -> plugin)
* successfully applied to the quote (we need to apply it to both parent and immutable quotes)
* at this point the state of the cart has changed, but the change is not reflected yet to the underlying shopping cart page, only in checkout modal
* this means there's an update somewhere in DB that connects GC to the quote (we need to find this and use it in the cache identifier, eventually)
5. Keep the Bolt modal open and directly reload the cart page
* at some point (after the page is loaded) a call to cart/data is made
* in the process the plugin loads the quote, creates an immutable quote and generates the cache identifier
* the quote / cart now has the discounts array with one element, the gift card data
* the cache identifier should be different than previously created one because there's a gift card addition to the cart object (even with no adding new property to the cache identifier, so perhaps it is not necessary after all)
* a new cart is generated and fetched, ready to be loaded in modal eventually
6. Delete the applied gift card by clicking “Remove” button of Amasty GiftCard
* I assume this is done via an AJAX call
* when the response is received, the state of the cart (totals updated, no gift card applied) reflects on the front end, shopping cart page
* our plugin detects the change and calls Boltcheckout.configure with a cart parameter as a promise, calling cart/data again and fetches a new cart state as a resolved promise (PERHAPS THE PROBLEM IS HERE, DOES THIS CALL REALLY HAPPENS?)
* expectation is that the cached cart from the step 3. is received (the one with no GC applied)
7. Open the Bolt modal
* actually the cart has not been changed, it remains the same as in step 5.
8. _What happens when you reload the page again?_

**Conclusion**:
* It seems that I was wrong regarding the need of adding to the cache identifier.
* On the other hand, according to the usual flow and functionality (actually the old one I am familiar with, without the bolt cart in sections) this issue should not happen.

**Proposal**: 
* Try with cart caching disabled to validate or invalidate this as a caching issue 
* Verify the create cart / Boltcheckout.configure is triggered in the step 6.

Note: I will reiterate my previous comment from this thread
* _This looks like another side effect of using sections the way we do, and we are still fixing the consequences instead of dealing with the root cause (for a month or so already)_

dumega:
@danac-gs 

> 8. What happens when you reload the page again?

The issue sill exist when you reload the page.

> 
This looks like another side effect of using sections the way we do, and we are still fixing the consequences instead of dealing with the root cause (for a month or so already)

I will verify if this issue is related to Magento sectionns

danac-gs:
Please also try with caching disabled

dumega:
@danac-gs I setup a fresh M2 site and install Bolt plugin v2.7.0 which does not have Magento sections, then this issue is still reproducible. So I believe this is not related to Magento sections.

dumega:
> Please also try with caching disabled

Well, once I set **Cache Bolt Order Token (Performance Optimization)**  to No, the cart can be updated properly. so it is related to cache 

BTW also tested with v2.7.0

danac-gs:
> > Please also try with caching disabled
> 
> Well, once I set **Cache Bolt Order Token (Performance Optimization)** to No, the cart can be updated properly. so it is related to cache
> 
> BTW also tested with v2.7.0

@dumega then I guess updating cacheIdentifier should fix it after all

dumega:
> > > Please also try with caching disabled
> > 
> > 
> > Well, once I set **Cache Bolt Order Token (Performance Optimization)** to No, the cart can be updated properly. so it is related to cache
> > BTW also tested with v2.7.0
> 
> @dumega then I guess updating cacheIdentifier should fix it after all

@danac-gs It just double confirms this is related to cache issue, and my solution would fix it.

danac-gs:
> > > > Please also try with caching disabled
> > > 
> > > 
> > > Well, once I set **Cache Bolt Order Token (Performance Optimization)** to No, the cart can be updated properly. so it is related to cache
> > > BTW also tested with v2.7.0
> > 
> > 
> > @dumega then I guess updating cacheIdentifier should fix it after all
> 
> @danac-gs It just double confirms this is related to cache issue, and my solution would fix it.

@dumega I am sure your solution works, I never doubted that. My point is that this is not the best solution in regard of how caching was implemented and meant to be used. It's more like a workaround.

dumega:
> > > > > Please also try with caching disabled
> > > > 
> > > > 
> > > > Well, once I set **Cache Bolt Order Token (Performance Optimization)** to No, the cart can be updated properly. so it is related to cache
> > > > BTW also tested with v2.7.0
> > > 
> > > 
> > > @dumega then I guess updating cacheIdentifier should fix it after all
> > 
> > 
> > @danac-gs It just double confirms this is related to cache issue, and my solution would fix it.
> 
> @dumega I am sure your solution works, I never doubted that. My point is that this is not the best solution in regard of how caching was implemented and meant to be used. It's more like a workaround.

@danac-gs Yes, I know your point is to update the cacheIdentifier instead. But I have explained in the previous comment that updating cacheIdentifier wont fix this issue, cause same cart content is always related to same cacheIdentifier (I have double checked how we implement cache), especially there is not any other extended data like GiftCard in the cart. 

Or do you mean generating different cacheIdentifier for same cart content? That seems to be illogical. 

And if you insist on your idea, will you be able to give it a try on your side? Maybe I missed something, please correct me with coding.

danac-gs:
@dumega The issue is actually more generic and affects all coupons applied through Bolt modal. What really happens is the following:

1. Create a cart without discounts. Bolt order (cart) is cached (order token).
2. Apply a discount. The cart on Bolt end tied with the order token is updated with discount info.
3. Reload the page. A new immutable quote / cart / order token is generated (with discount).
4. Remove the discount. Plugin checks for cache, there's no discount in cart and the cache identifier is identical to the first cart (1.)
5. The cached (1.) version is returned and with its order token passed to checkout modal. Checkout fetches the cart with that token from Bolt, however on Bolt end that cart has the discount (2.)

This was a bit tricky one, I had persistent and mostly wrong assumptions in my comments. Fortunately, that lead us to a good resolution and a better understanding as well as an appropriate solution.

Please check https://github.com/BoltApp/bolt-magento2/pull/833

dumega:
> @dumega The issue is actually more generic and affects all coupons applied through Bolt modal. What really happens is the following:
> 
> 1. Create a cart without discounts. Bolt order (cart) is cached (order token).
> 2. Apply a discount. The cart on Bolt end tied with the order token is updated with discount info.
> 3. Reload the page. A new immutable quote / cart / order token is generated (with discount).
> 4. Remove the discount. Plugin checks for cache, there's no discount in cart and the cache identifier is identical to the first cart (1.)
> 5. The cached (1.) version is returned and its order token passed to checkout modal. Checkout fetches the cart with that token from Bolt, however on Bolt end that cart has the discount (2.)
> 
> This was a bit tricky one, I had persistent and mostly wrong assumptions in my comments. Fortunately, that lead us to a good resolution and a better problem understanding as well as an appropriate solution.
> 
> Please check #833

@danac-gs Finally we reach an agreement :), your description is exactly as same as I described in my previous comment, so ur solution would cover all the dicount types, right?

danac-gs:
> @danac-gs Finally we reach an agreement :), your description is exactly as same as I described in my previous comment, so ur solution would cover all the dicount types, right?

@dumega Yes, clearing the cache by tag every time discount validation hook is called solves it in general case.

vitaliyreznikov:
@dumega Are you ok to close this one since we are probably want to merge another one https://github.com/BoltApp/bolt-magento2/pull/806 ?

PR #812, opened 2020-07-21
Use plugin for afterPlace rather than observer for orders

nickfiacco-bolt:
There were several issues with using an observer to ingest non-Bolt orders.
First of all, there isn't any consistency across various payment methods
on which events are emitted on order creation. Second of all, the order may
be in various states and possible incomplete/not saved at the time of the
event being emitted. This can lead to all sorts of weird behaviors. Instead,
this PR uses a plugin that should consistently fire after the initial order
is placed and saved.

# Description

Fixes: (link Jira ticket)

#changelog Use plugin for afterPlace rather than observer for orders

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Plugin/NonBoltOrderPlugin.php
+++ b/Plugin/NonBoltOrderPlugin.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Bolt\Boltpay\Observer;
+namespace Bolt\Boltpay\Plugin;
 
 use Bolt\Boltpay\Helper\Api as ApiHelper;
 use Bolt\Boltpay\Helper\Bugsnag;
@@ -9,24 +9,23 @@ use Bolt\Boltpay\Helper\Cart as CartHelper;
 use Bolt\Boltpay\Helper\FeatureSwitch\Decider;
 use Bolt\Boltpay\Helper\Log as LogHelper;
 use Bolt\Boltpay\Helper\MetricsClient;
+use Bolt\Boltpay\Helper\Order;
 use Bolt\Boltpay\Model\Payment;
 use Bolt\Boltpay\Model\Response;
-use Error;
 use Exception;
 use Magento\Customer\Api\Data\CustomerInterface;
 use Magento\Framework\DataObjectFactory;
-use Magento\Framework\Event\ObserverInterface;
-use Magento\Framework\Event\Observer;
 use Magento\Framework\Exception\LocalizedException;
 use Magento\Quote\Api\CartRepositoryInterface as QuoteRepository;
+use Magento\Sales\Api\OrderManagementInterface;
 use Zend_Http_Client_Exception;
 
 /**
- * Class NonBoltOrderObserver
+ * Class NonBoltOrderPlugin
  *
- * @package Bolt\Boltpay\Observer
+ * @package Bolt\Boltpay\Plugin
  */
-class NonBoltOrderObserver implements ObserverInterface
+class NonBoltOrderPlugin
 {
     /**
      * @var ApiHelper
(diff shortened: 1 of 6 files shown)


Plugin/NonBoltOrderPlugin.php
```diff
@@ -149,7 +146,7 @@ public function execute(Observer $observer)
             $response = $result->getResponse();
             if (empty($response)) {
                 $this->metricsClient->processCountMetric("non_bolt_order_creation.failure", 1);
-                return;
+                return $order;
             }
 
             $boltTransactionData = ['bolt_transaction_reference' => $response->reference];
```
danac-gs, line 152: @nickfiacco-bolt
Is it necessary to store this data with the payment instance info of some other payment method?
I still see it as a potential cause of issues on more than one level.
Could we add a column to the order table directly and store it there?
cc @njfaries @johnmccombs1 @matt-thomason @vitaliyreznikov @mccottry @daisy1754

vitaliyreznikov: Nice idea @danac-gs !

nickfiacco-bolt: I can do that-- although for the record PayPal uses this field even for non-PayPal orders in some cases, so the precedent is set if we want to use this field. The reason for the original issue was the timing of when the additional information was set.

nickfiacco-bolt: On second thought there's no downside to a new column, so I'll go with that

nickfiacco-bolt:
Updated to use a new column on the `sales_order_payment` table on @danac-gs suggestion.

I had to upgrade the plugin version in module.xml for the schema change to take place in my testing, should I include that in my PR?

danac-gs:
> Updated to use a new column on the `sales_order_payment` table on @danac-gs suggestion.
> 
> I had to upgrade the plugin version in module.xml for the schema change to take place in my testing, should I include that in my PR?

@nickfiacco-bolt I was thinking of using `sales_order` table not `sales_order_payment` but this seems OK too. I am missing the whole context of the task, so it might be required to be stored here. Anyway, better than mixing the data with other payment methods. @vitaliyreznikov @matt-thomason @njfaries to help decide.

Plugin/NonBoltOrderPlugin.php
```diff
@@ -149,21 +146,19 @@ public function execute(Observer $observer)
...
                 $this->metricsClient->processCountMetric("non_bolt_order_creation.failure", 1);
-                return;
+                return $order;
             }
 
-            $boltTransactionData = ['bolt_transaction_reference' => $response->reference];
-            $payment->setAdditionalInformation(
-                array_merge((array)$payment->getAdditionalInformation(), $boltTransactionData)
-            );
+            $payment->setBoltTransactionReference($response->reference);
```
vitaliyreznikov, line 152: We think that it's possible that `$payment` is null (see line 122). It's not a big deal because we are in `try {}` statement and maybe in real cases, `$payment` is always exists, but probably it's worth fixing it.
We can don't save if we have no payment or save in order table rather than in payment table

Plugin/NonBoltOrderPlugin.php
```diff
@@ -107,37 +106,35 @@ public function __construct(
...
 
     /**
-     * @param Observer $observer
+     * @param OrderManagementInterface $orderManagementInterface
+     * @param Order
+     *
+     * @return Order
      */
-    public function execute(Observer $observer)
+    public function afterPlace(OrderManagementInterface $orderManagementInterface, $order)
```
vitaliyreznikov, line 114: The code looks great and I see no way how we can have any issue with it.
But one more thing for discussion.

We add this code at a very critical moment when users create order.
I can imagine it takes something around 0.5 seconds (data collection + API request).

If someone add 0.5 second to bolt interaction with users I will be a little upset. :)
Can we find a more asynchronous way to collect the data? I guess it's not simple but maybe it worth doing.
Also, we can disable this by default.

What do you think @nickfiacco-bolt @matt-thomason @njfaries  ?

nickfiacco-bolt: This is a great point. What we can do is emit our own event using this plugin. Then add an observer which executed this logic asynchronously. Since we emit the event, we don't have to worry about inconsistent state. Will work on that today.

nickfiacco-bolt: On second look it seems observers are not async. We sort of need it to be synchronous anyway, since we need to ingest this data before it's used for tracking, etc

nickfiacco-bolt: Also-- it already disabled by default

nickfiacco-bolt: Another option would be to send the order information over in the Tracking observer, before sending the tracking info.

vitaliyreznikov: So we need to discuss if we want to add this "asynchronous" things to backlog or not. I am not sure about it.

I don't see any simple solution.
1. we can save non bolt order orders when we handle bolt hooks, this is a good place for asynchronous work. But in this case, we can don't have actual info when a user wants to see their's account
2. we can get most part of data as (1) but also make an additional call from hail to plugin when any user wants to open account

PR #819, opened 2020-07-24
M2P-172 Fix warning issues reported by the Magento coding standard tool in the view folder

johnmccombs1:
# Description

This PR fixes most of the warning issues reported by the Magento coding standard tool except for the **addcslashes** and **ob_start** issue which will be fixed in later PRs.

See the reported issues in the following link: https://gist.github.com/johnmccombs1/12e3e68b7b228d99dffe39a580623fd0

Fixes: https://boltpay.atlassian.net/browse/M2P-172

#changelog Fix warning issues reported by the Magento coding standard tool in the view folder

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Block/Js.php
+++ b/Block/Js.php
@@ -27,7 +27,7 @@ use Magento\Quote\Model\Quote;
 use Bolt\Boltpay\Helper\Bugsnag;
 
 /**
- * Js Block. The block class used in replace.phtml and track.phtml blocks.
+ * Js Block. The block class used in track.phtml block.
  *
  * @SuppressWarnings(PHPMD.DepthOfInheritance)
  */
--- a/Block/Js.php
+++ b/Block/Js.php
@@ -498,4 +498,21 @@ class Js extends Template
     {
         return $this->featureSwitches->isSaveCartInSections();
     }
+
+    /**
+     * @param $jsCode
+     * @param string $argName
+     * @return string
+     */
+    public function wrapWithCatch ($jsCode, $argName = '')
+    {
+        return "
+function($argName) {
+    try {
+        $jsCode
+    } catch (error) {
+        console.error(error);
+    }
+}";
+    }
 }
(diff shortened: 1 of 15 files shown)


mccottry:
@johnmccombs1 This is a good PR, but it also starts to tread down a slippery slope.  It is important to recognize that these are not errors and are not required to be "fixed" to submit our code to the Magento Marketplace.

The reason why I bring this up is that I see warnings for things like the use of `ob_start` is "discouraged."   In this case, if you tried to not use this method to achieve the same thing, the code will undoubtedly wind up looking extremely awkward, and possibly error-prone.  It is also possible that if Magento does not wrap this method in another for use like it used to with `setDirectOutput` in Magento 1, we will not be able to achieve the same functionality.

As long as we take these lint warnings as suggestions and not hard-errors, we should be ok.

Also, at which level are we running the winter?

johnmccombs1:
> @johnmccombs1 This is a good PR, but it also starts to tread down a slippery slope. It is important to recognize that these are not errors and are not required to be "fixed" to submit our code to the Magento Marketplace.
> 
> The reason why I bring this up is that I see warnings for things like the use of `ob_start` is "discouraged." In this case, if you tried to not use this method to achieve the same thing, the code will undoubtedly wind up looking extremely awkward, and possibly error-prone. It is also possible that if Magento does not wrap this method in another for use like it used to with `setDirectOutput` in Magento 1, we will not be able to achieve the same functionality.
> 
> As long as we take these lint warnings as suggestions and not hard-errors, we should be ok.
> 
> Also, at which level are we running the winter?

@mccottry , Thanks for pointing it out. I'll fix the error issues first. 

For the ob_start issue, we're minifying the JS in the view layer which isn't recommended. We should delegate it to the View Model layer (Block). I've created a PR to handle that here: https://github.com/BoltApp/bolt-magento2/pull/820/files

I ran the following command to get and fix the warning issues in the view folder:
vendor/bin/phpcs --standard=Magento2 app/code/Bolt/Boltpay/view

PR #822, opened 2020-07-25
M2P-161 Stop pushing model/quote JS script

vitaliyreznikov:
We introduced an issue in https://github.com/BoltApp/bolt-magento2/pull/796. Actually it was known magento issue that was fixed in new Magento versions, see https://github.com/magento/magento2/pull/23099 , https://github.com/magento/magento2/issues/14412, https://github.com/magento/magento2/pull/21432

Under some conditions, Magento JS script produced an error because they initialated wrong.

We had a fix https://github.com/BoltApp/bolt-magento2/pull/808 but now I see it doesn't affect the root of the issue.

In PR 796 we added `require(['Magento_Checkout/js/model/quote'] ...)` thing. This requireJS API means we push `model/quote` script to load if it wasn't loaded before. Under some very rare conditions, it triggers incorrect initializing.

So solution the same as in https://github.com/magento/magento2/pull/21432 : don't load module/quote script before DOM.loaded. We know that when DOM.loaded we have all parameters to initialize this script.

The issue about a race condition, so it's quite hard to reproduce it. Looks like it happens very rarely.
We need some (or all) of the following conditions:
- site works slowly
- browser work slowly (IE11 for example)
- the first page load, so we don't have scripts in browser cache.
- site works in developer mode

Fixes: [M2P-161]

#changelog M2P-161 Stop pushing model/quote JS script

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -2321,12 +2321,12 @@ if (!$block->isSaveCartInSections()) { ?>
     });
 
     if (trim(location.pathname, '/') === 'checkout/cart') {
+        require(['domReady!'], function() {
         require([
         'jquery',
-        'Magento_Customer/js/customer-data',
         'Magento_Checkout/js/model/quote',
         'domReady!'
-        ], function ($, customerData, quote) {
+        ], function ($, quote) {
             // If quote total was changed through a method we didn't create,
             // we should reload the bolt cart since totals are probably now out of sync.
             quote.totals.subscribe(function (newValue) {
@@ -2341,6 +2341,7 @@ if (!$block->isSaveCartInSections()) { ?>
                 }
                 $(document).trigger('bolt:createOrder');
             });
+       })
        })
     }
 </script>
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -2322,26 +2322,25 @@ if (!$block->isSaveCartInSections()) { ?>
 
     if (trim(location.pathname, '/') === 'checkout/cart') {
         require(['domReady!'], function() {
-        require([
-        'jquery',
-        'Magento_Checkout/js/model/quote',
-        'domReady!'
-        ], function ($, quote) {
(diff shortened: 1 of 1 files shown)


view/frontend/templates/js/replacejs.phtml
```diff
@@ -2321,12 +2321,12 @@ if (!$block->isSaveCartInSections()) { ?>
     });
 
     if (trim(location.pathname, '/') === 'checkout/cart') {
+        require(['domReady!'], function() {
         require([
         'jquery',
-        'Magento_Customer/js/customer-data',
         'Magento_Checkout/js/model/quote',
         'domReady!'
```
matt-thomason, line 2328: This means we do not need this second `domReady!`, right?

view/frontend/templates/js/replacejs.phtml
```diff
@@ -2321,12 +2321,12 @@ if (!$block->isSaveCartInSections()) { ?>
     });
 
     if (trim(location.pathname, '/') === 'checkout/cart') {
+        require(['domReady!'], function() {
         require([
```
matt-thomason, line 2325: Can you fix indentation?

vitaliyreznikov:
> What about the other two instances in this replacejs.phtml file where we rely on domReady!?

Good point, but other things are work :) and we haven't changed them for months. Looks like only '`Magento_Checkout/js/model/quote'` is brittle enough and we use it only in place I fix in this PR.

Maybe it makes sense to run all code after domReady, but I think it's too big change for hotfix PR. Let's discuss this during other JS refactoring tasks..

view/frontend/templates/js/replacejs.phtml
vitaliyreznikov: Correct, thank you

view/frontend/templates/js/replacejs.phtml
vitaliyreznikov: Done

matt-thomason:
That is reasonable, thanks for talking it out.

PR #833, opened 2020-07-29
Resolve issue with discounts applied from Bolt modal persisting after being removed

danac-gs:
# Description
Improvement upon https://github.com/BoltApp/bolt-magento2/pull/806
Actually proved to be a bigger issue affecting other (all?) discounts applied from Bolt modal where cached Bolt cart (order token) would be re-used. This was confirmed and tested for Amasty_GiftCard and Magento Discount Coupon.

Fixes: https://boltpay.atlassian.net/browse/M2P-155

#changelog Resolve issue with discounts applied from Bolt modal persisting after being removed

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Model/Api/DiscountCodeValidation.php
+++ b/Model/Api/DiscountCodeValidation.php
@@ -248,7 +248,7 @@ class DiscountCodeValidation implements DiscountCodeValidationInterface
         $this->regionModel = $regionModel;
         $this->totalsCollector = $totalsCollector;
         $this->orderHelper = $orderHelper;
-        $this->cache = $cache ?? \Magento\Framework\App\ObjectManager::getInstance()
+        $this->cache = $cache ?: \Magento\Framework\App\ObjectManager::getInstance()
                 ->get(\Magento\Framework\App\CacheInterface::class);
     }
 
--- a/Model/Api/DiscountCodeValidation.php
+++ b/Model/Api/DiscountCodeValidation.php
@@ -18,6 +18,7 @@
 namespace Bolt\Boltpay\Model\Api;
 
 use Bolt\Boltpay\Api\DiscountCodeValidationInterface;
+use Magento\Framework\App\CacheInterface;
 use Magento\Framework\Exception\LocalizedException;
 use Magento\Framework\Exception\NoSuchEntityException;
 use Magento\Framework\Webapi\Rest\Request;
@@ -165,6 +166,10 @@ class DiscountCodeValidation implements DiscountCodeValidationInterface
      * @var OrderHelper
      */
     protected $orderHelper;
+    /**
+     * @var CacheInterface
+     */
+    private $cache;
 
     /**
      * DiscountCodeValidation constructor.
@@ -192,6 +197,7 @@ class DiscountCodeValidation implements DiscountCodeValidationInterface
      * @param RegionModel             $regionModel
      * @param TotalsCollector         $totalsCollector
      * @param OrderHelper             $orderHelper
+     * @param CacheInterface          $cache
      */
     public function __construct(
         Request $request,
@@ -216,7 +222,8 @@ class DiscountCodeValidation implements DiscountCodeValidationInterface
         DiscountHelper $discountHelper,
         RegionModel $regionModel,
         TotalsCollector $totalsCollector,
-        OrderHelper $orderHelper
+        OrderHelper $orderHelper,
+        CacheInterface $cache = null
     ) {
         $this->request = $request;
         $this->response = $response;
@@ -241,6 +248,8 @@ class DiscountCodeValidation implements DiscountCodeValidationInterface
         $this->regionModel = $regionModel;
         $this->totalsCollector = $totalsCollector;
         $this->orderHelper = $orderHelper;
+        $this->cache = $cache ?? \Magento\Framework\App\ObjectManager::getInstance()
+                ->get(\Magento\Framework\App\CacheInterface::class);
     }
 
     /**
@@ -451,6 +460,9 @@ class DiscountCodeValidation implements DiscountCodeValidationInterface
                 // Already sent a response with error, so just return.
                 return false;
             }
+
+            //remove previously cached Bolt order since we altered related immutable quote by applying a discount
+            $this->cache->clean([CartHelper::BOLT_ORDER_TAG . '_' . $parentQuoteId]);
 
             $this->sendSuccessResponse($result, $immutableQuote);
         } catch (WebApiException $e) {
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -776,7 +776,7 @@ class Cart extends AbstractHelper
             $this->saveToCache(
                 $boltOrder,
                 $cacheIdentifier,
-                [self::BOLT_ORDER_TAG],
+                [self::BOLT_ORDER_TAG, self::BOLT_ORDER_TAG . '_' . $cart['order_reference']],
                 self::BOLT_ORDER_CACHE_LIFETIME
             );
         }
--- a/Test/Unit/Helper/CartTest.php
+++ b/Test/Unit/Helper/CartTest.php
@@ -2356,7 +2356,7 @@ ORDER
         $currentMock->expects(static::once())->method('saveToCache')->with(
             $boltOrder,
             self::CACHE_IDENTIFIER,
-            [BoltHelperCart::BOLT_ORDER_TAG],
+            [BoltHelperCart::BOLT_ORDER_TAG, BoltHelperCart::BOLT_ORDER_TAG . '_' . self::PARENT_QUOTE_ID],
             3600
         );
         $result = $currentMock->getBoltpayOrder(false, '');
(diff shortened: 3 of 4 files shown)


Model/Api/DiscountCodeValidation.php
```diff
@@ -241,6 +248,8 @@ public function __construct(
         $this->regionModel = $regionModel;
         $this->totalsCollector = $totalsCollector;
         $this->orderHelper = $orderHelper;
+        $this->cache = $cache ?? \Magento\Framework\App\ObjectManager::getInstance()
+                ->get(\Magento\Framework\App\CacheInterface::class);
```
danac-gs, line 252: The approach was used also in Magento core to make changes to constructor safer

vitaliyreznikov:
@danac-gs Could you please provide steps to reproduce with Magento Discount Coupon?

danac-gs:
> @danac-gs Could you please provide steps to reproduce with Magento Discount Coupon?

@vitaliyreznikov Identical to Amasty GC, as in the screencast behind this link
https://drive.google.com/file/d/1VDE_7akFWsFojZ2yLLWdy69-Ko_qF4hF/view

vitaliyreznikov:
@Daniel Please check if I understand the issue correctly

We use some sort of [md5(bolt_cart_content)](https://github.com/BoltApp/bolt-magento2/blob/ac7ee152a28cab9dc1ef0924b9d21a78a222236e/Helper/Cart.php#L581
) as cache_identifier 

1) User adds product to cart <-- we create order with bolt order_token_1 we save it with cache_id_1
2) User adds discount to cart on bolt modal <- we save it with cache_id_2. It's the still the same order_token_1
3) User remove discount on magento cart <- we get result from cache with cache_id_1 and order_token_1

We have correct bolt_cart in the cache, but we didn't call Bolt API because we found data in cache, and order_token_1 has cart content with a discount.
But we need to cal and get new order token: let's say order_token_2.

@Daniel's solution: when we know that discount applied, invalidate the cache (1). It works, but 
1) we need to create a special tag for each quote_id, I think this is additional SQL request
2) this work only for a discount. Once we have another place where the bolt server asks plugin to update cart (new update_order step), we need to repeat this workaround.

What about another approach? We can use immutableQuote id as cache_id and save md5(cart) as part of the payload.
When we get data from the cache by quote id, we compare md5(cart) with an actual value and invalidate the cache if we see a difference.
Let me know if you see any disadvantages in this approach. @danac-gs @matt-thomason @njfaries

Model/Api/DiscountCodeValidation.php
```diff
@@ -241,6 +248,8 @@ public function __construct(
         $this->regionModel = $regionModel;
         $this->totalsCollector = $totalsCollector;
         $this->orderHelper = $orderHelper;
+        $this->cache = $cache ?? \Magento\Framework\App\ObjectManager::getInstance()
```
matt-thomason, line 251: null coalesce is not backwards compatible with PHP5

danac-gs: Done.

danac-gs:
> 1. we need to create a special tag for each quote_id, I think this is additional SQL request

I don't believe this is an issue. It's a feature of the caching system that exists to be used. Tagging every cart (quote) cache entry with the quote id is very handy actually. It lead to an elegant solution with no complexity added.

> 2. this work only for a discount. Once we have another place where the bolt server asks plugin to update cart (new update_order step), we need to repeat this workaround.

All we would need to do is to clear the cache by tag, quote_id which is available in cart data (order_reference). One line of code in the right place. Clear and simple. We have everything else in place.

> What about another approach? We can use immutableQuote id as cache_id and save md5(cart) as part of the payload.

This approach won't work. The purpose of caching the order token is to avoid calling Bolt for every immutable quote created. This way we would save the orders to cache but perhaps never read them from it. 

> When we get data from the cache by quote id, we compare md5(cart) with an actual value and invalidate the cache if we see a difference.

In the approach in the pull request we don't compare anything. More notably, we don't collect any additional  data to be compared (additional SQL requests?). Just clear the cache in the right place.

Model/Api/DiscountCodeValidation.php
```diff
@@ -452,6 +461,9 @@ public function validate()
                 return false;
             }
 
+            //remove previously cached Bolt order since we altered related immutable quote by applying a discount
+            $this->cache->clean([CartHelper::BOLT_ORDER_TAG . '_' . $parentQuoteId]);
```
dumega, line 465: @danac-gs Does it mean we clear all the cache associated with the parent quote? if so, it seems to be aggressive somehow.

danac-gs: @dumega 

All that is cached (and cleared) is the response from bolt (cart with order token). It may seem aggressive but it's cache after all, not essential for correct and stable functionality.

I still believe that this is overall the most optimal solution. We clear the cache when discount is applied. In vast majority of cases we won't be needing the old cached values after this step.

That said, it would be better (almost perfect) if we could tag the cache entry with the Bolt order token and remove only that single entry. Unfortunately, order token is not passed with the cart in the discount validation request.

Anyway, this change is something that won't be ever noticed in practice in terms of website execution speed. The best (and only) place for optimization here is making the code simple and clean.

danac-gs:
@matt-thomason @njfaries @vitaliyreznikov could someone check the CI pipeline configuration?
```
#!/bin/bash -eo pipefail
mkdir ./artifacts
export COMPOSER_MEMORY_LIMIT=3G
export TEST_ENV=php72
export MAGENTO_VERSION=2.3.0
.circleci/scripts/test-merchant-branches.sh ${CIRCLE_BRANCH} false
The following branches will be tested after rebasing against M2P-155-Bugfix---Unable-to-remove-Amasty-GiftCard-V2
ci/jz/rebase-custom1
error: pathspec 'ci/jz/rebase-custom1' did not match any file(s) known to git.
Failed to checkout branch ci/jz/rebase-custom1

Exited with code exit status 1
CircleCI received exit code 1
```

vitaliyreznikov:
@danac-gs Sorry, my comment was wrong. I meant parent quote id, not immutable quote id. Updated now.

danac-gs:
@vitaliyreznikov I don't agree. 

> @danac-gs Looks like the tags weren't developed for this type of use. In the documentation, I found only cases when we use "tagA" and "tagB" to clean part of the cache once it becomes invalid (for example we updated catalog and want to invalidate all cache records related to any catalog items).

This is exactly what is done here, _clean part of the cache once it becomes invalid_. Cleaning all the cache entries for the particular quote doesn't make a difference in practice, as I explained here https://github.com/BoltApp/bolt-magento2/pull/833#discussion_r463434380

> If we look into specific realization, Zend cache for files, we can see that it should go throw all cache items to clear by tag.
> https://github.com/zendframework/zend-cache/blob/1b83f1ce31c9717bc2f340ec177315975d9f8c64/src/Storage/Adapter/Filesystem.php#L391

What Zend does is _Remove items matching given tags._ Exactly what we need and how it's used here.
 
> Looks like we should use another approach and go ahead with quote_id as cache id instead of md5(our_data)

Definitely not. First, the code in this PR is tested and it works. Moreover, it is quite simple and generic. Second, our caching implementation is used for quote a while, proved to do the job well on probably hundreds of instances so far. If it ain't broke, don't fix it (https://www.reddit.com/r/ProgrammerHumor/comments/7vngkm/if_it_aint_broke_dont_fix_it/, https://devrant.com/rants/929065/if-it-aint-broke-dont-fix-it-savage-programmer). Changing it for this particular case is not justified. Finally, having quote_id as cache id as a solution won't work. That way we would always have only one cache instance per quote, overwriting it with the latest state on every change, but with no possibility to go back to some previous state (adding and then removing the product, for instance). 

> I've created task in backlog https://boltpay.atlassian.net/browse/M2P-198

I think this is a wrong approach and it should be discussed.

@matt-thomason @njfaries @dumega @johnmccombs1 @mccottry @roopakv @daisy1754

vitaliyreznikov:
> What Zend does is Remove items matching given tags. Exactly what we need and how it's used here.

If you have 10 000 records in cache, Zend opens 10 000 files to remove cache by tags.
We can have a performance issue here. What's your opinion @danac-gs

danac-gs:
> > What Zend does is Remove items matching given tags. Exactly what we need and how it's used here.
> 
> If you have 10 000 records in cache, Zend opens 10 000 files to remove one cache record by tags.
> We can have a performance issue here. What's your opinion @danac-gs

@vitaliyreznikov I get your point, your concerns are justified, good catch. However, grepping through Magento code base I found it used in about 20 places. Then I realized Zend is not used for file system caching for the exact performance reason you brought up. 

https://github.com/magento/magento2/blob/2.4/lib/internal/Magento/Framework/App/Cache/Frontend/Factory.php#L63
https://github.com/colinmollenhour/Cm_Cache_Backend_File/blob/master/README.md

vitaliyreznikov:
@danac-gs You are right, we shouldn't have performance issue here and I am ok to merge this.
But I still think we should refactor the caching down the road

> That way we would always have only one cache instance per quote, overwriting it with the latest state on every change, but with no possibility to go back to some previous state (adding and then removing the product, for instance).

That's true, but it doesn't look like a problem.

danac-gs:
> @danac-gs You are right, we shouldn't have performance issue here and I am ok to merge this.
> But I still think we should refactor the caching down the road

I really don't see the need. I mean, what's the gain? What's wrong in the current approach that would be better in the new one? What additional value would it add? On the other hand, there is a lot of space for (real) improvements in the plugin.

PR #863, opened 2020-08-05
[M2P-152] [WIP] Removed quote id from response

njfaries:
# Description
As a simple interim step for cleaning up the display ID, we've removed the quote id from the response. I've not been able to get this to change things on my devbox but that's probably more due to my devbox set-up. It's possible there's something I'm missing here so figured opening a PR to get eyes on it makes more sense. This solution seems _too_ easy.

Still need to update unit test

Fixes: go/j/M2P-152

#changelog [M2P-152] [WIP] Removed quote id from response

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Model/Api/CreateOrder.php
+++ b/Model/Api/CreateOrder.php
@@ -226,7 +226,7 @@ class CreateOrder implements CreateOrderInterface
             $this->sendResponse(200, [
                 'status'    => 'success',
                 'message'   => "Order create was successful. Order Data: $orderData",
-                'display_id' => $createdOrder->getIncrementId() . ' / ' . $immutableQuoteId,
+                'display_id' => $createdOrder->getIncrementId(),
                 'total'      => CurrencyUtils::toMinor($createdOrder->getGrandTotal(), $currency),
                 'order_received_url' => $this->getReceivedUrl($immutableQuote),
             ]);
--- a/Test/Unit/Model/Api/CreateOrderTest.php
+++ b/Test/Unit/Model/Api/CreateOrderTest.php
@@ -418,7 +418,7 @@ class CreateOrderTest extends TestCase
         $this->currentMock->expects(self::once())->method('sendResponse')->with(
             200,
             [
-                'display_id'   => 'XXXXX / 456',
+                'display_id'   => 'XXXXX',
                 'message'   => 'Order create was successful. Order Data: {"id":"1111","increment_id":"XXXXX","grand_total":"11.00"}',
                 'order_received_url'   => '',
                 'status'    => 'success',


vitaliyreznikov:
I've tested and it looks good for me.

image


This is the same as we do in other plugins: some technical display_id before we create cart side order and real order id after.
So I don't expect any problem here.

@danac-gs your opinion?

vitaliyreznikov:
We have interaction that happend after magento order creation: hooks handling.

Could you please refactor function `getOrderStoreIdByDisplayId`? They should provide correct info for both cases:
- when we have display_id like <order increment ID / immutable quote ID>
- when we have display_id like <order increment ID>

https://github.com/BoltApp/bolt-magento2/blob/ac7ee152a28cab9dc1ef0924b9d21a78a222236e/Model/Api/OrderManagement.php#L167-L168

I am ok to merge the PR after this fix and updating tests

danac-gs:
I expect this to work in a "normal" case but there are special cases where immutable quote is really needed. They are introduced to solve problems we could not solve differently.

danac-gs:
Guys, when introducing changes like this that can have unpredictable side effects and are checkout / order related, please run all the checkout cases (card types) from the production readiness spreadsheet during the development phase, before the PR is created and code passed to QA. No point in trying to release the code that we as developers are not at least 80% sure that works (100% is much better). Otherwise the release process gets slowed down by creating new release candidates and starting testing from the beginning again and again.

matt-thomason:
Yes, let us hold this off, this is too risky for 2.12 in its current state.

njfaries:
After talking with @vitaliyreznikov earlier the plan was I was going to close this PR for now and work on going through all the uses of display_id and immutableQuoteId, but should I just put this ticket to rest for now and we come back to it later?

vitaliyreznikov:
One more idea: we can save immutable quote id in the magento order for now.
In this case, we can take it from the magento order when we handle hooks and reduce the impact on the rest of the program

vitaliyreznikov:
Or save immutable quote id in metadata of bolt order. This idea looks like the simplest.
- When we create bolt order we continue to save immutable quote id in dispay_id and save it in metadata as well.
- When we create magento order we remove immutable quote id from display_id
- When we handle webhook we can find immutable quote id in transaction metadata

danac-gs:
> One more idea: we can save immutable quote id in the magento order for now.
> In this case, we can take it from the magento order when we handle hooks and reduce the impact on the rest of the program

Perhaps we could store it in the bolt cart metadata. It was not available at the time immutable quotes were introduced and we had to place it somewhere. This way no real change in logic is required, just in storing and reading the value.
@vitaliyreznikov @matt-thomason @njfaries

danac-gs:
> Or save immutable quote id in metadata of bolt order. This idea looks like the simplest.
> 
> * When we create bolt order we continue to save immutable quote id in dispay_id and save it in metadata as well.
> * When we create magento order we remove immutable quote id from display_id
> * When we handle webhook we can find immutable quote id in transaction metadata

Or we make it uniform, storing / reading it from one place only - metadata, and remove it from display_id. A little bit more work but it's better (more clear) to not having it in multiple places imho.

PR #877, opened 2020-08-11
MA-36 Add support for plugin Mageplaza Shipping Restriction

vitaliyreznikov:
# Description

Fixes: [MA-36](https://boltpay.atlassian.net/browse/MA-36)

#changelog MA-36 Add support for plugin Mageplaza Shipping Restriction

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Session.php
+++ b/Helper/Session.php
@@ -199,7 +199,7 @@ class Session extends AbstractHelper
         }
         $this->replaceQuote($quote);
 
-        // trird party plugin support
+        // third party plugin support
         $mageplazaHelper = $this->mageplazaShippingRestrictonHelper->getInstance();
         if ($mageplazaHelper && $mageplazaHelper->isEnabled()) {
             $this->coreRegistry->register('mp_shippingrestriction_cart', $quote->getBoltParentQuoteId());
--- a/Helper/Session.php
+++ b/Helper/Session.php
@@ -25,10 +25,12 @@ use Magento\Customer\Model\Session as CustomerSession;
 use Bolt\Boltpay\Helper\Log as LogHelper;
 use Magento\Framework\App\CacheInterface;
 use Magento\Quote\Model\Quote;
+use Magento\Framework\Registry;
 use Magento\Framework\App\State;
 use Magento\Framework\App\Area;
 use Magento\Framework\Data\Form\FormKey;
 use Bolt\Boltpay\Helper\Config as ConfigHelper;
+use Bolt\Boltpay\Model\ThirdPartyModuleFactory;
 
 /**
  * Boltpay Session helper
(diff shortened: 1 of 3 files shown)


Helper/Session.php
```diff
@@ -184,6 +198,13 @@ public function loadSession($quote)
             $this->customerSession->loginById($customerId);
         }
         $this->replaceQuote($quote);
+
+        // trird party plugin support
+        $mageplazaHelper = $this->mageplazaShippingRestrictonHelper->getInstance();
+        if ($mageplazaHelper && $mageplazaHelper->isEnabled()) {
+            $this->coreRegistry->register('mp_shippingrestriction_cart', $quote->getBoltParentQuoteId());
+            $this->coreRegistry->register('mp_shippingrestriction_address', $quote->getShippingAddress());
```
danac-gs, line 206: This should be better done in a plugin for loadSession than here directly.

vitaliyreznikov: Could you explain more? Why do you think it's better to create the plugin?

danac-gs: It's more atomic and easier to maintain if we have more similar requests like this in the future. Not for this method only but in general. Moreover, if written like that the integration extension (which is done first in many cases) and the eventual core support would look the same. Anyway, I'll remove my change request and make it into a comment. It should not be my decision only.

vitaliyreznikov: Thanks, Daniel. In my opinion, we shouldn't add plugins to our own classes.
When we add code into the specific method we have better readability.
Also, plugins don't provide any performance advantage.

danac-gs: I think that, ideally, our core plugin should deal with core Magento only and everything that's 3rd party related should come in some form of extension, eg. a plugin to our own classes in cases like this.

vitaliyreznikov: I understand your point. Maybe once we support more 3p plugins we can create separate classes for each 3p plugins, but I think we should call them directly, not over the plugins (interceptors).

danac-gs: After taking another look, I definitely think this should have no place in core. Although there's similar (wrong) approach used in many places before. Many of them done by me, I admit. If we hardcode a 3rd party module support like this chances are when supported module changes / updates we would need to split it somehow to support both new and old versions. That adds complexity to the core but could be solved by having 2 plugins that are version constrained.

vitaliyreznikov: I think we can consider different ways how to handle complexity, including support different plugins versions, but I don't think these considerations should block this PR, especially taking into account we have a lot of similar code, as you mentioned.
Let's think about this, but using plugins(interceptors) still doesn't looks like the best solution for me, because this approach reduces code readability.

matt-thomason: I understand both points laid out here. I think in the short term the easiest solution is to separate 3p plugins out into separate classes for readability. Splitting out each 3p plugin into its own module might be something we can do later down the line, but I don't think we get enough benefits there at this time.

In either case, I think refactoring is beyond the scope of this PR. I think this code is fine as it is for now.

PR #879, opened 2020-08-11
Admin Notification of version updates

mccottry:
Original Proposal:
https://docs.google.com/document/d/1Z67GkiGMfgtN_ejPMfWAzSSPKphIhUepPjOeWYVtLaI

Updating the Bolt plugin on merchant sites has generally been handled by contractors via merchant experience, integration engineers, or merchant self-upgrades, usually initiated by request of either support or customer success managers in response to production problems. This is not ideal.

This allows update requests to be triggered automatically upon any major, minor, or patch version updates. The available update is detected by comparing the installed version to the latest on Magento Marketplace (retrieved via Magento composer repository). Alternatively, activating `M2_BOLT_UPDATE_USE_GITHUB` feature switch will instead use the latest Github release for comparison.

Update check happens in admin area, after login. If an update is available:

1. Display custom BoltPopup on the first page after admin detailing steps needed to update
image
image
2. Display notification on top of every page until Bolt is updated
image
3. Add a message to admin notification inbox
image

#changelog Admin Notification of version updates

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Config.php
+++ b/Helper/Config.php
@@ -125,6 +125,11 @@ class Config extends AbstractHelper
     const XML_PATH_PRODUCT_PAGE_CHECKOUT = 'payment/boltpay/product_page_checkout';
 
     /**
+     * Disable notifications for non-critical updates
+     */
+    const XML_PATH_DISABLE_NOTIFICATIONS_FOR_NON_CRITICAL_UPDATES = 'payment/boltpay/disable_notifications_for_non_critical_updates';
+
+    /**
      * Enable Bolt order management
      */
     const XML_PATH_PRODUCT_ORDER_MANAGEMENT = 'payment/boltpay/order_management';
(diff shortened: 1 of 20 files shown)


mccottry:
> Could you please create Jira task for this and make sure it's approved by Abhinay?
> From my point of view, it can be a little aggressive to notify in this way.
> Also, I disagree that we want to show each update from 2.x.0 -> 2.x.1 as addressed a critical issue. We need to mark critical issues manually and also when we have a critical bug we usually want to communicate with merchants personally.
> I don't think we should show notice at the same minute as the version launched. It's better to have a gap, for example, 1 hour, in case we published release by error and fixed It shortly.
> We can want to show different messages to merchants who didn't update for months and for a merchant who updated almost every release.
> Maybe I wrong, but looks like we can discuss the approach before we discuss the code.

@vitaliyreznikov 
Consult Abhinay directly regarding Jira.  The task certainly did exist in Jira last week.

> Also, I disagree that we want to show each update from 2.x.0 -> 2.x.1 as addressed a critical issue.

When using [semantic versioning](https://semver.org/
), moving from 2.x.0  to 2.x.1 implies a bug patch.  These are seen as critical.  

>I don't think we should show notice at the same minute as the version launched. It's better to have a gap, for example, 1 hour, in case we published release by error and fixed It shortly.

Releases should not be published in error.  If that is happening, then the release process is broken.  Nevertheless, merchants are not notified "the same minute".  They are notified on their first login after there has been a release.

> We can want to show different messages to merchants who didn't update for months and for a merchant who updated almost every release.

I'm not sure what the value to this is.  A uniform message seems more logical and is the approach used by all other software that I see.

> Maybe I wrong, but looks like we can discuss the approach before we discuss the code.

The discussion was here:
https://docs.google.com/document/d/1Z67GkiGMfgtN_ejPMfWAzSSPKphIhUepPjOeWYVtLaI

mccottry:
.

PR #883, opened 2020-08-14
Aheadworks Gift Card (General M2 3rd-Party Module Extension Architecture )

mccottry:
# Description
This pull request adds support for Aheadworks Gift Cards while introducing a concise, robust architecture for adding support of 3rd-party modules to the Bolt plugin.

I will post a document here outlining the features and rationale as time permits, which will most likely be on Monday.
*** update: This will have to come next week 24.8-30.8 ***

Fixes: [M2P-92](https://boltpay.atlassian.net/browse/M2P-92)

#changelog Aheadworks Gift Card (General M2 3rd-Party Module Extension Architecture )

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Plugin/ThirdPartySupport/AbstractPlugin.php
+++ b/Plugin/ThirdPartySupport/AbstractPlugin.php
@@ -1,0 +1,79 @@
+<?php
+/**
+ * Bolt magento2 plugin
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ *
+ * @category   Bolt
+ * @package    Bolt_Boltpay
+ * @copyright  Copyright (c) 2017-2020 Bolt Financial, Inc (https://www.bolt.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+namespace Bolt\Boltpay\Plugin\ThirdPartySupport;
+
+/**
+ * Class AbstractPlugin
+ *
+ * @property \Bolt\Boltpay\Helper\Log                    logHelper Bolt logging helper
+ * @property \Bolt\Boltpay\Helper\Bugsnag                bugsnagHelper Bolt bugsnag helper
+ * @property \Magento\Framework\Module\Manager           moduleManager Magento module manager
+ * @property \Magento\Framework\Module\ResourceInterface moduleResource Magento module resourc
+ * @property string                                      supportedModule Defines the module that needs to be enabled in order for the plugin to execute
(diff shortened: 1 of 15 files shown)


Plugin/ThirdPartySupport/AbstractPlugin.php
```diff
@@ -0,0 +1,79 @@
...
+     * and version is inside the configured threshold
+     *
+     * @return bool true if plugin should run, otherwise false
+     */
+    public function shouldRun()
+    {
+        $moduleVersion = $this->moduleResource->getDataVersion($this->supportedModule);
+        return $this->moduleManager->isEnabled($this->supportedModule)
+            && ($this->versionFrom == null || version_compare($this->versionFrom, $moduleVersion, '<='))
+            && ($this->versionTo == null || version_compare($this->versionTo, $moduleVersion, '>='));
```
dumega, line 61: Do we really need versionTo? I do not think we only support specific versions, then abandon the upcoming versions.

danac-gs: @dumega If I understand correctly, we do because this approach gives us great flexibility. Now we can support multiple versions of the same module with separate plugins, without bloating the main plugin code by introducing any new logic there. 
For example we could have:

Plugin/ThirdPartySupport/Aheadworks/Giftcard/V1 that supports versions up to, say, 1.2.3
Plugin/ThirdPartySupport/Aheadworks/Giftcard/V2 with support for 1.2.4 - 1.3.1
...
Plugin/ThirdPartySupport/Aheadworks/Giftcard/V3 with support for versions greater than or equal to  2.0.0

Ideally, we should move all non-core Magento module support from our module core to its internal extensions - plugins (interceptors). That would add a lot to maintainability and scalability in the future and the solution implemented here seems like a way to go.

mccottry: @dumega 
Daniel is exactly right.  Yes, we do need versionTo.  We are never guaranteed that the the future releases of the third party code will be compatible with the past, so our extensions can declare boundaries surrounding the test and approved version for that extension code and then release another version for any future non-compatible changes.  I observed that you had a PR in which this would simplify the solution and somewhat served as an inspiration for this construct.

dumega: I see, that make sense.

johnmccombs1 (approved):

danac-gs:
@matt-thomason @njfaries @vitaliyreznikov 
Before we merge it I think this is an important milestone that everyone should be aware of. IE team too.

vitaliyreznikov:
I think `versionFrom` and `versionTo` is great stuff. I also agree we should create a separate classes for code related to each 3p module.
But I am not sure about using plugins inside our modules. It looks like overkill. We use plugins if we are unable to change the parent class but in this case, we can change it and just call the necessary methods directly.

We discussed this one briefly with @matt-thomason and @njfaries and decided to not include this PR into 2.13 and make a decision next week.

mccottry:
.

PR #884, opened 2020-08-17
M2P-197 Allow merchant extensions to hide bolt button under certain conditions

vitaliyreznikov:
* Reuse existing bolt plugin setting toggle_checkout. It was developed to allow toggle Bolt/Magento checkout depends on the cart content
* If toggle_checkout enabled hide bolt button by default. We show it again when call configure() if conditions are met
* Change the method how we show/hide Magento buttons for toggle_checkout. Now it works with “display:none!important” which is default setting.
* Move variable restricted for toggle_checkout into BoltState
* Create variable globalRestricted for disabling bolt for each cart (for example, depending on user type)
* Add barrier to allow extensions postpone configure() calls

Extension JS code example (it shows native Magento buttons after 10 seconds:  
```
BoltState.beforeConfigureBarrier = boltBarrier();
setTimeout(function () {
    BoltState.globalRestricted = true;
    BoltState.beforeConfigureBarrier.resolve(true);                
}, 10000);
```

Additional configuration should be

```
     {
       "toggleCheckout": {
         "active": true,
         "magentoButtons": [                       
           "#top-cart-btn-checkout",
           "button[data-role=proceed-to-checkout]"
         ],
         "showElementsOnLoad": [                      
           ".checkout-methods-items",                
           ".block-minicart .block-content > .actions > .primary"
         ]
       }
     }
```

Fixes: M2P-197

#changelog M2P-197 Allow merchant extensions to hide bolt button under certain conditions

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Block/Js.php
+++ b/Block/Js.php
@@ -458,11 +458,18 @@ class Js extends Template
      */
     public function getButtonCssStyles()
     {
-        $buttonColor = $this->configHelper->getButtonColor();
-        if (!$buttonColor) {
-            return "";
+        $style = "";
+
+        $toggleCheckout = $this->configHelper->getToggleCheckout();
+        if ($toggleCheckout && $toggleCheckout->active) {
+            $style .= 'display:none;';
         }
-        return '--bolt-primary-action-color:' . $buttonColor;
+
+        $buttonColor = $this->configHelper->getButtonColor();
+        if ($buttonColor) {
+            $style .= '--bolt-primary-action-color:' . $buttonColor .';';
+        }
+        return $style;
     }
 
     /**
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -240,9 +240,9 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
         initiateCheckout: null,
 
         /**
-         * @var bool true if BoltCheckout.configure() called at least once
+         * @var bool true if we loaded all JS code from this module and extensions
          */
-        congigureCalled: false,
+        codeIsLoaded:  boltBarrier(),
 
         /**
          * @var boltBarrier allow extensions to postpone all configure() calls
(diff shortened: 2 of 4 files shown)


view/frontend/templates/js/replacejs.phtml
```diff
@@ -443,8 +472,14 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
...
             // Show magento buttons if bolt checkout is restricted, otherwise hide it
+            var style;
+            if (boltRestricted) {
+                style = 'display:inline-block !important'
+            } else {
+                style = 'display:none !important'
+            }
             $.each(toggleCheckoutConfig.magentoButtons, function( index, button ) {
-                $(button).toggle(boltRestricted);
+                $(button).attr('style', style);
```
danac-gs, line 482: If done like this we are overwriting and losing eventually existing additional inline styles.

vitaliyreznikov: Thank you Daniel, good catch. I've addressed this, could you please review again?

Block/Js.php
```diff
@@ -458,11 +458,18 @@ public function isBoltProductPage()
      */
     public function getButtonCssStyles()
     {
+        $style = "";
+
+        $toggleCheckout = $this->configHelper->getToggleCheckout();
+        if ($toggleCheckout && $toggleCheckout->active) {
+            $style .= 'display:none;';
+        }
```
njfaries, line 466: not sure if we stick carefully to affirmative naming but not sure about the toggle being `active` to hide the button. Thoughts @matt-thomason @vitaliyreznikov @danac-gs ?

vitaliyreznikov: $toggleCheckout->active is a static parameter.
So config can be like
```
     {
       "toggleCheckout": {
         "active": true,
         "magentoButtons": [                       
           "#top-cart-btn-checkout",
           "button[data-role=proceed-to-checkout]"
         ],
         "showElementsOnLoad": [                      
           ".checkout-methods-items",                
           ".block-minicart .block-content > .actions > .primary"
         ]
       }
     } 
```

If we change `active:true` to `active: false` it means the same as if we delete all config.
This is like things work in our codebase and I leave it here to be consistent.

njfaries: Okay, if it's internally consistent then looks good to me

PR #890, opened 2020-08-20
M2P-179 JS refactoring

vitaliyreznikov:
This refactoring opens the opportunity to use promises from the beginning of the code.
In IE11 we use polyfill library from connect.js that allows promises.
So if we use promises before connect.js loading it's not IE compatible.

In this PR:
- move Magento cart listener into function createOrder
- check if Magento cart or Bolt cart was initialized before we setup listener (I can't reproduce this but potentially it could be an issue.
- make sure we call createOrder exactly once (for multistep for now; will fix payment only in the corresponding task
- rename createOrder to boltCheckoutSetup

Fixes: (link Jira ticket)

#changelog M2P-179 JS refactoring

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -352,7 +352,6 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
             var publishableKey = getCheckoutKey();
             if (scriptTag) {
                 scriptTag.setAttribute('data-publishable-key', publishableKey);
-                createOrder();
                 return;
             }
             scriptTag = document.createElement('script');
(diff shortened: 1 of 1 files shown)


view/frontend/templates/js/replacejs.phtml
```diff
@@ -352,7 +352,6 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
             var publishableKey = getCheckoutKey();
             if (scriptTag) {
                 scriptTag.setAttribute('data-publishable-key', publishableKey);
-                createOrder();
```
danac-gs, line 355: The whole `insertConnectScript` function seems redundant. The connect.js is already included on every page.

vitaliyreznikov: I agree with you, but let me handle this in another PR.

PR #895, opened 2020-08-21
M2P-217 Return empty discount type for empty coupon code

vitaliyreznikov:
For store integral discount coupon code could be empty and we need to return empty discount type.
Without this fix magento generates exception and checkout is blocked for users.

Fixes: (link Jira ticket)

#changelog M2P-217 Return empty discount type for empty coupon code

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Discount.php
+++ b/Helper/Discount.php
@@ -1294,7 +1294,7 @@ class Discount extends AbstractHelper
     public function convertToBoltDiscountType($couponCode)
     {
         if ($couponCode == "") {
-            return "";
+            return "fixed_amount";
         }
         $coupon = $this->loadCouponCodeData($couponCode);
         // Load the coupon discount rule
--- a/Helper/Discount.php
+++ b/Helper/Discount.php
@@ -1320,7 +1320,7 @@ class Discount extends AbstractHelper
                 return "shipping";
         }
 
-        return "";
+        return "fixed_amount";
     }
     
     /**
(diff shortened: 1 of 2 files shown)


Helper/Discount.php
```diff
@@ -1293,6 +1293,9 @@ public function loadCouponCodeData($couponCode)
      */
     public function convertToBoltDiscountType($couponCode)
     {
+        if ($couponCode == "") {
+            return "";
+        }
```
danac-gs, line 1298: We probably need this case covered at Bolt's side for consistency. Meanwhile, perhaps some existing discount type should be returned instead of an empty string. `fixed_amount` maybe, cause this discounts behave similar to  (the same as?) coupons, except there's no coupon and it is auto-applied to cart when some conditions match.

danac-gs (changes requested):

Helper/Discount.php
vitaliyreznikov: Yes, valuable comment. I changed to fixed_amount for now and wiil discuss with @melvin0008 if it's the best option.

danac-gs:
please fix tests

danac-gs (approved):

Helper/Discount.php
vitaliyreznikov: @melvin0008 confirmed that this is a correct way.

PR #899, opened 2020-08-27
M2P-223 Use parent quote id as a fallback to get the existing order

johnmccombs1:
# Description
This PR resolves the following case by using the parent quote id to get the existing order

1. The customer clicks the Pay button in the Bolt modal
2. The create_order hook is sent to Magento and receives a time out error
3. There is a third party that updates the Magento increment id for the created order
4. Because of #2, there isn’t any successful response sent back to Bolt, so the Bolt increment id isn’t updated on Bolt
5. The pending hook is sent to Magento, it isn’t able to get the existing order by the Bolt increment id, so it tries to create the order and then it throws an exception

=> The expected behavior for #5 is that the pending hook is able to get the existing order to authorize and update the order

Fixes: https://boltpay.atlassian.net/browse/M2P-223

#changelog M2P-223 Use parent quote id as a fallback to get the existing order

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -830,7 +830,7 @@ class Order extends AbstractHelper
         ///////////////////////////////////////////////////////////////
 
         // check if the order exists
-        $order = $this->getExistingOrder($incrementId);
+        $order = $this->getExistingOrder($incrementId, $parentQuoteId);
 
         // if not create the order
         if (!$order || !$order->getId()) {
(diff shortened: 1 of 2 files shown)


Helper/Order.php
```diff
@@ -1255,12 +1255,19 @@ public function deleteOrderByIncrementId($displayId)
...
+     * @return false|OrderInterface|OrderModel
      */
-    public function getExistingOrder($orderIncrementId)
+    public function getExistingOrder($orderIncrementId, $quoteId = null)
     {
         /** @var OrderModel $order */
-        return $this->cartHelper->getOrderByIncrementId($orderIncrementId, true);
+        $order = $this->cartHelper->getOrderByIncrementId($orderIncrementId, true);
+
+        if ($quoteId && (!$order || !$order->getId())) {
```
vitaliyreznikov, line 1266: Could you please add a small comment why we need this: like if order_id changed by 3p module and we had timeout issue

vitaliyreznikov (approved):
Looks good, one minor comment

danac-gs (approved):

Helper/Order.php
johnmccombs1: @vitaliyreznikov I've updated. Can you re-check?

danac-gs (approved):

vitaliyreznikov (approved):

PR #900, opened 2020-08-30
M2P-92 Add class for support third party plugins

vitaliyreznikov:
Revert https://github.com/BoltApp/bolt-magento2/pull/877/ and make the same plugin customization using new scalable approach.
Instead of injecting ThirdPartyModuleFactory class for each third party class we want to support this PR introduces new class with the following responsibility:
- Persist (in const) list of events with classes we need to run when the event happened and if classes exist
- Check that 3p classes exists and call the specific method if so.

Details:
https://docs.google.com/document/d/1d46e2IQSk9Eih1FqYbu3JP1ZvxbEHVJCf7dMZRPf2b4/edit#

Fixes: [M2P-92](https://boltpay.atlassian.net/browse/M2P-92)

#changelog M2P-92 Add class for support third party plugins

# Type of change

- [ ] Bug fix (change which fixes an issue)
- [x] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?
Please validate that you have tested your change in at least one of the following areas:

- [x] Successfully tested locally (or docker image)
- [ ] Successfully tested on a staging or sandbox server
- [ ] Successfully tested on a merchant's staging server

# For PR Reviewer 
- [ ] Reviewed unit tests to make sure we are using real components rather than mocks as much as possible?
- [ ] For any major change (observer, new Bolt feature, core Magento interaction) we must add a feature switch, did you verify this?

# Checklist:

- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have created or modified unit tests to sufficiently cover my changes.
- [x] I have added my Jira ticket link and provided a changelog message.

PR #908, opened 2020-09-01
M2P-92 Add support for module Aheadworks/Giftcart

vitaliyreznikov:
# Description
- Add new entity - filters - into class EventsForThirdPartyModules.php
- Add support for module Aheadworks/Giftcart:
Refactor code from https://github.com/BoltApp/bolt-magento2/pull/883 with the approach proposed in https://github.com/BoltApp/bolt-magento2/pull/900

Fixes: [M2P-92](https://boltpay.atlassian.net/browse/M2P-92)

#changelog M2P-92 Add support for module Aheadworks/Giftcart

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -63,6 +63,7 @@ use Bolt\Boltpay\Helper\MetricsClient;
 use Bolt\Boltpay\Helper\FeatureSwitch\Decider as DeciderHelper;
 use Magento\Catalog\Model\Config\Source\Product\Thumbnail as ThumbnailSource;
 use Magento\Framework\Serialize\Serializer\Serialize;
+use Bolt\Boltpay\Model\EventsForThirdPartyModules;
 
 /**
  * Boltpay Cart helper
@@ -187,6 +188,11 @@ class Cart extends AbstractHelper
      */
     private $customerRepository;
 
+    /**
+     * @var EventsForThirdPartyModules
+     */
+    private $eventsForThirdPartyModules;
+
     // Billing / shipping address fields that are required when the address data is sent to Bolt.
     private $requiredAddressFields = [
         'first_name',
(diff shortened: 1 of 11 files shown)


danac-gs (commented):
I can not approve this PR, nor deny it. The code itself looks great. However, I think the approach is wrong on many levels. Probably a meeting is needed to discuss them. In a nutshell, this is reinventing the wheel, introducing some new ways and patterns (for Magento, probably usual in WP) for the common cases that have standard, well known solutions, accepted as best practices. The same problem is already solved here https://github.com/BoltApp/bolt-magento2/pull/883. Moreover, that is a proposal for the generic approach of supporting different versions of the same 3rd party modules that, imho, is a great peace of work that should became our model for all 3rd parties. Strategically, that is the way to go, otherwise we would have real scaling issues in the future.
cc @matt-thomason @dumega @johnmccombs1 @usmanismail-bolt @roopakv

vitaliyreznikov:
From my point of view, this one is ready to merge. We can't use plugins to support third-party modules as it proposed in https://github.com/BoltApp/bolt-magento2/pull/883.
- It is not readable. When we read cart helper code we don't see that plugins will apply after that
- It needs a lot of code to implement, many changes in di.xml
- I can't say that this is a standard solution. We can't see in Magento documentation that they propose to use plugins to override classes in the same module. Plugin is a heavy approach that needed when we need to change other module behavior.

Some details here https://docs.google.com/document/d/1d46e2IQSk9Eih1FqYbu3JP1ZvxbEHVJCf7dMZRPf2b4/edit# and here https://github.com/BoltApp/bolt-magento2/pull/900

danac-gs:
We should reconsider our strategy and think months (years?) in advance. My opinion is that this is a bad approach, probably even a dangerous one. We should definitely use plugins. Being unreadable and requiring a lot of code to implement are not valid arguments. They can be very readable with a good GUI. Lot of code maybe, but using (Magento) standard patterns that everyone should understand. Not just we here at Bolt, but Magento devs in general (working for merchants for instance).

njfaries (approved):
Approved following discussion earlier today and pending conflict fixes

njfaries (approved):

matt-thomason (approved):

PR #909, opened 2020-09-01
EN-674 Ensure the returned order in the create_order hook is the latest updated order

johnmccombs1:
# Description
This PR resolves the following case:
1. The customer clicked the Pay button on the Bolt modal
2. The create_order hook (thread 1) was sent to Magento and received a timeout response because **$this->quoteManagement->submit($quote);** took a long time to run. See here - https://prnt.sc/u97pim . The order was created with the status of Pending Payment, and the create_hook process was still running.
3. The authorize/payment hook (thread 2) was sent to Magento to update the order state to processing.
4. The create_order hook (thread 1) updated the order state to pending payment.

This PR fixes the above case by ensuring the returned order in the create_order hook is the latest updated order to prevent #4 from happening.

Fixes: https://boltpay.atlassian.net/browse/EN-674

#changelog Ensure the returned order in the create_order hook is the latest updated order

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -1095,7 +1095,8 @@ class Order extends AbstractHelper
     {
         /** @var OrderModel $order */
         $order = $this->quoteManagement->submit($quote);
-        if ($order === null) {
+
+        if (!$order) {
             $this->bugsnag->registerCallback(function ($report) use ($quote) {
                 $report->setMetaData([
                     'CREATE ORDER' => [
@@ -1115,6 +1116,21 @@ class Order extends AbstractHelper
                 CreateOrder::E_BOLT_GENERAL_ERROR
             );
         }
+        // When the create_order hook (thread 1) takes a lot of time to execute and returns a timeout, the authorize/payment hook (thread 2) is sent to Magento.
+        // It updates the order prior to the create_order hook (thread 1) process.
+        // This ensures that the returned order in the create_order hook is the latest updated order.
+        $existingOrder =  $this->cartHelper->getOrderById($order->getId());
+        $orderPayment = $existingOrder->getPayment();
+        if ($orderPayment && $orderPayment->getMethod() !== Payment::METHOD_CODE) {
+            throw new LocalizedException(__(
+                'Payment method assigned to order %1 is: %2',
+                $order->getIncrementId(),
+                $orderPayment->getMethod()
+            ));
+        }
+
+        $order = $existingOrder;
+
         $order->addStatusHistoryComment(
             "BOLTPAY INFO :: This order was created via Bolt Pre-Auth Webhook"
         );
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -432,6 +432,24 @@ class Cart extends AbstractHelper
     }
 
     /**
+     * Load Order by order id
+     *
+     * @param $orderId
+     * @return OrderInterface|mixed
+     */
+    public function getOrderById($orderId)
+    {
+        $searchCriteria = $this->searchCriteriaBuilder
+            ->addFilter('main_table.entity_id', $orderId)->create();
+
+        $collection = $this->orderRepository
+            ->getList($searchCriteria)
+            ->getItems();
+
+        return reset($collection);
+    }
+
+    /**
      * Save quote via repository
      *
      * @param CartInterface $quote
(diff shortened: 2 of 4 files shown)


Helper/Order.php
```diff
@@ -1094,8 +1094,12 @@ public function processExistingOrder($quote, $transaction)
     public function processNewOrder($quote, $transaction)
     {
         /** @var OrderModel $order */
-        $order = $this->quoteManagement->submit($quote);
-        if ($order === null) {
+        $this->quoteManagement->submit($quote);
+        // When the create_order hook (thread 1) takes a lot of time to execute and returns a timeout, the authorize/payment hook (thread 2) is sent to Magento.
+        // It updates the order prior to the create_order hook (thread 1) process.
+        // This ensures that the returned order in the create_order hook is the latest updated order.
+        $order = $this->getExistingOrder($quote->getReservedOrderId());
```
vitaliyreznikov, line 1101: Very soon (like today) we will stop to use reserverOrderId. Can we use parentQuoteID instead?

johnmccombs1: @vitaliyreznikov I've updated the code to get the order by the quote id

johnmccombs1: Can you re-check?

vitaliyreznikov: I am sorry John I didn't consider that function getOrderByQuoteId is a litle heavy. Could you please make one more change:
- At line 1097 save order_id
- Find order by id and only if doesn't exist (3p plugin changed number) find by quote id. I think we have a helper for this.

johnmccombs1: @vitaliyreznikov 
**"Find order by id and only if doesn't exist (3p plugin changed number) find by quote id. I think we have a helper for this."**
I'm confused about this point. The order id is the primary key and it can't be changed by the 3rd party plugin. Are you saying that we should fetch the order by the increment id first and then if it doesn't exist, fetch it by the quote id as a fallback?

vitaliyreznikov: Thank's for your explanation, John. You are right. Could we use order id for our purpose?

Helper/Order.php
```diff
@@ -1094,8 +1094,12 @@ public function processExistingOrder($quote, $transaction)
     public function processNewOrder($quote, $transaction)
     {
         /** @var OrderModel $order */
-        $order = $this->quoteManagement->submit($quote);
-        if ($order === null) {
+        $this->quoteManagement->submit($quote);
+        // When the create_order hook (thread 1) takes a lot of time to execute and returns a timeout, the authorize/payment hook (thread 2) is sent to Magento.
+        // It updates the order prior to the create_order hook (thread 1) process.
+        // This ensures that the returned order in the create_order hook is the latest updated order.
+        $order = $this->cartHelper->getOrderByQuoteId($quote->getId());
```
danac-gs, line 1101: we should verify the payment method here, just in case

danac-gs (changes requested):

Helper/Order.php
johnmccombs1: @vitaliyreznikov I've updated the code. Could you please re-check?

Helper/Order.php
```diff
@@ -1115,6 +1116,14 @@ public function processNewOrder($quote, $transaction)
                 CreateOrder::E_BOLT_GENERAL_ERROR
             );
         }
+        // When the create_order hook (thread 1) takes a lot of time to execute and returns a timeout, the authorize/payment hook (thread 2) is sent to Magento.
+        // It updates the order prior to the create_order hook (thread 1) process.
+        // This ensures that the returned order in the create_order hook is the latest updated order.
+        $existingOrder =  $this->cartHelper->getOrderById($order->getId());
+        if ($existingOrder->getPayment() && $existingOrder->getPayment()->getMethod() == Payment::METHOD_CODE) {
+            $order = $existingOrder;
+        }
```
danac-gs, line 1125: otherwise some other payment method got in the way and we should throw exception. perhaps. rare case but seems like everything is possible.

johnmccombs1: @danac-gs  I've updated the change as your suggestion. Could you please re-check?

danac-gs (approved):

danac-gs (approved):

vitaliyreznikov (approved):

PR #910, opened 2020-09-02
Removed setting a reserved ID

njfaries:
# Description
To fix the issue MUJI was having with IDs out of order we want to not reserve the order ID so early in the process. This PR removes the setting of it and adjusts places it was used to retrieve orders to use different IDs

Fixes: M2P-220

#changelog Remove setting a reserved ID

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:


vitaliyreznikov (approved):

Helper/Order.php
```diff
@@ -1037,8 +1031,8 @@ public function dispatchPostCheckoutEvents($order, $quote)
             ]
         );
 
-        // Nullify bolt_reserved_order_id. Prevents dispatching more then once.
-        $quote->setBoltReservedOrderId(null);
+        // Set dispatched to be true. Prevents dispatching more then once.
+        $quote->setBoltDispatched(true);
```
danac-gs, line 1035: This wont work among subsequent hooks because it is not persistent anywhere. The old "flag" was added to the table here https://github.com/BoltApp/bolt-magento2/blob/master/Setup/UpgradeSchema.php#L47 so we could use it. You need to do a similar thing.

danac-gs (changes requested):
This is a huge change in functionality that breaks backward compatibility. Legacy flow wont be supported anymore.

1. Do we know how many merchants are not using pre-auth?
2. Are we 100% sure that we want to do it?

If not, we can easily make this conditional and support both ways. There's a `is_pre_auth` coniguration flag here https://github.com/BoltApp/bolt-magento2/blob/master/etc/adminhtml/system.xml#L127 that we can utilize.

Moreover, the miss mentioned in my other comment indicates that this was not thoroughly tested. This type of change requires running all different web hook type test cases (from production readiness spreadsheet). With regular checkout and PPC. Merging this to master is risky. If we rush things out we would spend more time fixing them that building them.

vitaliyreznikov:
I've fixed a few small issues (for non-pre-auth flow mostly) and tested. It works with pre-auth and non-pre-auth flow. It works with PPC, both logged-in and guest flow. I don't think we should postpone this PR.
Non-pre-auth email we discussed in slack https://boltpay.slack.com/archives/CQ864R9JQ/p1597361439300500 can't be an issue because bolt doesn't send an email if display_id is empty.

@njfaries We checked logic and tested independently. Logic is great and I think we are ready to go ahead once you address Daniel's comment (add a new field for a quote) and fix the test.

vitaliyreznikov (dismissed):
LGTM

vitaliyreznikov (approved):

matt-thomason (approved):

vitaliyreznikov (approved):

matt-thomason (approved):

PR #937, opened 2020-09-15
M2P-254 Add user group id into bolt cart under feature switch

vitaliyreznikov:
# Description
Add user group id into bolt cart under feature switch

Fixes: [M2P-254](https://boltpay.atlassian.net/browse/M2P-254)

#changelog M2P-254 Add user group id into bolt cart under feature switch

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -1528,6 +1528,17 @@ class Cart extends AbstractHelper
     }
 
     /**
+     * Return user group id. "0" for guest users
+     *
+     */
+    private function getUserGroupId() {
+        if (!$this->customerSession->isLoggedIn()) {
+            return "0";
+        }
+        return $this->customerSession->getCustomer()->getGroupId();
+    }
+
+    /**
      * Get the address for totals calculation based on the quote physical / virtual type
      *
      * @param Quote $quote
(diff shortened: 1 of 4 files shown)


Helper/Cart.php
```diff
@@ -1527,6 +1527,17 @@ private function getProductToGetImageForGroupedItem ($item)
         return $childProduct;
     }
 
+    /**
+     * Return user group id. "0" for guest users
```
matt-thomason, line 1531: Is `0` as guest our creation or a default for Magento?

matt-thomason (approved):

Helper/Cart.php
vitaliyreznikov: Like our creation. Magento returns "1" for guest users and it's weird.

matt-thomason: Maybe it is worth explaining this in a comment then. There might be weird behaviour down the road one day and knowing this is our creation and not Magento's convention in a comment could be helpful.

matt-thomason: In fact, is there a reason we don't use `1`? We know it won't collide with another 1 since Magento reserves it for guest

vitaliyreznikov: No, Magento uses "1" for guest users although "1" is real group_id. So in some cases we need to check if user logged in or not to know what this "1" mean. At least what I read in StackExchange today.
I will double-check this and update the comment.

danac-gs (dismissed):

Helper/Cart.php
vitaliyreznikov: In Magento we use something like
```
 if ($customerSession->isLoggedIn()) {
      echo 'Customer Group Id: ' .  $customerSession->getCustomer()->getGroupId();
  }
```
So group_id started from 1 and make sense for logged in users only.

vitaliyreznikov: Fixed comment. Could you please check @matt-thomason

matt-thomason (approved):
Thanks for the discussion @vitaliyreznikov 👍

PR #975, opened 2020-10-02
M2P-262 - Resolve issue with Aheadworkds_Sarp2

danac-gs:
# Description
Backoffice key is dynamically loaded on Bolt order create [here](https://github.com/BoltApp/bolt-magento2/blob/a396f7342eff78e44aef056f4c866b506f603059/view/adminhtml/templates/boltpay/js/replacejs.phtml#L395). It is expected that, at the time of the order creation, an appropriate store is already chosen.

This was not the case for this client’s (Andrew Christian) store for the following reason: Magento 2, by default, performs some checks for each payment method before it can be rendered. These checks are defined [here](https://github.com/magento/magento2/blob/7ddd47e87e217ba20b883e9ee22e5e3ef618e6df/app/code/Magento/Sales/etc/di.xml#L154). Aheadworks_Sarp2 attempted to add a check for recurring payments, unfortunately they added it in the adminhtml area as opposed to global, where Magento has defined the default ones.

This will, instead of adding, completely replace the default checks with Aheadworks’. This behavior is defined [here](https://devdocs.magento.com/guides/v2.4/extension-dev-guide/build/di-xml-file.html): “When Magento loads a new configuration at a later time, either by a more specific scope or through code, then any array definitions in the new configuration will replace the loaded config instead of merging.”.

This means that the Bolt checkout button was present on the order create page long before we expected (on creation page load), and triggered order creation.

Fixes: [M2P-262](https://boltpay.atlassian.net/browse/M2P-262)

#changelog M2P-262 - Resolve issue with Aheadworkds_Sarp2

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Observer/Adminhtml/ActionPredispatch.php
+++ b/Observer/Adminhtml/ActionPredispatch.php
@@ -1,0 +1,78 @@
+<?php
+/**
+ * Bolt magento2 plugin
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ *
+ * @category   Bolt
+ * @package    Bolt_Boltpay
+ * @copyright  Copyright (c) 2017-2020 Bolt Financial, Inc (https://www.bolt.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+namespace Bolt\Boltpay\Observer\Adminhtml;
+
+use Magento\Framework\Event\ObserverInterface;
+use Magento\Framework\Event\Observer;
+
+/**
+ * Observer for controller_action_predispatch
+ * @see \Magento\Framework\App\Action\Action::dispatch
+ */
+class ActionPredispatch implements ObserverInterface
+{
+    /**
+     * @var \Bolt\Boltpay\Model\EventsForThirdPartyModules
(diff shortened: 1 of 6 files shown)


Observer/Adminhtml/ActionPredispatch.php
```diff
@@ -0,0 +1,78 @@
...
+     * Redirects the predispatch event to the {@see \Bolt\Boltpay\Model\EventsForThirdPartyModules}
+     *
+     * @param Observer $observer
+     * @return void
+     */
+    public function execute(Observer $observer)
+    {
+        /** @var \Magento\Framework\App\RequestInterface $request */
+        $request = $observer->getData('request');
+        $this->eventsForThirdPartyModules->dispatchEvent(
```
vitaliyreznikov, line 53: @danac-gs Looks like you added few observers but only one listener. Did I miss something?

danac-gs: @vitaliyreznikov 
The additional event dispatches along with the one in use are there for future use.

The intention was to port the [core Magento 2 pre-dispatch functionality](https://github.com/magento/magento2/blob/5a9fcab6a35b6b39e10f47ca1ea22dfca4b794a4/lib/internal/Magento/Framework/App/Action/Plugin/EventDispatchPlugin.php#L108-L116) to our Events For Third-Party Modules. You can see that 3 events are dispatched there too, with varying degrees of specificity. 

Since we don't have area support in our implementation there is also a `adminhtml` prefix to each of them.
The two not used ones can, of course, be removed.

vitaliyreznikov (approved):

PR #994, opened 2020-10-12
MA-34: Create dummy debug save endpoint

meganschnedar-bolt:
# Description
MA-34: Creating a new endpoint that takes information from debug and sets then sets config settings contained within the debug information.

Fixes: (link Jira ticket) https://boltpay.atlassian.net/browse/MA-34

#changelog MA-34: Create dummy debug save endpoint

# Type of change

- [X ] New feature (change which adds functionality)

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Config.php
+++ b/Helper/Config.php
@@ -1785,7 +1785,7 @@ class Config extends AbstractHelper
      * @param string $settingName
      * @param mixed $settingValue
      */
-    public function setConfigSetting($settingName, $settingValue = null)
+    public function setConfigSetting($settingName, $settingValue = null, $storeId = null)
     {
         $currentValue = $this->getScopeConfig()->getValue(
             self::CONFIG_SETTING_PATHS[$settingName],
@@ -1797,7 +1797,8 @@ class Config extends AbstractHelper
             $this->getConfigWriter()->save(
                 self::CONFIG_SETTING_PATHS[$settingName], 
                 $settingValue, 
-                ScopeInterface::SCOPE_STORE
+                ScopeInterface::SCOPE_STORE,
+                $storeId
             );
         }
     }
--- a/Helper/Config.php
+++ b/Helper/Config.php
@@ -27,6 +27,7 @@ use Magento\Store\Model\ScopeInterface;
 use Magento\Store\Model\Store;
 use Magento\Directory\Model\RegionFactory;
 use Magento\Framework\Composer\ComposerFactory;
+use Magento\Framework\App\Config\Storage\WriterInterface;
 
 /**
  * Boltpay Configuration helper
@@ -381,6 +382,15 @@ class Config extends AbstractHelper
         \Magento\Catalog\Model\Product\Type::TYPE_BUNDLE,
         \Magento\GroupedProduct\Model\Product\Type\Grouped::TYPE_CODE
     ];
+
+    /**
+     * Map of human-readable config names to their XML paths
+     */
+    const CONFIG_SETTING_PATHS = [
+        "publishable_key_checkout" => self::XML_PATH_PUBLISHABLE_KEY_CHECKOUT 
+        # TODO: Fill in the rest of the config settings
+    ];
+
     /**
      * @var ResourceInterface
      */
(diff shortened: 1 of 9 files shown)


Model/Api/DebugSave.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+    
+    /**
+     * This request parse the debug info and map the configuration to the debug info's configuration.
+     *
+     * @return void
+     * @param mixed $debug_info
+     * @api
+     */
+    public function execute(
```
vitaliyreznikov, line 45: We need to verify that request is signed - the same way as in other API calls

Model/Api/DebugSave.php
```diff
@@ -0,0 +1,58 @@
...
+        $debug_info
+    ) {
+        # parse debug_info into array
+        $debug_info_decoded = json_decode($debug_info, true);
```
vitaliyreznikov, line 49: Could you please add error handling? try catch block with return error answer in catch I think enough for now.

Model/Api/DebugSave.php
```diff
@@ -0,0 +1,58 @@
...
+
+        # extract bolt config data
+        $config_data = $debug_info["pluginConfigSettings"];
+
+        # TODO: Loop through config settings 
+        # Don't set "api_key" or "signing_secret" since their values are not displayed in the debug info
+        # $configHelper->setConfigSetting("configSettingName", configSettingValue)
```
vitaliyreznikov, line 56: @matt-thomason What are we doing with api_key and signing_secret? Do we want to update them via this API.
From my point of view this is OK to include them into debug answer.
We can include keys only if site works via https.

@meganschnedar-bolt Could you for now add other setting instead of key, so we make sure that code works.

etc/di.xml
```diff
@@ -62,6 +62,9 @@
 
     <!-- For debug endpoint -->
     <preference for="Bolt\Boltpay\Api\DebugInterface" type="Bolt\Boltpay\Model\Api\Debug" />
+
+    <!-- For debug save endpoint -->
+    <preference for="Bolt\Boltpay\Api\DebugSaveInterface" type="Bolt\Boltpay\Model\Api\DebugSave" />
```
vitaliyreznikov, line 67: We want to update settings so maybe we can call this API Bolt\Boltpay\Model\Api\UpdateSettings

Helper/Config.php
```diff
@@ -1761,6 +1773,21 @@ public function getAllConfigSettings()
...
+    /**
+     * Set config setting name to the given value
+     *
+     * @param string $settingName
+     * @param mixed $settingValue
+     */
+    public function setConfigSetting($settingName, $settingValue = null)
+    {
+        $this->configWriter->save(
+            self::CONFIG_SETTING_PATHS[$settingName],
```
vitaliyreznikov, line 1785: I am not sure but can you consider saving only if the setting really changed? Updating setting even if with the same value can take time and trigger same events.

Model/Api/DebugSave.php
meganschnedar-bolt: I added `$this->hookHelper->preProcessWebhook($this->storeManager->getStore()->getId());` which was used in the debug endpoint. Is this correct?

vitaliyreznikov (commented):
Looks very well!

1) Could you please add tests?
2) Should we return something if we updated setting properly
3) Do we have the opportunity to clear Magento cache (but it can be subject of another PR)

meganschnedar-bolt:
@vitaliyreznikov I could return a success response or error response? Similar to what we do in `UpdateCart`.

vitaliyreznikov:
Yes, I think we can use approach from update_cart endpoint

send success answer
https://github.com/BoltApp/bolt-magento2/blob/652b5b5b7f2b086800143c63505580c9cf2e31ec/Model/Api/UpdateCart.php#L227

send error answer
https://github.com/BoltApp/bolt-magento2/blob/652b5b5b7f2b086800143c63505580c9cf2e31ec/Model/Api/UpdateCart.php#L230

Helper/Config.php
```diff
@@ -381,6 +382,15 @@ class Config extends AbstractHelper
         \Magento\Catalog\Model\Product\Type::TYPE_BUNDLE,
         \Magento\GroupedProduct\Model\Product\Type\Grouped::TYPE_CODE
     ];
+
+    /**
+     * Map of human-readable config names to their XML paths
+     */
+    const CONFIG_SETTING_PATHS = [
+        "publishable_key_checkout" => self::XML_PATH_PUBLISHABLE_KEY_CHECKOUT 
+        # TODO: Fill in the rest of the config settings
```
usmanismail-bolt, line 391: Does this todo need to be cleaned up?

vitaliyreznikov (approved):
Nice job

PR #1011, opened 2020-10-23
M2P-331 Allow to create order on the second payment attempt

vitaliyreznikov:
If users try to pay a second time after an unsuccessful attempt Magento plugin can return an error "Order is in pending payment. Waiting for the hook update.".It happens if Magento didn't receive failed payment hook yet.

It happens quite often (few times per day) because hooks work asynchronously, use the queue. We can't expect hook work instantly.

Order in pending payment status can mean another thing: the order was irreversibly rejected before auth (blacklist e.g.) but the hook wasn't sent yet.
In this case, we will delete the order as well. This is not the best behavior, but it's not an issue: eventually, the merchant will have an order with canceled status, the same as before this PR.
Down the road, we are going to change behavior on the server side and don't create orders in the cart if payment rejected before auth.

**Solution:**
If we see created order in pending payment status we can delete it and create a new one

Fixes: [M2P-331](https://boltpay.atlassian.net/browse/M2P-331)

#changelog M2P-331 Allow to create order on the second payment attempt

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -1097,14 +1097,18 @@ class Order extends AbstractHelper
             }
 
             if ($order->getState() === OrderModel::STATE_PENDING_PAYMENT) {
-                throw new BoltException(
-                    __(
-                        'Order is in pending payment. Waiting for the hook update. Quote ID: %1',
-                        $quote->getId()
-                    ),
-                    null,
-                    CreateOrder::E_BOLT_GENERAL_ERROR
+                // Order for same quote is created, it is in pending payment status
+                // and we try to create a new one
+                // It means order was created for unsuccessful payment attempt and wasn't delete yet.
+                // We can safely delete it
+                $incrementId = $order->getIncrementId();
+                $immutableQuoteId = $this->cartHelper->getImmutableQuoteIdFromBoltOrder($transaction->order);
+                $this->bugsnag->notifyError(
+                    "Delete order in pending payment",
+                    "OrderNo: {$incrementId}, quote id: ".$quote->getId()
                 );
+                $this->deleteOrderByIncrementId($incrementId, $immutableQuoteId);
+                return false;
             }
 
             if ($this->hasSamePrice($order, $transaction)) {
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -1070,6 +1070,25 @@ class Order extends AbstractHelper
     }
 
     /**
+     * Cancel or detete the failed payment order depending on settings via feature switches
+     *
+     * @param string $displayId
+     * @param string $immutableQuoteId
+     * @return string log message
+     * @throws \Magento\Framework\Exception\AlreadyExistsException
+     */
+    public function deleteOrCancelFailedPaymentOrder($display_id, $immutableQuoteId)
+    {
+        if ($this->featureSwitches->isCancelFailedPaymentOrderInsteadOfDeleting()) {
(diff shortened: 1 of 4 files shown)


danac-gs (commented):
@vitaliyreznikov 
This flow is there for a reason and should probably stay as is. The main reason is to not allow further transactions in cases when the first one was suspectedly fraudulent. There's no other way to know that but to wait for the first (asynchronous) hook. That's why the failed_payment and irreversible_rejected (blacklisted cards) hooks are to be sent immediately.
Moreover, we are moving away from deleting the orders since that is not natural Magento behavior and it causes issues with some ERP systems.
cc @matt-thomason @njfaries @dumega @johnmccombs1

vitaliyreznikov:
@danac-gs I believe we should fix this: this is an real issue affecting decent amount of users.
But you are right, I haven't considered the case when the first attempt was irreversibly rejected.
Will check this case.

vitaliyreznikov:
@danac-gs I think we can go ahead with this code.
We have three reasons to decline the order and move it into rejective_irreversible status:
1) email in blacklist
2) decline before authorization
3) decline during authorization

In (3) the user is redirected to the success page, so they don't have a chance for a second payment attempt.
So we are speaking about (1) and (2)

In the common case, when user's email is in blacklist, then on the second attempt plugin returns pre-auth error.
But after this fix, rarely we will return success answer (if user makes a new attempt before the plugin receives rejective_irreversible). But it changes nothing:
- For user: although pre-auth return trues, the user will see the error again because of blacklist or decline before auth
- For merchant: although we deleted one order and created a new one, eventually, a new hook comes, and we will have one canceled order.

danac-gs:
@vitaliyreznikov

If we are about to solve this in a best possible way, it should not all be done on the plugin end but in the core. The issue is that the pre-auth order creation is called for credit cards that are blacklisted, and we know (on Bolt side) in advance that they would need to be canceled. This blacklist check should be done before the pre-auth call and not after. That would eliminate one case and left us dealing with failed payment only.

As far as your solution that involves order deletion, that would cause issues with ERP systems and gaps in order sequence that we are trying to prevent (by eliminating reserving order id in the recent past and removing the deletion in this ongoing task https://boltpay.atlassian.net/browse/M2P-297 - you ca find there more details and discussion regarding the above issues).

As I see it, this solves one problem but generates (replaces it with) another one. On the other hand, sending failed payment hook immediately, if set, minimizes the possibility of this issue to happen. It have been working like that for months, probably years by now.

Since we have different ways of looking at the subject (again :smiley: ) I will not approve nor reject this PR and let others decide.

vitaliyreznikov:
> This blacklist check should be done before the pre-auth call and not after.

Will discuss with team if this possible / has it's own disadvantages

> removing the deletion in this ongoing task https://boltpay.atlassian.net/browse/M2P-297

We can merge this one after M2P-297 to reflect changes properly

> the reported issues might be related to misconfiguration

I believe this is a real, reproducible issue. It happens because we need some time to deliver the hook.

vitaliyreznikov:
I've updated the description. Although behavior with orders rejected before auth may be changed, it doesn't important. We are ok to fix this issue.

matt-thomason:
@vitaliyreznikov had a discussion with @nirajjayant and this seems to be fine from a Risk perspective. I think this is fine going out as is.

matt-thomason (approved):

PR #1030, opened 2020-11-02
M2P-359 Correct the total amount calculation logic when building Bolt cart

johnmccombs1:
# Description
Currently, there are some rounding issues if the customer adds a product with a large qty. So this PR is to correct the total amount calculation logic when building Bolt cart

FYI, this is the same as the way that we handle on M1. See here: https://github.com/BoltApp/bolt-magento1/blob/master/app/code/community/Bolt/Boltpay/Model/BoltOrder.php#L150

Fixes: https://boltpay.atlassian.net/browse/ON-389

#changelog Correct the total amount calculation logic when building Bolt cart

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -1341,7 +1341,7 @@ class Cart extends AbstractHelper
                 $unitPrice   = $item->getCalculationPrice();
                 $itemTotalAmount = $unitPrice * $item->getQty();
 
-                $roundedTotalAmount = CurrencyUtils::toMinor($itemTotalAmount, $currencyCode);
+                $roundedTotalAmount = CurrencyUtils::toMinor($unitPrice, $currencyCode) * round($item->getQty());
 
                 // Aggregate eventual total differences if prices are stored with more than 2 decimal places
                 $diff += CurrencyUtils::toMinorWithoutRounding($itemTotalAmount, $currencyCode) -$roundedTotalAmount;


danac-gs (approved):

Helper/Cart.php
```diff
@@ -1341,7 +1341,7 @@ function ($item) use ($imageHelper, &$totalAmount, &$diff, $storeId, $currencyCo
                 $unitPrice   = $item->getCalculationPrice();
                 $itemTotalAmount = $unitPrice * $item->getQty();
 
-                $roundedTotalAmount = CurrencyUtils::toMinor($itemTotalAmount, $currencyCode);
+                $roundedTotalAmount = CurrencyUtils::toMinor($unitPrice, $currencyCode) * round($item->getQty());
```
vitaliyreznikov, line 1344: @johnmccombs1 Do you know why we need to round item Qty? I see we have the same code below.

vitaliyreznikov (approved):

Helper/Cart.php
johnmccombs1: @vitaliyreznikov On M2, we rounded the quantity when sending to Bolt. See this: https://github.com/BoltApp/bolt-magento2/blob/master/Helper/Cart.php#L1361 So we should also round item qty when calculating the total amount.

PR #1034, opened 2020-11-06
M2P-297 - On failed payment hook change the order status to cancel instead of deleting it

danac-gs:
# Description
Deleting orders (on failed payment hook) confuses ERP platforms.
Both failed payment and blacklisted transactions are now canceled.
In the failed_payment case, the quote_id is removed from the order record, thus allowing the quote to be used in subsequent payment attempt. So, after failed payment hook we allow crating a new order for the same cart with another cc.
On irreversible_rejected one we just do the regular cancel, the quote_id stays, the further checkout retries are ignored.

Fixes: https://boltpay.atlassian.net/browse/M2P-297

#changelog On failed payment hook change the order status to cancel instead of deleting it

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -1195,7 +1195,7 @@ class Order extends AbstractHelper
      */
     protected function deleteOrder($order)
     {
-        $this->eventsForThirdPartyModules->dispatchEvent("beforeDeleteOrder", $order);
+        $this->eventsForThirdPartyModules->dispatchEvent("beforeFailedPaymentOrderSave", $order);
         try {
             $order->cancel()->save()->delete();
         } catch (\Exception $e) {
@@ -2355,7 +2355,7 @@ class Order extends AbstractHelper
             $parentQuote->setBoltCheckoutType(CartHelper::BOLT_CHECKOUT_TYPE_PPC);
             $this->cartHelper->quoteResourceSave($parentQuote->setIsActive(false));
         }
-        $this->eventsForThirdPartyModules->dispatchEvent("beforeDeleteOrder", $order);
+        $this->eventsForThirdPartyModules->dispatchEvent("beforeFailedPaymentOrderSave", $order);
 
         $order->addData(['quote_id' => null]);
         $order->addCommentToStatusHistory(__('BOLTPAY INFO :: Order was canceled due to Processor rejection'));
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -36,6 +36,7 @@ use Magento\Quote\Model\Quote;
 use Magento\Quote\Model\Quote\Address;
 use Magento\Quote\Model\QuoteManagement;
 use Magento\Sales\Api\Data\OrderInterface;
+use Magento\Sales\Api\OrderManagementInterface;
 use Magento\Sales\Api\OrderRepositoryInterface as OrderRepository;
 use Magento\Sales\Model\Order as OrderModel;
 use Magento\Sales\Model\Order\Email\Container\InvoiceIdentity as InvoiceEmailIdentity;
@@ -247,6 +248,11 @@ class Order extends AbstractHelper
     private $transactionInfo;
 
     /**
+     * @var OrderManagementInterface
+     */
+    private $orderManagement;
+
+    /**
      * Order constructor.
      * @param Context $context
      * @param Api $apiHelper
@@ -277,6 +283,7 @@ class Order extends AbstractHelper
      * @param CreditmemoFactory $creditmemoFactory
      * @param CreditmemoManagementInterface $creditmemoManagement
      * @param EventsForThirdPartyModules $eventsForThirdPartyModules
+     * @param OrderManagementInterface|null $orderManagement
      */
     public function __construct(
         Context $context,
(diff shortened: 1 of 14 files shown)


johnmccombs1:
@danac-gs Test failed. Can you fix it?

danac-gs:
@johnmccombs1 done

johnmccombs1 (approved):
LGTM

Helper/Order.php
```diff
@@ -338,6 +346,8 @@ public function __construct(
         $this->creditmemoFactory = $creditmemoFactory;
         $this->creditmemoManagement = $creditmemoManagement;
         $this->eventsForThirdPartyModules = $eventsForThirdPartyModules;
+        $this->orderManagement = $orderManagement
+            ?: \Magento\Framework\App\ObjectManager::getInstance()->get(OrderManagementInterface::class);
```
vitaliyreznikov, line 350: Are you sure Magento allows this type of ObjectManager using?
Why we need it?

Helper/Order.php
```diff
@@ -1072,10 +1082,15 @@ public function processExistingOrder($quote, $transaction)
         if ($order = $this->getExistingOrder(null, $quote->getId())) {
 
             if ($order->isCanceled()) {
+                if ($this->featureSwitches->isCancelFailedPaymentOrderInstadOfDeleting()
+                    && $order->getData('bolt_failed_payment')) {
```
vitaliyreznikov, line 1086: What if we just disabled the feature flag? Maybe the second condition is enough?

Model/Api/OrderManagement.php
```diff
@@ -239,8 +239,14 @@ private function saveUpdateOrder($reference, $type, $display_id, $storeId)
             $immutableQuoteId = $this->cartHelper->getImmutableQuoteIdFromBoltOrder($transaction->order);
 
             if ($type === 'failed_payment' || $type === 'failed') {
-                $this->orderHelper->deleteOrderByIncrementId($display_id, $immutableQuoteId);
-                $this->setSuccessResponse('Order was deleted: ' . $display_id);
+                if ($this->decider->isCancelFailedPaymentOrderInstadOfDeleting()) {
```
vitaliyreznikov, line 242: nit: isCancelFailedPaymentOrderInst**e**adOfDeleting

Helper/Order.php
```diff
@@ -2293,4 +2310,59 @@ public function getOrderStoreIdByDisplayId($displayId)
...
+        // reactivate session quote - the condiotion excludes PPC quotes
+        if ($parentQuoteId != $immutableQuoteId) {
+            $this->cartHelper->quoteResourceSave($parentQuote->setIsActive(true));
+        }
+        // reset PPC quote checkout type so it can be treated as active
+        if ($parentQuote->getBoltCheckoutType() == CartHelper::BOLT_CHECKOUT_TYPE_PPC_COMPLETE) {
+            $parentQuote->setBoltCheckoutType(CartHelper::BOLT_CHECKOUT_TYPE_PPC);
+            $this->cartHelper->quoteResourceSave($parentQuote->setIsActive(false));
+        }
+        $this->eventsForThirdPartyModules->dispatchEvent("beforeDeleteOrder", $order);
```
vitaliyreznikov, line 2362: Why we call beforeDeleteOrder?

Helper/Order.php
vitaliyreznikov: And the second question here. Looks like we can't find the order with 'bolt_failed_payment', because in the same moment when we set this field, we reset quote_id.
Could you check?

Helper/Order.php
dumega: It seems that we need to restores Amasty Giftcard balances, but I think we would better create another event for this, just separate from deleting order.

dumega: It seems that we need to restores Amasty Giftcard balances, but I think we would better create another event for this, just separate from deleting order.

Helper/Order.php
danac-gs: This was done just in case an integration overrides this helper's constructor. Magento does it all the time when introducing new attribute in constructor on class change / update: 

https://github.com/magento/magento2/blob/9544fb243d5848a497d4ea7b88e08609376ac39e/lib/internal/Magento/Framework/App/PageCache/Kernel.php#L84

https://github.com/magento/magento2/blob/89d3cda99975c0db2967165311184be84d26e9b4/lib/internal/Magento/Framework/MessageQueue/Consumer.php#L102

https://github.com/magento/magento2/blob/9544fb243d5848a497d4ea7b88e08609376ac39e/app/code/Magento/PageCache/Controller/Block.php#L59

More: https://github.com/magento/magento2/search?q=%3F%3A+%5CMagento%5CFramework%5CApp%5CObjectManager%3A%3AgetInstance%28%29-%3Eget

Helper/Order.php
danac-gs: @vitaliyreznikov Great catch. Actually the bolt_failed_payment flag is redundant. Will post an update soon.

Model/Api/OrderManagement.php
danac-gs: tnx, will update.

Helper/Order.php
danac-gs: @dumega correct https://github.com/BoltApp/bolt-magento2/blob/master/Model/EventsForThirdPartyModules.php#L61
I will do this in a separate PR.

Model/Api/OrderManagement.php
danac-gs: done

Helper/Order.php
danac-gs: done

vitaliyreznikov (approved):

PR #1045, opened 2020-11-18
M2P-372 Bugfix: Total mismatch with Mirasvit rewards

dumega:
# Description
Mirasvit rewards supports two different points spending styles Flexible (partial_point) and Fixed (full_point), and these two rules have different rounding logic on the available points number. 
For Flexible, it allows to apply float discount amount.
For Fixed, it only allows to apply integer discount amount.

This PR is to support these rules.

Fixes: https://boltpay.atlassian.net/browse/M2P-372

#changelog M2P-372 Bugfix: Total mismatch with Mirasvit rewards

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:


vitaliyreznikov (approved):

ThirdPartyModules/Mirasvit/Rewards.php
```diff
@@ -268,19 +268,52 @@ private function getMirasvitRewardsAmount($quote)
                 $websiteId = $quote->getStore()->getWebsiteId();
                 $rules = $this->mirasvitRewardsSpendRulesListHelper->getRuleCollection($websiteId, $customer->getGroupId());
                 if ($rules->count()) {
+                    $totalAmount = 0;
+                    $totalPoints = 0;
+        
+                    $mockSubtotal = 999999999;
```
johnmccombs1, line 274: Can you explain why set the value to 999999999 here?

dumega: @johnmccombs1 Cause we always try to apply full available rewards amount to Bolt cart, we create a mock subtotal large enough to implement this purpose.

dumega:
@vitaliyreznikov @johnmccombs1 

Update: I add support for Mirasvit_Rewards multiple spending rules

danac-gs (approved):
LGTM (although a bit complicated, but if it works it's OK)

ThirdPartyModules/Mirasvit/Rewards.php
vitaliyreznikov: @dumega Can we create a class constant instead of a variable for this number?

vitaliyreznikov (commented):
This approach looks so complicated. Do I understand correctly you copied some code from the 3p module?
Any chance we can call this code instead of copying?

dumega:
@vitaliyreznikov `function isRuleValid` and `function calcPointsPerRule` are both private functions in the 3rd-party module, and `function calMaxSpendAmount`. `function getCartPointsRange` and `function calcPointsPerRule` all have our own custom logic, so we can not directly call them from the 3rd-party module

ThirdPartyModules/Mirasvit/Rewards.php
dumega: @vitaliyreznikov Done

vitaliyreznikov:
@dumega Got it! Could you please add comments about this into each phpdoc block? Like this function is copied from class ... because it is private in the class.

vitaliyreznikov (commented):
@dumega Got it! Could you please add comments about this into each phpdoc block? Like this function is copied from class ... because it is private in the class.

dumega:
@vitaliyreznikov  Done

danac-gs (approved):
LGTM. One note: could we use reflection or closures for calling private methods instead of copying them?

vitaliyreznikov (approved):

johnmccombs1 (approved):

PR #1046, opened 2020-11-18
M2P-371 Remove zero discount text from shipping option label

dumega:
# Description
In some certain contexts, the shipping discounts amount could be some very small values like 0.00002, then the Bolt plugin collect it and covert it into [$0.00 discount], cause priceHelper->currency round 0.00002 into 0.00.

Actually in m2, if the number is less than 0.005, it can be treated as zero.

image

Fixes: https://boltpay.atlassian.net/browse/M2P-371

#changelog Remove zero discount text from shipping option label

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Discount.php
+++ b/Helper/Discount.php
@@ -57,7 +57,7 @@ class Discount extends AbstractHelper
     const BOLT_DISCOUNT_CATEGORY_AUTO_PROMO = 'automatic_promotion';
     
     // In Magento 2, if the discount amount is less than 0.005, we can treat it as zero.
-    const ZERO_VALUE = 0.004;
+    const ZERO_VALUE = 0.00499999;
 
     /**
      * @var ResourceConnection $resource
--- a/Helper/Discount.php
+++ b/Helper/Discount.php
@@ -55,6 +55,9 @@ class Discount extends AbstractHelper
     const BOLT_DISCOUNT_CATEGORY_COUPON = 'coupon';
     const BOLT_DISCOUNT_CATEGORY_GIFTCARD = 'giftcard';
     const BOLT_DISCOUNT_CATEGORY_AUTO_PROMO = 'automatic_promotion';
+    
+    // In Magento 2, if the discount amount is less than 0.005, we can treat it as zero.
+    const ZERO_VALUE = 0.004;
 
     /**
      * @var ResourceConnection $resource
--- a/Helper/Discount.php
+++ b/Helper/Discount.php
@@ -56,8 +56,9 @@ class Discount extends AbstractHelper
     const BOLT_DISCOUNT_CATEGORY_GIFTCARD = 'giftcard';
     const BOLT_DISCOUNT_CATEGORY_AUTO_PROMO = 'automatic_promotion';
     
-    // In Magento 2, if the discount amount is less than 0.005, we can treat it as zero.
-    const ZERO_VALUE = 0.00499999;
+    // In Magento 2, 0.005 would be converted into 0.01 which is greater than 0 while 0.0049999 would be converted into 0,
+    // so we can treat 0.00499999 as the closest number to zero in M2.
+    const MIN_NONZERO_VALUE = 0.00499999;
 
     /**
      * @var ResourceConnection $resource
--- a/Helper/Discount.php
+++ b/Helper/Discount.php
@@ -57,8 +57,8 @@ class Discount extends AbstractHelper
     const BOLT_DISCOUNT_CATEGORY_AUTO_PROMO = 'automatic_promotion';
     
     // In Magento 2, 0.005 would be converted into 0.01 which is greater than 0 while 0.0049999 would be converted into 0,
-    // so we can treat 0.00499999 as the closest number to zero in M2.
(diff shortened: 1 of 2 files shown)


danac-gs (approved):

Helper/Discount.php
```diff
@@ -55,6 +55,9 @@ class Discount extends AbstractHelper
     const BOLT_DISCOUNT_CATEGORY_COUPON = 'coupon';
     const BOLT_DISCOUNT_CATEGORY_GIFTCARD = 'giftcard';
     const BOLT_DISCOUNT_CATEGORY_AUTO_PROMO = 'automatic_promotion';
+    
+    // In Magento 2, if the discount amount is less than 0.005, we can treat it as zero.
+    const ZERO_VALUE = 0.004;
```
vitaliyreznikov, line 60: This is a little confusing. Why we have 0.005 in comment and 0.004 in code? Do we have corresponding code in Magento core? Could we have the same?

dumega: @vitaliyreznikov In comment it says if less than 0.005, then we can treat it as zero. But 0.005 is not the zero value, I setup a most approximate number to represent zero value. BTW I change this zero value to 0.0049999 now.

In M2 I can not find such a const or function (@danac-gs  correct me if there is), and I test with Magento\Framework\Pricing, it works as expected.

vitaliyreznikov: Why you selected 0.005? Just to understand, is it your choice or some M2 standard?

dumega: In M2, 0.005 would be converted into 0.01 which is greater than 0 while 0.0049999 would be converted into 0

danac-gs (dismissed):

Helper/Discount.php
vitaliyreznikov: Got it. Can we use ">=0.5" then?

dumega: Hmm so do you mean we set the constant "ZERO_VALUE" to 0.05? If so, actually it does not equal to 0

Helper/Discount.php
```diff
@@ -55,6 +55,9 @@ class Discount extends AbstractHelper
     const BOLT_DISCOUNT_CATEGORY_COUPON = 'coupon';
     const BOLT_DISCOUNT_CATEGORY_GIFTCARD = 'giftcard';
     const BOLT_DISCOUNT_CATEGORY_AUTO_PROMO = 'automatic_promotion';
+    
+    // In Magento 2, if the discount amount is less than 0.005, we can treat it as zero.
+    const ZERO_VALUE = 0.00499999;
```
adentong, line 60:
```suggestion
    const MIN_NONZERO_VALUE = 0.005;
```

Model/Api/ShippingMethods.php
```diff
@@ -780,7 +780,7 @@ public function getShippingOptions($quote, $addressData)
 
             $taxAmount = round(CurrencyUtils::toMinorWithoutRounding($shippingAddress->getTaxAmount(), $currencyCode) + $diff);
 
-            if ($discountAmount) {
+            if ($discountAmount > DiscountHelper::ZERO_VALUE) {
```
adentong, line 783:
```suggestion
            if ($discountAmount >= DiscountHelper::MIN_NONZERO_VALUE) {
```

adentong:
I think if we just rename the variable we can get around the awkward 0.00499999

dumega:
@adentong Thanks for the suggestion, I will change it.

Helper/Discount.php
```diff
@@ -55,6 +55,10 @@ class Discount extends AbstractHelper
     const BOLT_DISCOUNT_CATEGORY_COUPON = 'coupon';
     const BOLT_DISCOUNT_CATEGORY_GIFTCARD = 'giftcard';
     const BOLT_DISCOUNT_CATEGORY_AUTO_PROMO = 'automatic_promotion';
+    
+    // In Magento 2, 0.005 would be converted into 0.01 which is greater than 0 while 0.0049999 would be converted into 0,
+    // so we can treat 0.00499999 as the closest number to zero in M2.
+    const MIN_NONZERO_VALUE = 0.00499999;
```
adentong, line 61:
```suggestion
    const MIN_NONZERO_VALUE = 0.005;
```

Model/Api/ShippingMethods.php
```diff
@@ -780,7 +780,7 @@ public function getShippingOptions($quote, $addressData)
 
             $taxAmount = round(CurrencyUtils::toMinorWithoutRounding($shippingAddress->getTaxAmount(), $currencyCode) + $diff);
 
-            if ($discountAmount) {
+            if ($discountAmount > DiscountHelper::MIN_NONZERO_VALUE) {
```
adentong, line 783:
```suggestion
            if ($discountAmount >= DiscountHelper::MIN_NONZERO_VALUE) {
```

adentong:
since you decided to change the variable name, the value should also get updated to 0.005. otherwise the meaning of the name doesn't actually match the value.

adentong (approved):

danac-gs (approved):

PR #1047, opened 2020-11-19
ON-429 Show cart type and last four digits for all payment processors

vitaliyreznikov:
Now we show the last four cart digits for vantiv only.
But we have this info for adyen, alliance, braintree, nmi, and stripe as well.

After this fix, we show last4 for each case when we have it.

Fixes: [ON-429](https://boltpay.atlassian.net/browse/ON-429)

#changelog ON-429 Show cart type and last four digits for all payment processors

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Block/Info.php
+++ b/Block/Info.php
@@ -33,14 +33,12 @@ class Info extends \Magento\Payment\Block\Info
         $boltProcessor = $info->getAdditionalInformation('processor');
         $data = [];
         
-        if (empty($boltProcessor) || $boltProcessor == \Bolt\Boltpay\Helper\Order::TP_VANTIV) {
-            if ($ccType = $info->getCcType()) {
-                $data[(string)__('Credit Card Type')] = strtoupper($ccType);
-            }
+        if ($ccType = $info->getCcType()) {
+            $data[(string)__('Credit Card Type')] = strtoupper($ccType);
+        }
     
-            if ($ccLast4 = $info->getCcLast4()) {
-                $data[(string)__('Credit Card Number')] = sprintf('xxxx-%s', $ccLast4);
-            }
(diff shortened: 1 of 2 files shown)


Block/Info.php
```diff
@@ -33,14 +33,12 @@ protected function _prepareSpecificInformation($transport = null)
         $boltProcessor = $info->getAdditionalInformation('processor');
```
dumega, line 33: @vitaliyreznikov For for adyen, alliance, braintree, nmi, and stripe, what's the value of $info->getAdditionalInformation('processor'); ?

vitaliyreznikov: @dumega For now we have following processor's type:

adyen_gateway
affirm	
afterpay	
alliance_data
amazon_pay	
bolt	
braintree
nmi
paypal	
stripe
vantiv

njfaries (dismissed):

Block/Info.php
dumega: But for Paypal and afterpay , we do not have cc type and cc last 4.

vitaliyreznikov: Yes, but we still can use getter and receive empty value.
I don't want to have a full list of supported processors here because we need to update it each time we support new payment processor

dumega: Make sense. Then pleae remove `$boltProcessor = $info->getAdditionalInformation('processor');` as well, cause now it is useless in this function.

dumega (approved):

danac-gs (approved):

Block/Info.php
vitaliyreznikov: Good catch @dumega ! Done.

danac-gs (approved):

PR #1049, opened 2020-11-23
M2P-384 - Pre auth double check if order was created

danac-gs:
# Description
It's possible that Magento throws an exception during pre-auth call when order is already created.It's better for us to check if the order looks good and return a success answer to not block checkout for users

Fixes: https://boltpay.atlassian.net/browse/M2P-384

#changelog Pre auth double check if order was created

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Order.php
+++ b/Helper/Order.php
@@ -1140,6 +1140,39 @@ class Order extends AbstractHelper
 
     /**
      * @param Quote $quote
+     * @return false|OrderModel|null
+     * @throws \Exception
+     */
+    public function submitQuote($quote)
+    {
+        /** @var OrderModel $order */
+        try {
+            $order = $this->quoteManagement->submit($quote);
+        } catch (\Exception $e) {
+            $order = $this->getOrderByQuoteId($quote->getId());
+            if ($order && $order->getPayment() && $order->getPayment()->getMethod() === Payment::METHOD_CODE) {
+                $this->bugsnag->registerCallback(
+                    function ($report) use ($order) {
+                        $report->setMetaData(
+                            [
+                                'CREATE ORDER' => [
+                                    'order_id' => $order->getId(),
+                                    'order_increment_id' => $order->getIncrementId(),
+                                ]
+                            ]
+                        );
+                    }
+                );
+                $this->bugsnag->notifyException($e);
+            } else {
+                throw $e;
+            }
+        }
+        return $order;
+    }
+
+    /**
+     * @param Quote $quote
      * @param \stdClass $transaction
      * @return OrderModel
      * @throws BoltException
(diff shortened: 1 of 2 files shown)


Helper/Order.php
```diff
@@ -1138,6 +1138,39 @@ public function processExistingOrder($quote, $transaction)
...
+     * @throws \Exception
+     */
+    public function submitQuote($quote)
+    {
+        /** @var OrderModel $order */
+        try {
+            $order = $this->quoteManagement->submit($quote);
+        } catch (\Exception $e) {
+            $order = $this->getOrderByQuoteId($quote->getId());
+            if ($order && $order->getPayment() && $order->getPayment()->getMethod() === Payment::METHOD_CODE) {
```
dumega, line 1153: @danac-gs Just curious, is there any case `$order->getPayment()->getMethod() != Payment::METHOD_CODE` ?

dumega (approved):

Helper/Order.php
danac-gs: @dumega not sure but @johnmccombs1 put it there in case of something (even more) strange happens, as a safety measure, as done in other places (with a reason).

johnmccombs1 (approved):
LGTM

Helper/Order.php
vitaliyreznikov: @danac-gs What if we put this condition into the existing catch block?
In this case we can handle exception in all code after order creating, fo example in `$this->orderPostprocess($order, $quote, $transaction);` or whatever, no only in `$order = $this->quoteManagement->submit($quote);`

danac-gs: @vitaliyreznikov not sure exactly where do you mean we should place it. This is the exact place the issue happens, and is resolved right away. We need that order just after submit is called (in processNewOrder), it can not wait to be resolved at the execution end in a catch block because we need it earlier. Many (all) things depend on it from this point to the final catch block. (If I understood your request correctly)

vitaliyreznikov: I meant if we put the new check into catch blocks here
https://github.com/BoltApp/bolt-magento2/blob/67674aa9371750e5cd974bc2df6f72d0e08d3da8/Model/Api/CreateOrder.php#L233
maybe we are able to catch some other issues, not only exact the same we faced recently

danac-gs: The new code catches all exceptions, not only the one we faced. If there was an exception, but we still have the order created, we process that order normally (and send bugsnag notification). If we postpone order existence checking to that existing try catch block, which is at the end of the pre-auth flow, all the necessary order processing in between would be skipped. And we would need to go back and repeat the process from some point in the execution flow (right after quote submit). Although possible, it does not seem like an optimal way of solving this.

vitaliyreznikov (approved):

PR #1054, opened 2020-11-27
MA-86 Refactor coupon verification to return responses only in endpoint handler

njfaries:
# Description
Continuation of work started in https://github.com/BoltApp/bolt-magento2/pull/949 to move error responses to the endpoints and throw exception from internal methods.

Fixes: MA-86

#changelog [WIP] MA-86 Refactor coupon verification to return responses only in endpoint handler

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Model/Api/UpdateDiscountTrait.php
+++ b/Model/Api/UpdateDiscountTrait.php
@@ -414,24 +414,20 @@ trait UpdateDiscountTrait
         try{
             if(array_key_exists($couponCode, $discounts)){
                 if ($discounts[$couponCode] == 'coupon') {
-                    //throw \Exception
+                    //sends response
                     $this->removeCouponCode($quote);
                 } else if ($discounts[$couponCode] == DiscountHelper::BOLT_DISCOUNT_CATEGORY_STORE_CREDIT) {
                     //handles exceptions already, no return value
                     $this->eventsForThirdPartyModules->dispatchEvent("removeAppliedStoreCredit", $couponCode, $quote, $websiteId, $storeId);
                 } else {
-                    //throws BoltException
+                    //throws BoltException that will be caught in UpdateCart::execute()
                     $result = $this->verifyCouponCode($couponCode, $websiteId, $storeId);        
                     list(, $giftCard) = $result;
-                    //throws BoltException
+                    //sends response
                     $this->removeGiftCardCode($couponCode, $giftCard, $quote);
                 }
             } else {
-                throw new BoltException(
-                    __('Coupon code %1 does not exist!', $couponCode),
-                    null,
-                    BoltErrorResponse::ERR_CODE_INVALID
-                );
+                throw new \Exception(__('Coupon code %1 does not exist!', $couponCode));
             }
         } catch (\Exception $e) {
             $this->sendErrorResponse(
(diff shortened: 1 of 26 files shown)


njfaries:
The test `removeDiscount_removeGiftCard()` was failing silently before. After the changes introduced by this PR I noticed that it was never properly finishing the testing the method it was supposed to test; it was calling the `verifyCoupon` method I changed and originally returning explicitly false instead of throwing an exception. Now that `verifyCoupon` throws an exception the method under test, `removeDiscount`, was returning false from the `catch` statement when before it was just returning true. I updated the test to actually have a giftcard mock to remove.

danac-gs (changes requested):
You should also catch the specific BoltException here https://github.com/BoltApp/bolt-magento2/blob/master/Model/Api/UpdateDiscountTrait.php#L470

njfaries:
> You should also catch the specific BoltException here https://github.com/BoltApp/bolt-magento2/blob/master/Model/Api/UpdateDiscountTrait.php#L470

I believe this should actually be more similar to the response sending that I removed, right? I think the `removeDiscount` method should throw or rethrow a `boltException` to be handled in the `UpdateCart` class, in the same way the exception thrown by `verifyCouponCode` is being handled.

njfaries:
This has recursed into quite the PR. I've been changing helper functions to not send responses but throw `BoltExcepctions` instead, and have been hunting down other uses to update them to expect the exceptions as well. Tests still have yet to be updated. I believe this is generally the right track though to be on.

njfaries:
I've been making my way through trying to narrow the scope to just what is strictly necessary for the `validate` method in `DiscountCodeValidation.php`, while making sure that any methods that now throw exceptions that are used elsewhere still have the exceptions caught. In general they should, as the overlap is with other endpoint handlers (primarily the `UpdateCart` one), which have everything wrapped up in a large try-catch. I've only had to make a few minor tweaks to unit tests, specifically just the ones testing internal methods as they changed from expecting a response to expecting an exception. The handler tests all still pass, so I believe this is sound.  I'll be going through the checkout/discount flow manually in the morning to make sure and hopefully removing the WIP tag.

Test/Unit/Model/Api/UpdateDiscountTraitTest.php
```diff
@@ -976,42 +976,39 @@ public function applyingCouponCode_virtualImmutableQuote()
...
+
+    //     $quote = $this->getQuoteMock();
+
+    //     $exception = new \Exception('General exception');
+    //     $this->discountHelper->expects(self::once())->method('setCouponCode')
+    //         ->with($quote, self::COUPON_CODE)->willThrowException($exception);
+
+    //     $result = TestHelper::invokeMethod($this->currentMock, 'applyingCouponCode', [self::COUPON_CODE, $coupon, $quote]);
+
+    //     $this->assertFalse($result);
```
danac-gs, line 1010: remove commented code if redundant

danac-gs (approved):
lgtm

danac-gs (dismissed):

Model/Api/UpdateDiscountTrait.php
```diff
@@ -297,70 +281,49 @@ protected function applyingCouponCode($couponCode, $coupon, $quote, $addQuote =
...
             if (!in_array(0, $groupIds)) {
                 $this->logHelper->addInfoLog('Error: coupon requires login.');
-                $this->sendErrorResponse(
-                    BoltErrorResponse::ERR_CODE_REQUIRES_LOGIN,
-                    sprintf('The coupon code %s requires login', $couponCode),
-                    422,
-                    $quote
-                );
-                return false;
+                    throw new BoltException(
```
vitaliyreznikov, line 307: Do we need to fix indents here?

vitaliyreznikov (approved):
LGTM

Model/Api/UpdateDiscountTrait.php
njfaries: Good catch, fixing

vitaliyreznikov (approved):

PR #1072, opened 2020-12-18
M2P-416 Bugfix: Bolt PPC error if Aheadworks GiftCard is enabled

dumega:
# Description
We create a fresh quote for Bolt PPC, then after the data of this quote updates, it should be saved via repository.

Fixes: https://boltpay.atlassian.net/browse/M2P-416

#changelog Bugfix: Bolt PPC error if Aheadworks GiftCard is enabled

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -2354,9 +2354,11 @@ class Cart extends AbstractHelper
         }
 
         $quote->setIsActive(false);
-
+        $this->saveQuote($quote);
         $cart_data = $this->getCartData(false, '', $quote);
         $this->quoteResourceSave($quote);
+        $this->saveQuote($quote);
+        
         return $cart_data;
     }
 
--- a/Test/Unit/Helper/CartTest.php
+++ b/Test/Unit/Helper/CartTest.php
@@ -5366,6 +5366,8 @@ ORDER
         $this->productRepository->expects(static::once())->method('getById')->with(self::PRODUCT_ID)
             ->willReturn($this->productMock);
         $this->quoteMock->expects(static::once())->method('setIsActive')->with(false);
+        $this->quoteResource->expects(static::once())->method('save')->with($this->quoteMock);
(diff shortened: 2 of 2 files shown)


Helper/Cart.php
```diff
@@ -2354,9 +2354,11 @@ public function createCart($items, $metadata = null)
         }
 
         $quote->setIsActive(false);
-
+        $this->saveQuote($quote);
         $cart_data = $this->getCartData(false, '', $quote);
         $this->quoteResourceSave($quote);
+        $this->saveQuote($quote);
```
danac-gs, line 2360: Why is it necessary to call `$this->saveQuote($quote);` twice?

dumega: We first add products to the quote, if we do not call `$this->saveQuote($quote)`; before `$cart_data = $this->getCartData(false, '', $quote);`, then `$quoteRepository->get($quoteId)->getItemsCount` returns 0 which means the items are not saved properly into quoteRepository (\Magento\Quote\Api\CartRepositoryInterface).

Similarly, since the code in the function getCartData may update the quote, we need to call `$this->saveQuote($quote)` again to make sure everything is saved.

Then in my opinion, quoteResourceSave is to talk to the database while saveQuote is for run-time logic, so they are both needed at the end.

danac-gs (approved):

Helper/Cart.php
vitaliyreznikov: I am not sure it makes sense, but maybe we can move saveQuote into the exact place where are we updating the quote? Let me know what do you think

dumega: @vitaliyreznikov What do you mean "the exact place"? every time when the quote is updatd?

vitaliyreznikov: Ok, I just thought we didn't update the quote each time, but now I see that it's not the case. LGTM.

vitaliyreznikov (approved):

PR #1085, opened 2021-01-06
MA-84 Universal Api Support

njfaries:
# Description
PR for Universal API support in M2. Consolidates endpoints into single /bolt/boltpay/api endpoint. 

Implemented:
- order.shipping_and_tax
- order.create
- order.shipping
- order.tax
- cart.update
- cart.create
- discounts.code.apply

Originally I was writing `UniversalApi.php` based on the existing endpoints defined in `etc/webapi.xml` but learned about [these docs](https://docs.bolt.com/api_reference#tag/Merchant-API) so am pivoting to match them. A messy factor is that some existing endpoints make use of `$this->response->sendResponse` instead of returning an interface object, for example in the `order.create` flow. `discounts.code.apply` may need a whole new function specifically for it since there's nothing in v1 that does exactly and only that for re-use.

The cases for order.create, cart.create, and discounts.code.apply all still send their own responses instead of returning an interface. This is something that should be looked at resolving with a refactor that's beyond the scope of this ticket, so I've not included the work I started on that here.

Ticket for cleaning up interfaces: https://boltpay.atlassian.net/browse/MA-493

Fixes: https://boltpay.atlassian.net/browse/MA-84

#changelog Support Universal Api in M2

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Api/Data/UniversalApiResultInterface.php
+++ b/Api/Data/UniversalApiResultInterface.php
@@ -62,7 +62,7 @@ interface UniversalApiResultInterface
      * Set response data
      * 
      * @api
-     * @param Bolt\Boltpay\Api\Data\ShippingOptionsInterface $data
+     * @param mixed $data
      * @return $this
      */
     public function setData($data);
@@ -71,7 +71,7 @@ interface UniversalApiResultInterface
      * Get response data
      * 
      * @api
-     * @return Bolt\Boltpay\Api\Data\ShippingOptionsInterface $this
+     * @return mixed $this
      */
     public function getData();
 }
--- a/Api/Data/UniversalApiResultInterface.php
+++ b/Api/Data/UniversalApiResultInterface.php
@@ -18,11 +18,11 @@
 namespace Bolt\Boltpay\Api\Data;
 
 /**
- * Webhook result interface. Wrapping response object for universal api
+ * UniversalApi result interface. Wrapping response object for universal api
  * 
  * @api
  */
-interface WebhookResultInterface
+interface UniversalApiResultInterface
 {
     /**
      * Set event string
--- a/Api/Data/UniversalApiResultInterface.php
+++ b/Api/Data/UniversalApiResultInterface.php
@@ -71,7 +71,7 @@ interface UniversalApiResultInterface
      * Get response data
      * 
      * @api
-     * @return Bolt\Boltpay\Api\Data\ShippingOptionsInterface
+     * @return Bolt\Boltpay\Api\Data\ShippingOptionsInterface $this
      */
     public function getData();
 }
--- a/Api/Data/UniversalApiResultInterface.php
+++ b/Api/Data/UniversalApiResultInterface.php
@@ -71,7 +71,7 @@ interface UniversalApiResultInterface
      * Get response data
      * 
      * @api
-     * @return mixed $this
+     * @return mixed
      */
     public function getData();
 }
--- a/Api/Data/UniversalApiResultInterface.php
+++ b/Api/Data/UniversalApiResultInterface.php
@@ -62,8 +62,8 @@ interface UniversalApiResultInterface
      * Set response data
      * 
      * @api
-     * @param $data
-     * @return Bolt\Boltpay\Api\Data\ShippingOptionsInterface $this
+     * @param Bolt\Boltpay\Api\Data\ShippingOptionsInterface $data
+     * @return $this
      */
     public function setData($data);
 
--- a/Model/Api/OrderManagement.php
+++ b/Model/Api/OrderManagement.php
@@ -260,12 +260,20 @@ class OrderManagement implements OrderManagementInterface
             return;
         }
 
+        $request = $this->request->getBodyParams();
+        $this->logHelper->addInfoLog('management checking if request has [data]');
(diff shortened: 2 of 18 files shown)


Model/Api/UniversalApi.php
```diff
@@ -0,0 +1,250 @@
...
+                case "order.shipping_and_tax":
+                    //Returns ShippingOptionsInterface
+                    $this->result->setData(
+                        $this->shippingMethods->getShippingMethods(
+                            isset($data['cart']) ? $data['cart'] : null,
+                            isset($data['shipping_address']) ? $data['shipping_address'] : null
+                        )
+                    );
+                    break;
+                case "order.shipping":
```
vitaliyreznikov, line 171: Could you explain why we have differenct code for "order.shipping_and_tax" and "order.shipping"?

Model/Api/UpdateCart.php
```diff
@@ -223,6 +223,17 @@ public function execute($cart, $add_items = null, $remove_items = null, $discoun
 
             $result = $this->generateResult($immutableQuote);
 
+            //this is only true for universal api requests, in which case we want to return the result and
+            //let UniversalApi::execute deal with sending the response
+            if (isset($this->request->getBodyParams()['data'])) {
```
vitaliyreznikov, line 228: Should we have early return in this IF?

vitaliyreznikov (commented):
Looks good, added few questions, just to better understand the approach

Model/Api/OrderManagement.php
```diff
@@ -294,13 +321,20 @@ private function setSuccessResponse($message)
...
+     * @param boolean $isUniversal
      *
      * @throws \Exception
      */
-    private function handleCartCreateApiCall()
+    private function handleCartCreateApiCall($isUniversal = false)
     {
         $request = $this->request->getBodyParams();
 
+        if ($isUniversal)
```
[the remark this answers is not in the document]
njfaries: Good call will do

Model/Api/UniversalApi.php
njfaries: The [merchant API docs](https://docs.bolt.com/api_reference#tag/Merchant-API) specify different payloads and responses. It's probably possible to streamline this, as well as the "order.tax", but for now I was simply re-using the existing handlers as much as I could.

Model/Api/UpdateCart.php
njfaries: I had to remove it and start sending the response from here even for the universal endpoint due to issues with the DataObjectProcessor and how it would parse the objects in the returned UniversalApiInterface. This is something that will be addressed in MA-493, but here I simply forgot to update the comment.

danac-gs:
@njfaries [cart.create](https://docs.bolt.com/api_reference#operation/cart.create) response should also return auto-applied (couponless) discounts, that should be passed back to the checkout. It does so in the old API. I assume it behaves the same in Merchant API but it's not mentioned in the docs, so worth checking.

image

danac-gs:
@njfaries @vitaliyreznikov  I don't have access to: 

Ticket for cleaning up interfaces: https://boltpay.atlassian.net/browse/MA-493
Fixes: https://boltpay.atlassian.net/browse/MA-84

njfaries:
> @njfaries [cart.create](https://docs.bolt.com/api_reference#operation/cart.create) response should also return auto-applied (couponless) discounts, that should be passed back to the checkout. It does so in the old API. I assume it behaves the same in Merchant API but it's not mentioned in the docs, so worth checking.

I believe this should be included in the cart being sent back already, will double check.

njfaries:
> > @njfaries [cart.create](https://docs.bolt.com/api_reference#operation/cart.create) response should also return auto-applied (couponless) discounts, that should be passed back to the checkout. It does so in the old API. I assume it behaves the same in Merchant API but it's not mentioned in the docs, so worth checking.
> 
> I believe this should be included in the cart being sent back already, will double check.

Update: tested it in my m2 store and works, here's the call in postman: 
image @danac-gs

Api/Data/UniversalApiResultInterface.php
```diff
@@ -0,0 +1,77 @@
...
+     * @param mixed $data
+     * @return $this
+     */
+    public function setData($data);
+
+    /**
+     * Get response data
+     * 
+     * @api
+     * @return mixed $this
```
danac-gs, line 74:
```suggestion
     * @return mixed
```

danac-gs:
> > > @njfaries [cart.create](https://docs.bolt.com/api_reference#operation/cart.create) response should also return auto-applied (couponless) discounts, that should be passed back to the checkout. It does so in the old API. I assume it behaves the same in Merchant API but it's not mentioned in the docs, so worth checking.
> > 
> > 
> > I believe this should be included in the cart being sent back already, will double check.
> 
> Update: tested it in my m2 store and works, here's the call in postman:
> image @danac-gs

Cool. The https://docs.bolt.com/api_reference#operation/cart.create response should be updated to show that discounts may be included
cc @vitaliyreznikov @matt-thomason

danac-gs (approved):

PR #1091, opened 2021-01-15
M2P-418 - Improve session emulation by passing session identifiers and data in cart metadata

danac-gs:
# Description
Fix for a known issue where session emulation won't work when the site is load-balanced between multiple servers (when they don't use the same cache). This causes ID.me discount to not work properly with logged-in users. Full context in jira task linked below.

Fixes: https://boltpay.atlassian.net/browse/M2P-418

#changelog Improve session emulation by passing session identifiers and data in cart metadata

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:


danac-gs:
@vitaliyreznikov @matt-thomason @njfaries

One important note: This solution does not work with the Update Cart API. In update cart, as it seems, the cart metadata does not get passed along.

In that case the legacy (cache) solution is used, as you may see here https://github.com/BoltApp/bolt-magento2/pull/1091/files#diff-360e780b0fffe39e9474d40828d43c5aa1724813c74cfd628db1320cfebbd5d1R205. Which, of course, does not cover the issues similar to the one in this task. Luckily, Cariloha does not use the Update Cart API. And it shouldn't until this gets fixed in core.

Not sure if this is a known issue. If not, it should be added to the core team queue.

Helper/Cart.php
```diff
@@ -1702,6 +1700,17 @@ public function buildCartFromQuote($quote, $immutableQuote, $items, $placeOrderP
         //Store immutable quote id in metadata of cart
         $cart['metadata']['immutable_quote_id'] = $immutableQuote->getId();
 
+        $cart['metadata'][SessionHelper::ENCRYPTED_SESSION_ID_KEY] = $this->encryptMetadataValue(
```
vitaliyreznikov, line 1703: Can we save this under the feature switch?

danac-gs: @vitaliyreznikov we might do it but I am not sure if we should. This feature is an improvement of our previous implementation that, in case of missing metadata in update cart, falls back to the old functionality of using cache. So it covers more cases better, except one when we rely on the old code. When missing metadata issue gets fixed in core we should just remove the old code. Looks to me like overuse of feature switches but if you say so I will add it.

vitaliyreznikov: This is not because of update_cart, I agree we should include metadata into it.
I have two other reasons under my proposal:
- This is an important change so we need feature flags to have the ability to disable it in any cases.
- This is a lot of additional data that will be persisted into our database.

Do I understand correctly we need this PR only for one merchant right now?

danac-gs: @vitaliyreznikov fair points, will do.
The update is made per request / issues from one merchant, correct. However, it applies to all the stores / instances that work behind the load balancer.

danac-gs: @vitaliyreznikov done

Helper/Cart.php
```diff
@@ -1702,6 +1700,20 @@ public function buildCartFromQuote($quote, $immutableQuote, $items, $placeOrderP
         //Store immutable quote id in metadata of cart
         $cart['metadata']['immutable_quote_id'] = $immutableQuote->getId();
 
+        // Transmit session ID via cart metadata, making it available for session emulation in API calls
+        if ($this->deciderHelper->isAddSessionIdToCartMetadata()) {
```
vitaliyreznikov, line 1704: Just to double check. If it ok that only first expression is under feature switch?

Test/Unit/Helper/CartTest.php
```diff
@@ -2674,6 +2682,8 @@ public function getCartData_whenPaymentOnlyAndHasOrderPayload_returnsCartData()
         $quote->collectTotals()->save();
 
         TestUtils::setQuoteToSession($quote);
+
+        $sessionToMetadataSwitch = TestUtils::saveFeatureSwitch(Definitions::M2_ADD_SESSION_ID_TO_CART_METADATA);
```
vitaliyreznikov, line 2686: Should we run this (and other) tests with both options: with and without feature switch?

Helper/Cart.php
danac-gs: The other expression is for the case when we need a reliable way to transmit some 3rd party (session) data so we can reset / use it in API call session, for instance ID.me data mentioned in the task (uuid, group, subgroups). It's not commonly used, only when it's required by the supported plugins (ID.me in this case). It is a new thing, not supported previously.

Test/Unit/Helper/CartTest.php
danac-gs: We should. Will implement. Thanks.

Helper/Cart.php
vitaliyreznikov: So it will work even without feature switch if merchant doesn't use load balancer. Correct?

danac-gs: Correct.

vitaliyreznikov: Got it, thank you

Test/Unit/Helper/CartTest.php
danac-gs: done

vitaliyreznikov (approved):

PR #1101, opened 2021-01-26
MA-83 Universal Webhook Support

njfaries:
# Description
This PR goes along with https://github.com/BoltApp/bolt-magento2/pull/1085. This adds a webhook endpoint for V2 webhooks to be sent to and simply pulls the data needed to pass to the existing `OrderManagement::manage` method and returns either a success response or the error thrown by `OrderManagement::manage`

The inclusion of the result interface is a bit excessive for now as we are continuing to create and send the responses manually as opposed to returning the interface to the Magento REST handler. I'm including it here as in https://boltpay.atlassian.net/browse/MA-493 I hope to clean up the response handling to make better use of interfaces if we still cannot simply return interfaces to the Magento REST handler.

[Relevant slack discussion](https://boltpay.slack.com/archives/CQ864R9JQ/p1610377344014400)

Fixes: https://boltpay.atlassian.net/browse/MA-83

#changelog MA-83 Universal Webhook Support
Support for V2 Universal Webhook

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Model/Api/UniversalWebhook.php
+++ b/Model/Api/UniversalWebhook.php
@@ -50,6 +50,11 @@ class UniversalWebhook implements UniversalWebhookInterface
     private $logHelper;
 
     /**
+     * @var BoltErrorResponse
+     */
+    protected $errorResponse;
+
+    /**
      * @var Response
      */
     private $response;
@@ -59,6 +64,7 @@ class UniversalWebhook implements UniversalWebhookInterface
         Result $result,
         Bugsnag $bugsnag,
         LogHelper $logHelper,
+        BoltErrorResponse $errorResponse,
         Response $response
     )
     {
@@ -66,6 +72,7 @@ class UniversalWebhook implements UniversalWebhookInterface
         $this->result = $result;
         $this->bugsnag = $bugsnag;
         $this->logHelper = $logHelper;
+        $this->errorResponse = $errorResponse;
         $this->response = $response;
     }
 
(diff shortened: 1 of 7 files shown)


Model/Api/UniversalWebhook.php
```diff
@@ -0,0 +1,176 @@
...
+     */
+    public function execute(
+        $type = null,
+        $object = null,
+        $data = null
+    )
+    {
+        try {
+            // This just reuses the existing manage function by pulling the necessary data from the request.
+            $this->orderManagement->manage(
```
vitaliyreznikov, line 96: The purpose of the orderManagement class is to handle the specific API endpoint.
If we need the same code for handling two different API endpoints I'd rather prefer to move the code to the third class and call it from two API classes then call one API from another.

njfaries: While I don't disagree with your point here, this is already what's been done in https://github.com/BoltApp/bolt-magento2/pull/1085 now: calling existing handler methods from the universal handler. Do you think this is something that could be scoped as a refactor after getting the MVP version of the universal webhook and api out? Or should this block merging altogether and we set up tickets to refactor the universal API stuff as well?

vitaliyreznikov: I am ok to merge this, please add task to the backlog for refactoring

vitaliyreznikov (approved):
LGTM

PR #1112, opened 2021-02-03
M2P-466 Bugfix: Unable to remove items from Cart with Update Cart Endpoint

dumega:
# Description
If the customer tries to remove an item from the cart, there is no need to check its valid quantity in the store.

Fixes: https://boltpay.atlassian.net/browse/M2P-466

#changelog Bugfix: Unable to remove items from Cart with Update Cart Endpoint

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Model/Api/UpdateCartItemTrait.php
+++ b/Model/Api/UpdateCartItemTrait.php
@@ -130,6 +130,10 @@ trait UpdateCartItemTrait
             $qtyToCheck = ($updateItem['update'] == 'add')
                     ? (int)$quoteItem['quantity'] + (int)$updateItem['quantity']
                     : (int)$quoteItem['quantity'] - (int)$updateItem['quantity'];
+            
+            if ($updateItem['update'] == 'remove' && $qtyToCheck == 0) {
+                return true;
+            }
 
             // The module Magento_InventorySales has plugin to replace legacy quote item check,
             // so we send $qtyToCheck as $itemQty to follow its logic


vitaliyreznikov (approved):
Would you mind to add test for this case? Maybe even in separate PR, because I am going to create RC tag right now.

Model/Api/UpdateCartItemTrait.php
```diff
@@ -130,6 +130,10 @@ protected function verifyItemData($product, $updateItem, $quoteItem, $websiteId)
             $qtyToCheck = ($updateItem['update'] == 'add')
                     ? (int)$quoteItem['quantity'] + (int)$updateItem['quantity']
                     : (int)$quoteItem['quantity'] - (int)$updateItem['quantity'];
+            
+            if ($updateItem['update'] == 'remove' && $qtyToCheck == 0) {
```
danac-gs, line 134: Why do we need `$qtyToCheck == 0` check here?

dumega: if the customer want to remove the item from cart, the update-cart api hook would have the quantity value of request equal to quote item quantity, so $qtyToCheck == 0 means the same behavior. 
Then $qtyToCheck > 0 that's for updating quantity
$qtyToCheck < 0 I believe that's an exception,

dumega:
@vitaliyreznikov We already have unit test for this case. https://github.com/BoltApp/bolt-magento2/blob/2.18.1/Test/Unit/Model/Api/UpdateCartItemTraitTest.php#L384

Model/Api/UpdateCartItemTrait.php
danac-gs: gotcha. if we have, say, 2 units of some item in cart, we can currently only remove both (all) of them via the update cart request. not decreasing the quantity to 1. correct?

dumega: @danac-gs 
1. Under current logic, the update-cart does not allow empty the cart, so if you try to remove the last item, the Bolt modal would directly block that behavior.
2. We can decrease the quantity to 1.
3. If you set the quantity of item to 0 on the Bolt modal, it actually remove the item.

vitaliyreznikov: @dumega Question here: if we have 3 unit of the same item in cart and going to decrease it to 2. Should we check inventory in this case or not? Going to merge this PR, we can discuss this case separately.

vitaliyreznikov (approved):

Model/Api/UpdateCartItemTrait.php
dumega: @vitaliyreznikov we only ignore the quantity checking if $qtyToCheck == 0, cause that means the customer try to remove the item from cart, then for other cases, we would always process validation.

PR #1151, opened 2021-02-23
M2P-501 Create feature switch for customizable options support

dumega:
# Description
To enable/disable customizable options support by feature switch

Fixes: https://boltpay.atlassian.net/browse/M2P-501

#changelog Create feature switch for customizable options support

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:


Helper/Cart.php
```diff
@@ -1357,18 +1357,28 @@ function ($item) use ($imageHelper, &$totalAmount, &$diff, $storeId, $currencyCo
...
+                if ($this->deciderHelper->isCustomizableOptionsSupport()) {
+                    $itemProduct = $item->getProduct();
+                    $customizableOptions = $this->getProductCustomizableOptions($itemProduct);
+                    $itemSku = trim($item->getSku());
+                    
+                    if ($customizableOptions) {
+                        $itemSku = $this->getProductActualSkuByCustomizableOptions($itemSku, $customizableOptions);
+                    }
+                    
+                    $_product = $this->productRepository->get($itemSku);
```
vitaliyreznikov, line 1370: What if trying to catch error here and if so continue with `$itemProduct`?
What do you think about it?

vitaliyreznikov: Another point, I think we shouldn't search product by sku if we didn't change it. We just added additional SQL calls without benefits.
So something like
```
                    if ($customizableOptions) {
                        $itemSku = $this->getProductActualSkuByCustomizableOptions($itemSku, $customizableOptions);
                         $_product = $this->productRepository->get($itemSku);
                    }
```

vitaliyreznikov: Also, can we add productId to bugsnag record if we have any exceptions?

dumega:
@vitaliyreznikov 

> Why do we get the product by SKU instead of ID?

Please refer to this PR https://github.com/BoltApp/bolt-magento2/pull/1089, cause from the quote item, we can not get the actual product directly if it is variable product.

And I have updated the PR as you suggest

Helper/Cart.php
```diff
@@ -1360,18 +1360,36 @@ function ($item) use ($imageHelper, &$totalAmount, &$diff, $storeId, $currencyCo
...
+                                'product_id' => $item->getProductId(),
+                                'product_name' => $item->getName(),
+                                'product_sku' => $item->getSku()
+                            ]);
+                        });
+                        $this->bugsnag->notifyError('Could not retrieve product', $e->getMessage());
+                        $customizableOptions = null;
+                    }
+                }
+                if (empty($customizableOptions)) {
```
vitaliyreznikov, line 1386: I know this code works, but could you add `$customizableOptions = null;` before line 1365?

dumega: Done

vitaliyreznikov: Could you push your change?

dumega:
@vitaliyreznikov The snippet within `if ($this->deciderHelper->isCustomizableOptionsSupport()) {` block is not only for customizable options but also for fixing the update-cart endpoint to edit qty of variable product. So whether the $customizableOptions is empty or not, we woud always use $_product = $this->productRepository->get($itemSku); to retrieve the product.

vitaliyreznikov:
@dumega I am ok to merge this for now, but when you are online could you please explain why we need `$_product = $this->productRepository->get($itemSku); ` even if `$customizableOptions is empty`?

vitaliyreznikov (approved):

dumega:
@vitaliyreznikov That's for get the product which is variable type. If the quote item is variable product, $item->getProductId(); just returns its parent product id, and we have to retrieve its proper id via sku.

vitaliyreznikov:
Can we narrow down this assignment to `variable type` products only then?

danac-gs (approved):

Helper/Cart.php
```diff
@@ -1360,18 +1360,53 @@ function ($item) use ($imageHelper, &$totalAmount, &$diff, $storeId, $currencyCo
...
+                $customizableOptions = null;
                 $itemSku = trim($item->getSku());
+                $itemReference = $item->getProductId();
+                $itemName = $item->getName();
 
-                if ($customizableOptions) {
-                    $itemSku = $this->getProductActualSkuByCustomizableOptions($itemSku, $customizableOptions);
+                if ($this->deciderHelper->isCustomizableOptionsSupport()) {
+                    try {
+
```
vitaliyreznikov, line 1372: nit: remove empty line

vitaliyreznikov (approved):

Helper/Cart.php
dumega: Fixed

danac-gs (approved):

PR #1153, opened 2021-02-25
ON-648 Bugfix: The Cart doesn't contain the xxxx item

dumega:
# Description
When transfer data from parent quote to the immutable quote ([code link](https://github.com/BoltApp/bolt-magento2/blob/2.18.1/Helper/Cart.php#L1105)), our plugin directly set the 'items' property with the data of parent quote ([code link](https://github.com/magento/magento2/blob/fbfac3e51730abe2b79c6072544bc9742fcf7606/app/code/Magento/Quote/Model/Quote.php#L560)), as a result, when calling $immutableQuote->getItems();, actually it returns the items in parent quote instead of immutable quote, and eventually causes the exception __('The %1 Cart doesn\'t contain the %2 item.', $cartId, $itemId)([code link](https://github.com/magento/magento2/blob/fbfac3e51730abe2b79c6072544bc9742fcf7606/app/code/Magento/Quote/Model/Quote/Item/CartItemPersister.php#L70))

Solution: Refer to M2 core code ([code link](https://github.com/magento/magento2/blob/fbfac3e51730abe2b79c6072544bc9742fcf7606/app/code/Magento/Quote/Model/QuoteRepository/LoadHandler.php#L46)), we can clone this property by $child->setItems($child->getAllVisibleItems());

Fixes: https://boltpay.atlassian.net/browse/ON-648

#changelog Bugfix: The Cart doesn't contain the xxxx item

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -1129,6 +1129,11 @@ class Cart extends AbstractHelper
             $child->unsetData("cached_items_all");
         }
         
+        // Update the property Quote::KEY_ITEMS with proper items.
+        if ($child instanceof \Magento\Quote\Model\Quote) {
+            $child->setItems($child->getAllVisibleItems());
+        }
+        
         if ($save) {
             $child->save();
         }
--- a/Test/Unit/Helper/CartTest.php
+++ b/Test/Unit/Helper/CartTest.php
@@ -1624,6 +1624,8 @@ class CartTest extends BoltTestCase
     {
         $currentMock = $this->getCurrentMock(['validateEmail']);
 
+        $pitem = $this->getQuoteItemMock();
+        
         $this->quoteMock->expects(static::once())->method('getData')->willReturn(
             [
                 'entity_id'          => self::PARENT_QUOTE_ID,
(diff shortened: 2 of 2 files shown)


vitaliyreznikov:
> when calling $immutableQuote->getItems();, actually it returns the items in parent quote instead of immutable quote

Do I understand correctly, we do not clone items right now? When we create an immutable quote we add "links" to the parent quote item?

dumega:
@vitaliyreznikov yes, you are right. so in the foreach block, $cartId = $item->getQuoteId(); return the parent quote id, not the immutable quote id

vitaliyreznikov:
What the steps to reproduce the issue? Something like if user changed parent quote but we didn't recreate the immutable one?

My concerns here: we create immutable quote very often, actually each time when we think that cart is changed. This is now plugin weakness because it works slow. If we add new SQL calls here, it will be even more slowly.

Can we move this code into the place where we decided to create bolt order from this immutable quote? After we check that cart from cache doesn't satisfy us?

dumega:
@vitaliyreznikov do you mean here https://github.com/BoltApp/bolt-magento2/blob/master/Helper/Cart.php#L1734 ?

vitaliyreznikov:
Here:
https://github.com/BoltApp/bolt-magento2/blob/358788072aa87cb9b106f4d1196290018f4f77af/Helper/Cart.php#L888

dumega:
> Here:
> https://github.com/BoltApp/bolt-magento2/blob/358788072aa87cb9b106f4d1196290018f4f77af/Helper/Cart.php#L888

Please check the [bugsnag logs](https://app.bugsnag.com/bolt/magento-2/errors/603730d329bdd700170cc3ef?filters[event.since][0][type]=eq&filters[event.since][0][value]=1d&filters[error.status][0][type]=eq&filters[error.status][0][value]=open&filters[metaData.META%20DATA.store_url][0][type]=eq&filters[metaData.META%20DATA.store_url][0][value]=https%3A%2F%2Fmigrate.andrewchristian.com%2F), the line you link I am afraid it is too late.

danac-gs (approved):
LGTM

vitaliyreznikov:
@dumega How we can reproduce this?

dumega:
@vitaliyreznikov For now I only can reproduce in the staging site of merchant andrewchristian, and it seems that the bug can be reproduced when the aheadworks/module-sarp2 and amasty/promo are both installed.

And when I output some debug info on my dev site, all the items of immutable quote fetched by getItems() actually belong to parent quote. You can verify this on your local as well.

vitaliyreznikov (approved):

PR #1171, opened 2021-03-03
[M2P395] Billing address is invalid if it was created or changed on native checkout page

mikeshutov-bolt:
# Description
Subscribing to billing address change and creating a new order token for the updated billing address. Also using cartBarrier instead of BoltState.boltcart for configure call to make sure we have the latest cart values.

Fixes: [M2P395](https://boltpay.atlassian.net/browse/M2P-395) - Billing address is invalid if it was created or changed on native checkout page

#changelog (WIP)[M2P395] Billing address is invalid if it was created or changed on native checkout page

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -346,9 +346,7 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
          * return string
          */
         var getCheckoutType = function() {
-            return trim(location.pathname, '/') === 'checkout'
-            && (settings.quote_is_virtual || location.hash === '#payment' || location.hash === '#shipping')
-                ? 'payment' : 'checkout';
+            return trim(location.pathname, '/') === 'checkout' ? 'payment' : 'checkout';
         };
 
         /**
@@ -406,9 +404,6 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
             });
         };
 
-        var isPaymentOnlyCheckout = function() {
-            return !(trim(location.pathname, '/') === 'checkout/cart') && trim(location.pathname, '/') === 'checkout'
-        }
         /**
          * Inject connect.js
          * return void
@@ -1196,7 +1191,7 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
             BoltState.codeIsLoaded.resolve();
         }
         /////////////////////////////////////////////////////
-        if (isPaymentOnlyCheckout()) {
+        if (getCheckoutType() === 'payment') {
             /////////////////////////////////////////////////////
             // Call Bolt order creation Magento endpoint on
             // customer billing address change on payment only page,
@@ -1205,7 +1200,7 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
             var billing_address_value;
             onAttributesChange(billing_address_selector, function(element) {
                 var address = element.value;
-                if(address === 'null' || address === '') return;
+                if (address === 'null' || address === '') return;
 
                 if (address !== billing_address_value) {
                     billing_address_value = address;
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -1036,10 +1036,11 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
 
                             hints = deepMergeObjects(hints, data.hints);
                             hints.prefill = prefill;
+
                             //////////////////////////
                             resolve(data.cart);
                             hintBarrier.resolve(hints);
-                            console.log(data.cart);
+
                             // open the checkout if auto-open flag is set
                             // one time only on page load
                             if (BoltState.initiateCheckout && allowAutoOpen) {
@@ -1197,7 +1198,6 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
         }
         /////////////////////////////////////////////////////
         if (isPaymentOnlyCheckout()) {
-            console.log('payment only');
             /////////////////////////////////////////////////////
             // Call Bolt order creation Magento endpoint on
             // customer billing address change on payment only page,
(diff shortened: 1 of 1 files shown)


njfaries (approved):

view/frontend/templates/js/replacejs.phtml
```diff
@@ -1485,6 +1489,15 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
             'jquery',
             'Magento_Checkout/js/model/quote',
             ], function ($, quote) {
+                var prevAddress;
+                //If billing address has been changed we need to recreate the order token.
+                quote.billingAddress.subscribe(function(newAddress){
+                    if((!newAddress ^ !prevAddress || prevAddress.getKey() !== newAddress.getKey())){
```
vitaliyreznikov, line 1495: can we save only prevAddress.key?

vitaliyreznikov (approved):
Looks good, one non-blocking proposal

view/frontend/templates/js/replacejs.phtml
```diff
@@ -1007,24 +1007,28 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
                             if (BoltState.cartRestricted) {
                                 if (popUpOpen) reject(new Error(data.message));
                                 hintBarrier.resolve(hints);
+                                cartBarrier.resolve(cart);
```
danac-gs, line 1010: Where is this `cart` variable defined? I couldn't find it.

view/frontend/templates/js/replacejs.phtml
```diff
@@ -1007,24 +1007,28 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
...
                                 return;
                             }
 
                             if (data.status !== 'success') {
                                 reject(new Error(data.message || 'Network request failed'));
                                 hintBarrier.resolve(hints);
+                                cartBarrier.resolve(cart);
```
danac-gs, line 1017: Same here. Where is this `cart` variable defined?

view/frontend/templates/js/replacejs.phtml
```diff
@@ -1007,24 +1007,28 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
...
                                 return;
                             }
 
                             if (!data.cart) {
                                 reject(new Error('The cart info is missing.'));
                                 hintBarrier.resolve(hints);
+                                cartBarrier.resolve(cart);
```
danac-gs, line 1024: The same. Where is this `cart` variable defined?

view/frontend/templates/js/replacejs.phtml
```diff
@@ -1007,24 +1007,28 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
...
                                 return;
                             }
 
                             if (!data.cart.orderToken) {
                                 reject(new Error('The cart is empty.'));
                                 hintBarrier.resolve(hints);
+                                cartBarrier.resolve(cart);
```
danac-gs, line 1031: And here. Where is this `cart` variable defined?

danac-gs (changes requested):
Please check the comments.

view/frontend/templates/js/replacejs.phtml
mikeshutov-bolt: I found a different solution for this while checking whole code fixed the old billingAddress listener

view/frontend/templates/js/replacejs.phtml
mikeshutov-bolt: I ended up redoing this and finding old billingAddress payment only listener and fixing it that way so this is no longer in the PR

njfaries (approved):

danac-gn (approved):

view/frontend/templates/js/replacejs.phtml
```diff
@@ -406,6 +406,9 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
             });
         };
 
+        var isPaymentOnlyCheckout = function() {
```
vitaliyreznikov, line 409: I think if we refactor something it's better to change across the code.
After this change sometimes we use `isPaymentOnlyCheckout()` and sometimes `getCheckoutType()`.
I don't know how important it is but function getCheckoutType() return checkout for page like http://m2.vitaliy.dev.bolt.me/checkout/ and allow mini cart work on this page.

Can you explain a little why you need this change?

mikeshutov-bolt: It's a fair point. I'll explain my logic and why I went with this route. 

getCheckoutType - onload of /checkout page the url will not always have a # value when this function runs for native checkout. This results in an inconsistant result where most of the time it is actually 'checkout' and not 'payment' so conditions which depends on 'payment' being there onload will not work. The only situation when 'payment' was returned as value was when I refreshed with # values in the url. 

The only reason I did not modify the existing function is I am concerned about additional items breaking if I were to change it. However I am not against modifying getCheckoutType to a similar condition where we check for this:
```javascript
        var getCheckoutType = function() {
            var trimmedUrl = trim(location.pathname, '/');
            return trimmedUrl === 'checkout' && trimmedUrl !== 'checkout/cart' && settings.quote_is_virtual  ? 'payment' : 'checkout';
        };
```

vitaliyreznikov: We jumped on call and Mike explained me the fix. I think we can go ahead and update getCheckoutType function.
Can it be just
```
    var getCheckoutType = function() {
            return trim(location.pathname, '/') === 'checkout'   ? 'payment' : 'checkout';
        };
```

vitaliyreznikov (approved):
LGTM with one minor comment

view/frontend/templates/js/replacejs.phtml
mikeshutov-bolt: Retesting with change.

lbliii:
@matt-thomason, catching up on release notes and i was wondering if this bugfix is worth externalizing? (did it happen frequently enough to shoppers? it seems minor and i feel like we usually override most native checkout pages but just want to check if I understand the scenario fully).

PR #1188, opened 2021-03-16
M2P-518 Postpone connect.js load when possible

dumega:
# Description
Some merchant claim that bolt script affect their side speed.

So we implement the following logic under feature switch M2_LOAD_CONNECT_JS_ON_SPECIFIC_PAGE (disabled by default):

1. Fetch connect.js on page load only under the one of the following conditions:
    1) On the cart page
    2) On the product page and PPC is enabled

2. If on product page, PPC is disabled and mini cart support is enabled, then we dynamically load connect.js under the one of the following conditions:
    1) The cart contains any item.
    2) The cart is empty, and the customer add product to the cart.

Fixes: https://boltpay.atlassian.net/browse/M2P-518

#changelog Postpone connect.js load when possible

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -32,7 +32,7 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
 }
 
 $connectJsUrl = $block->getConnectJsUrl();
-$isLoadConnectJsDynamic = !$block->isLoadConnectJsEnabled();
+$isLoadConnectJsDynamic = $block->isLoadConnectJsDynamic();
 ?>
 
 <script type="text/javascript">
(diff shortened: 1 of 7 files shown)


view/frontend/templates/js/replacejs.phtml
```diff
@@ -1470,6 +1499,19 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
...
+        
+        <?php
+        if ($isLoadConnectJsDynamic) {
+        ?>
+        magentoCartDataListenerLoadConnectJs = function(data) {
+            if (data.items !== undefined && data.items.length > 0) {
+                insertConnectScript();
+            }            
+        }
+        customerData.get('cart').subscribe(magentoCartDataListenerLoadConnectJs);
```
vitaliyreznikov, line 1511: Why you prefer to use a new listener instead of using existing one?

view/frontend/templates/js/replacejs.phtml
```diff
@@ -1400,14 +1410,33 @@ if (!$block->isOnPageFromWhiteList() && !$block->isMinicartEnabled()) { return;
...
+                // set flag to animate always present button when we catch cart updating
+                if (boltConfig.always_present_checkout) {
+                    newItemAddedToCart = true;
+                }
             }
+            <?php
+            if ($isLoadConnectJsDynamic) {
+            ?>
+            if (insertConnectScriptWithoutSrc()) {
+                $.getScript("<?= /* @noEscape */ $connectJsUrl; ?>", function(data, textStatus, jqxhr) {
```
vitaliyreznikov, line 1427: Could you please explain why you prefer getScript instead of inserting script with src?

view/frontend/templates/js/replacejs.phtml
dumega: Cause here we only need to load connect js, and actually we expect it is executed only once, if the tag bolt-connect already exists, insertConnectScript won't load js repeatedly. Then the existing one does something else which is not needed.

view/frontend/templates/js/replacejs.phtml
dumega: Cause getScript has a callback function that is executed if the request succeeds.

vitaliyreznikov (approved):

PR #1201, opened 2021-03-28
M2P-536 Bugfix: Infinite loop on checkout page

dumega:
# Description
When working for the task https://boltpay.atlassian.net/browse/M2P-254, I found that on the checkout page, especially the payment step, the page always stuck on the loading screen. After investigation, it is because each payment has its own billing fields, and with the specific theme and some payment gateways, the callback monitorAttributesChange (https://github.com/BoltApp/bolt-magento2/blob/2.20.0/view/frontend/templates/js/replacejs.phtml#L1389) is triggered repeatedly and result in infinite loop.

Solution: Only trigger the callback for specific elements related to Bolt.

Fixes: https://boltpay.atlassian.net/browse/M2P-536

#changelog Bugfix: Infinite loop on checkout page

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -1372,10 +1372,17 @@ $isLoadConnectJsDynamic = $block->isLoadConnectJsDynamic();
         }
         /////////////////////////////////////////////////////////////////////
 
+        if (getCheckoutType()==="payment") {
+            var country_selectors = ['div[name="billingAddressboltpay.country_id"] select[name=country_id]','div[name="shippingAddress.country_id"] select[name=country_id]','.payment-method-boltpay select[name=country_id]'];
+            var state_selectors = ['div[name="billingAddressboltpay.region_id"] select[name=region_id]','div[name="shippingAddress.region_id"] select[name=region_id]','.payment-method-boltpay select[name=region_id]'];
+        } else {
+            var country_selectors = ['select[name=country_id]'];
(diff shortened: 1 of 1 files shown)


danac-gs (approved):

vitaliyreznikov:
@mikeshutov-bolt Could you please take a look?

mikeshutov-bolt:
@dumega The change makes sense but I have a few questions. 
1. Do we need the else condition? As in does this code need to run on none payment only checkout flow? Because from what ive seen we dont have billing fields if its not the payment only checkout.
2. If we don't maybe we can just wrap whole block to run on payment only checkout?

mikeshutov-bolt (approved):
LGTM pending questions.

dumega:
> @dumega The change makes sense but I have a few questions.
> 
> 1. Do we need the else condition? As in does this code need to run on none payment only checkout flow? Because from what ive seen we dont have billing fields if its not the payment only checkout or do we still want to trigger configure when they use shipping estimator?
> 2. If we don't maybe we can just wrap whole block to run on payment only checkout using Bolt fields specifically?


@mikeshutov-bolt 
These codes exists for a long time, and I am not sure what they are exactly for. To avoid raising potential bug, I choose to keep them. After all, my purpose is to fix issue in native checkout page, there is no need to touch code for other logic.

mikeshutov-bolt (approved):

PR #1268, opened 2021-05-01
M2P-583 Bugfix: shipping coupon not applying

dumega:
# Description
For split shipping and tax, we do not calculate the shipping discount if the coupon has **Apply to Shipping Amount** enabled, as a result, the order creation fails due to the total mismatch error.

Solution: In the tax endpoint, we have selected shipping option, then we can calculate shipping discount and return updated shipping cost.

Fixes: https://boltpay.atlassian.net/browse/M2P-583

#changelog Bugfix: shipping coupon not applying

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Model/Api/Tax.php
+++ b/Model/Api/Tax.php
@@ -198,15 +198,20 @@ class Tax extends ShippingTax implements TaxInterface
     /**
      * @param TotalsInterface $totalsInformation
      * @param string $currencyCode
-     * @param array $shipping_option
+     * @param array $shippingOption
      * @return array
      * @throws \Exception
      */
-    protected function getShippingDiscount($totalsInformation, $currencyCode, $shipping_option)
+    protected function getShippingDiscount($totalsInformation, $currencyCode, $shippingOption)
     {
-        $shippingDiscountAmount = $this->eventsForThirdPartyModules->runFilter("collectShippingDiscounts", $totalsInformation->getShippingDiscountAmount(), $this->quote, $this->quote->getShippingAddress());
+        $shippingDiscountAmount = $this->eventsForThirdPartyModules->runFilter(
+            "collectShippingDiscounts",
+            $totalsInformation->getShippingDiscountAmount(),
+            $this->quote,
+            $this->quote->getShippingAddress()
+        );
         if ($shippingDiscountAmount >= DiscountHelper::MIN_NONZERO_VALUE) {
-            $service = $shipping_option['service'] ?? '';
+            $service = $shippingOption['service'] ?? '';
             $shippingCost = $totalsInformation->getShippingAmount() - $shippingDiscountAmount;
             $shippingRoundedCost = CurrencyUtils::toMinor($shippingCost, $currencyCode);
             $diff = CurrencyUtils::toMinorWithoutRounding($shippingCost, $currencyCode) - $shippingRoundedCost;
@@ -217,16 +222,16 @@ class Tax extends ShippingTax implements TaxInterface
                 $discount = $this->priceHelper->currency($shippingDiscountAmount, true, false);
                 $service .= " [$discount" . "&nbsp;discount]";
             }
-            $shipping_option['service'] = html_entity_decode($service);
+            $shippingOption['service'] = html_entity_decode($service);
         } else {
             $shippingRoundedCost = CurrencyUtils::toMinor($totalsInformation->getShippingAmount(), $currencyCode);
             $taxAmount = CurrencyUtils::toMinor($totalsInformation->getShippingTaxAmount(), $currencyCode);
         }
         
-        $shipping_option['cost'] = $shippingRoundedCost;
-        $shipping_option['tax_amount'] = $taxAmount;
+        $shippingOption['cost'] = $shippingRoundedCost;
+        $shippingOption['tax_amount'] = $taxAmount;
         
-        return $shipping_option;
+        return $shippingOption;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Model/Api/Tax.php
```diff
@@ -169,17 +181,53 @@ public function createInstorePickOption($totalsInformation, $currencyCode, $ship
...
-        $shipToStoreOption->setTaxAmount(CurrencyUtils::toMinor($totalsInformation->getShippingTaxAmount(), $currencyCode));
+        $shipToStoreOption->setTaxAmount($ship_to_store_option['tax_amount']);
 
         return $shipToStoreOption;
     }
+    
+    /**
+     * @param TotalsInterface $totalsInformation
+     * @param string $currencyCode
+     * @param array $shipping_option
```
danac-gs, line 201: Let's stick to camel case whenever possible

Model/Api/Tax.php
```diff
@@ -169,17 +181,53 @@ public function createInstorePickOption($totalsInformation, $currencyCode, $ship
...
+     * @return array
+     * @throws \Exception
+     */
+    protected function getShippingDiscount($totalsInformation, $currencyCode, $shipping_option)
+    {
+        $shippingDiscountAmount = $this->eventsForThirdPartyModules->runFilter("collectShippingDiscounts", $totalsInformation->getShippingDiscountAmount(), $this->quote, $this->quote->getShippingAddress());
```
danac-gs, line 207:
```suggestion
        $shippingDiscountAmount = $this->eventsForThirdPartyModules->runFilter(
	      "collectShippingDiscounts",
	      $totalsInformation->getShippingDiscountAmount(),
	      $this->quote,
	      $this->quote->getShippingAddress()
	  );
```

Model/Api/Tax.php
```diff
@@ -169,17 +181,53 @@ public function createInstorePickOption($totalsInformation, $currencyCode, $ship
...
+            $shippingRoundedCost = CurrencyUtils::toMinor($shippingCost, $currencyCode);
+            $diff = CurrencyUtils::toMinorWithoutRounding($shippingCost, $currencyCode) - $shippingRoundedCost;
+            $taxAmount = round(CurrencyUtils::toMinorWithoutRounding($totalsInformation->getShippingTaxAmount(), $currencyCode) + $diff);
+            if ($shippingRoundedCost == 0) {
+                $service .= ' [free&nbsp;shipping&nbsp;discount]';
+            } else {
+                $discount = $this->priceHelper->currency($shippingDiscountAmount, true, false);
+                $service .= " [$discount" . "&nbsp;discount]";
+            }
+            $shipping_option['service'] = html_entity_decode($service);
```
danac-gs, line 220: Since we are adding a suffix either way, let's decode (only) the service before.

danac-gs (changes requested):

Model/Api/Tax.php
dumega: Fixed

Model/Api/Tax.php
dumega: Done

Model/Api/Tax.php
dumega: @danac-gs I believe this is to convert the '&nbsp;' entity to ASCII code 160 properly. Actually this was created by you https://github.com/BoltApp/bolt-magento2/commit/be3986e883835acd2cf244c3220fa9343862fabf#diff-11cba1d080bcf8a37d870a6bffb136b052d3be87e055ac78ff8ff87fa20f5355R568

Model/Api/Tax.php
```diff
@@ -169,17 +181,58 @@ public function createInstorePickOption($totalsInformation, $currencyCode, $ship
...
+            $shippingRoundedCost = CurrencyUtils::toMinor($shippingCost, $currencyCode);
+            $diff = CurrencyUtils::toMinorWithoutRounding($shippingCost, $currencyCode) - $shippingRoundedCost;
+            $taxAmount = round(CurrencyUtils::toMinorWithoutRounding($totalsInformation->getShippingTaxAmount(), $currencyCode) + $diff);
+            if ($shippingRoundedCost == 0) {
+                $service .= ' [free&nbsp;shipping&nbsp;discount]';
+            } else {
+                $discount = $this->priceHelper->currency($shippingDiscountAmount, true, false);
+                $service .= " [$discount" . "&nbsp;discount]";
+            }
+            $shippingOption['service'] = html_entity_decode($service);
```
danac-gs, line 225:
```suggestion
        $shipping_option['service'] = sprintf(
            '%s [%s discount]',
            html_entity_decode($service),
            $shippingRoundedCost == 0
                ? 'free shipping'
                : $this->priceHelper->currency($shippingDiscountAmount, true, false)
        );
```
I don't remember why that was added, perhaps regarding something passed in the service title, not the whitespace. In that case, the above suggestion looks better. However, as we are not sure, you better leave it as is. Or test it with some live shipping options (not free and flat), provided out of the box in Magento (Fedex, UPS, USPS, something should work for testing without keys).

danac-gs (approved):

njfaries (approved):
Overall LGTM. If we are happy with leaving the comment above as-is we can go ahead and merge for 2.22 tag

Model/Api/Tax.php
dumega: I prefer to leave it as-is, cause it works as expected so far.

PR #1365, opened 2021-07-29
Prevent overwriting quote customer IP with Bolt (AWS) one in pre-auth call

danac-gs:
# Description
Merchant requires customer IP but we are only sending AWS IPs in our requests.
Merchant is reporting that as of 29/01/2020 they haven't received the customer IP in their DB when orders are created via Bolt.
This is a generic, although not so common, case - some 3party modules may trigger totals collection in pre-auth processing and get and store the IP from the session. This PR prevents that in our calls, leaving the initial, customer session IP in place.

Fixes: https://boltpay.atlassian.net/browse/EN-5023

#changelog Prevent overwriting quote customer IP with Bolt (AWS) one in pre-auth call

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/FeatureSwitch/Definitions.php
+++ b/Helper/FeatureSwitch/Definitions.php
@@ -217,6 +217,11 @@ class Definitions
      */
     const M2_IGNORE_TOTAL_VALIDATION_WHEN_CREDIT_HOOK_IS_SENT_TO_MAGENTO = 'M2_IGNORE_TOTAL_VALIDATION_WHEN_CREDIT_HOOK_IS_SENT_TO_MAGENTO';
 
+    /**
+     * Prevent Bolt IPs from being saved as customer IP on quote
+     */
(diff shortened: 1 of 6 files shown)


Helper/FeatureSwitch/Definitions.php
```diff
@@ -434,5 +439,11 @@ class Definitions
             self::DEFAULT_VAL_KEY => false,
             self::ROLLOUT_KEY     => 0
         ],
+        self::M2_PREVENT_SETTING_BOLT_IPS_AS_CUSTOMER_IP_ON_QUOTE => [
+            self::NAME_KEY        => self::M2_PREVENT_SETTING_BOLT_IPS_AS_CUSTOMER_IP_ON_QUOTE,
+            self::VAL_KEY         => true,
+            self::DEFAULT_VAL_KEY => false,
+            self::ROLLOUT_KEY     => 0
```
vitaliyreznikov, line 446: Why do you want to disable it by default?

danac-gs: Good point. Will switch to enabled.

danac-gs:
```suggestion
            self::ROLLOUT_KEY     => 100
```

vitaliyreznikov (approved):
LGTM

PR #1368, opened 2021-07-30
ON-792 ImaginationMedia_TmwGiftCard: Ignore invoice creation for Bolt order that has a virtual gift card when the hook is sent to Magento

johnmccombs1:
# Description

This PR resolve the following case:
1. The customer placed a Bolt order that has a virtual gift card. The order was created on both Magento and Bolt.
2. The pending hook was sent to Magento and Magento threw the exception “Transaction is not in authorized state.”, 

See the Bugsnag link here:
https://app.bugsnag.com/bolt/magento-2/errors/6102e1040639e600079da9e9?filters%5Bevent.since%5D%5B0%5D%5Btype%5D=eq&filters%5Bevent.since%5D%5B0%5D%5Bvalue%5D=30d&filters%5Bsearch%5D%5B0%5D%5Btype%5D=eq&filters%5Bsearch%5D%5B0%5D%5Bvalue%5D=%22m2dev.shopakira.com%22
This PR resolves the issue by ignoring the invoice creation for Bolt order that has virtual gift card when the hook is sent to Magento

Fixes: https://boltpay.atlassian.net/browse/ON-792

#changelog ON-792 ImaginationMedia_TmwGiftCard: Ignore invoice creation for Bolt order that has a virtual gift card when the hook is sent to Magento

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Plugin/ImaginationMedia/TmwGiftCard/ForceInvoicePlugin.php
+++ b/Plugin/ImaginationMedia/TmwGiftCard/ForceInvoicePlugin.php
@@ -1,0 +1,53 @@
+<?php
+/**
+ * Bolt magento2 plugin
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ *
+ * @category   Bolt
+ * @package    Bolt_Boltpay
+ * @copyright  Copyright (c) 2017-2021 Bolt Financial, Inc (https://www.bolt.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+namespace Bolt\Boltpay\Plugin\ImaginationMedia\TmwGiftCard;
+
+use Bolt\Boltpay\Helper\Hook;
+
+/**
+ * Class ForceInvoicePlugin
+ * @package Bolt\Boltpay\Plugin\ImaginationMedia\TmwGiftCard
+ */
+class ForceInvoicePlugin
+{
+    /**
+     * When the pending hook is sent to Magento, the $subject observer would try to create an online invoice for the order
+     * if it has a virtual gift card and because the Bolt transaction is not yet authorized, an error of
+     * “Transaction is not in authorized state.”  would be thrown.
+     *
+     * We don't need the logic in the $subject observer as our payment/capture hook will generate an offline invoice later.
+     * This would bypass the unneeded logic from the $subject observer.
+     *
+     * @param \ImaginationMedia\TmwGiftCard\Observer\ForceInvoice $subject
+     * @param callable $proceed
+     * @param $observer
+     */
+    public function aroundExecute(
+        \ImaginationMedia\TmwGiftCard\Observer\ForceInvoice $subject,
+        callable $proceed,
+        $observer
+    ) {
+        $payment = $observer->getEvent()->getPayment();
+        $order = $payment->getOrder();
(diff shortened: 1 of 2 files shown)


Plugin/ImaginationMedia/TmwGiftCard/ForceInvoicePlugin.php
```diff
@@ -0,0 +1,47 @@
...
+     * @param \ImaginationMedia\TmwGiftCard\Observer\ForceInvoice $subject
+     * @param callable $proceed
+     * @param $observer
+     */
+    public function aroundExecute(
+        \ImaginationMedia\TmwGiftCard\Observer\ForceInvoice $subject,
+        callable $proceed,
+        $observer
+    )
+    {
```
danac-gs, line 38:
```suggestion
    ) {
```

Plugin/ImaginationMedia/TmwGiftCard/ForceInvoicePlugin.php
```diff
@@ -0,0 +1,47 @@
...
+
+use Bolt\Boltpay\Helper\Hook;
+
+/**
+ * Class ForceInvoicePlugin
+ * @package Bolt\Boltpay\Plugin\ImaginationMedia\TmwGiftCard
+ */
+class ForceInvoicePlugin
+{
+    /**
```
danac-gs, line 28: Please add a brief description why this is needed.

danac-gs (changes requested):

vitaliyreznikov:
@johnmccombs1 

> Magento threw the exception “Transaction is not in an authorized state.”

Could you please explain why it happens? Not sure I understand.

Plugin/ImaginationMedia/TmwGiftCard/ForceInvoicePlugin.php
johnmccombs1: @danac-gs  I've added it. Can you check?

johnmccombs1:
@vitaliyreznikov 
When the pending hook is sent to Magento, the $subject observer would try to create an online invoice for the order if it has a virtual gift card and because the Bolt transaction is not yet authorized, an error of “Transaction is not in authorized state.”  would be thrown. We don't need the logic in the $subject observer as our payment/capture hook will generate an offline invoice later. This would bypass the unneeded logic from the $subject observer.

You can see the bugsnag link for the error here: 
https://app.bugsnag.com/bolt/magento-2/errors/6102e1040639e600079da9e9?filters%5Bevent.since%5D%5B0%5D%5Btype%5D=eq&filters%5Bevent.since%5D%5B0%5D%5Bvalue%5D=30d&filters%5Bsearch%5D%5B0%5D%5Btype%5D=eq&filters%5Bsearch%5D%5B0%5D%5Bvalue%5D=%22m2dev.shopakira.com%22

danac-gs:
@johnmccombs1 If I understand correctly, this 3p module triggers invoice creation earlier than usual (expected), while the transaction is still pending. This solution bypasses the online invoice creation at that point and relies on an offline one triggered by out plugin latter. I was wondering how this 3p module triggers the invoice creation and when. Does it happen only once, or, for instance, on every order save (with a check and skip if it was already created). If so, could we skip the invoice creation process while the transaction is pending and let it create it when the transaction is authorized? So a natural, online invoice gets created in Magento.

johnmccombs1:
@danac-gs **Does it happen only once, or, for instance, on every order save (with a check and skip if it was already created)**
=> it happens only once. Their 3p module is using event sales_order_payment_place_start to handle the logic for invoice creation and causing conflicts with Bolt logic.

**If so, could we skip the invoice creation process while the transaction is pending and let it create it when the transaction is authorized? So a natural, online invoice gets created in Magento.**
=> If we do this, we would need to modify a lot of logic in their 3p module which is not an ideal solution in my option.

danac-gs (approved):

PR #1454, opened 2021-11-04
EN-6094 Fix bug: Can not refund if an order is adjusted price mismatch

johnmccombs1:
# Description

Currently, we can't refund if an order is adjusted price mismatch. 
See the detail in this task: https://app.asana.com/0/564264490825835/1201256841759895

=> The root cause is that the grand total isn't correct in the admin new credit memo creation page. This PR is to correct the credit memo grand total by including the mismatch amount into the tax amount when adjusting price mismatch.

FYI, this logic is the same as M1 logic. See here: https://github.com/BoltApp/bolt-magento1/blob/master/app/code/community/Bolt/Boltpay/Model/Order.php#L518

Fixes: https://boltpay.atlassian.net/browse/EN-6094

#changelog EN-6094 Fix bug: Can not refund if an order is adjusted price mismatch

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/FeatureSwitch/Decider.php
+++ b/Helper/FeatureSwitch/Decider.php
@@ -396,4 +396,9 @@ class Decider extends AbstractHelper
     {
         return $this->isSwitchEnabled(Definitions::M2_ENABLE_PRODUCT_ENDPOINT);
     }
+
+    public function isIncludeMismatchAmountIntoTaxWhenAdjustingPriceMismatch()
+    {
+        return $this->isSwitchEnabled(Definitions::M2_INCLUDE_MISMATCH_AMOUNT_INTO_TAX_WHEN_ADJUSTING_PRICE_MISMATCH);
+    }
 }
--- a/Helper/FeatureSwitch/Definitions.php
+++ b/Helper/FeatureSwitch/Definitions.php
@@ -46,6 +46,11 @@ class Definitions
     const M2_BOLT_ENABLED = 'M2_BOLT_ENABLED';
 
     /**
+     * Enable including mismatch amount into tax when adjusting price mismatch
+     */
+    const M2_INCLUDE_MISMATCH_AMOUNT_INTO_TAX_WHEN_ADJUSTING_PRICE_MISMATCH = 'M2_INCLUDE_MISMATCH_AMOUNT_INTO_TAX_WHEN_ADJUSTING_PRICE_MISMATCH';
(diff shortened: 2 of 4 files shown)


vitaliyreznikov:
@johnmccombs1 
1. Please communicate this solution with cabinets.com developers, we need to make sure it works for them.
2. I am not sure I love these tax changes. Honestly, I don't love the idea of total adjustment at all. If we need it please do it under feature switch (it can be enabled by default).
3. Can you work on identifying the reason for total mismatch for cabinets.com?

johnmccombs1:
@vitaliyreznikov 
1. Please communicate this solution with cabinets.com developers, we need to make sure it works for them.
=> I was able to replicate the issue in an M2 default store and I confirmed this PR fixed the issue. Do we still need them to confirm it works before approving the PR?
2.  If we need it please do it under feature switch (it can be enabled by default).
=> Do you want a feature switch for only tax changes or total adjustment? 
3. I am not sure I love these tax changes. Honestly, I don't love the idea of total adjustment at all.
 => They're using the price fault tolerance feature. image
and the total adjustment is a part of this feature
4. Can you work on identifying the reason for the total mismatch for cabinets.com?
The reason for the total mismatch is the rounding discount issue. I asked the merchant for the SSH staging to debug the issue further. However, they don't want to give it to us. I'll ask their dev to debug this issue further.

vitaliyreznikov:
1. If you can reproduce the total mismatch bug with default m2 settings would you mind describing steps for reproducing in a separate asana? So we can discuss the best solution.
2. Regarding to this fix I am ok to approve it once the change is under feature switch and merchant confirms that tax change work for them

vitaliyreznikov:
@johnmccombs1 We can do it parallel and merge this one once feature switch is introduced

danac-gs:
Perhaps we should look into a better, generic solution for mismatches caused by more than two decimal places. Something like intercepting the calculated results and rounding them to 2 decimals early on, before they are used by Magento, Bolt or other modules. Thus eliminating the fault tolerance feature and the need of keeping track of diffs after every operation in cart generating process, which is ugly and prone to errors.

johnmccombs1:
@vitaliyreznikov I've added the feature switch. Could you please review?

vitaliyreznikov (approved):
LGTM

PR #1459, opened 2021-11-17
Do not create immutable quote for API driven integration

vitaliyreznikov:
- introduce feature switch M2_ENABLE_API_DRIVEN_INTEGRAION
- Under feature switch update **getCartData** function to create bolt order without immutable quote creation.

Eventually we will delete code for immutable quote so I prefer to have few <IF> statement for now rather than separate function.

Fixes: (link Jira ticket)

#changelog Do not create immutable quote for API driven integration

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -2656,7 +2656,7 @@ class Cart extends AbstractHelper
             }
 
             // get immutable quote id stored with cart data
-            if ($response) {
+            if (!$this->deciderHelper->isAPIDrivenIntegrationEnabled() && $response) {
                 $cartReference = $this->getImmutableQuoteIdFromBoltCartArray($responseData['cart']);
             } else {
                 $cartReference = '';
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -1840,23 +1840,36 @@ class Cart extends AbstractHelper
         // cart data is being created for sending to Bolt create
         // order API, otherwise skip this step
         ////////////////////////////////////////////////////////
-        if (!$immutableQuote) {
+        if (!$immutableQuote && !$this->deciderHelper->isAPIDrivenIntegrationEnabled()) {
             $immutableQuote = $this->createImmutableQuote($quote);
         }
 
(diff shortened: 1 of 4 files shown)


mikeshutov-bolt:
What is the purpose for having this extra feature? isRecalculateTotalForAPIDrivenIntegration im trying to understand why do we need this as a just in case feature?

vitaliyreznikov:
> What is the purpose for having this extra feature? isRecalculateTotalForAPIDrivenIntegration im trying to understand why do we need this as a just in case feature?

The total calculation is very heavy function. From my understanding, we do not need it because the total should be calculated before when a quote is updated for whatever reasons.
But I may be wrong, so I added feature switch we can enable across the board if recalculation is needed

mikeshutov-bolt (approved):
Not a fan of the isRecalculateTotalForAPIDrivenIntegration feature personally but I understand the reasoning, it just seems to add more complexity and switches.
LGTM otherwise.

dumega (dismissed):
@vitaliyreznikov I run some tests on my local, there is an error "Bolt order doesn't contain immutable order id" (https://github.com/BoltApp/bolt-magento2/blob/5c19d973839413a913e614c14617ed57512c1d33/Helper/Cart.php#L760) when creating Bolt cart. Could you double check?

vitaliyreznikov:
@dumega With this change, we do not expect shipping calls or ever another call should work. We already have shipping call, tax call,s and order creation in the new flow. I'll add some guide about how you can test this.

vitaliyreznikov:
@dumega PMed some instruction, please review again

vitaliyreznikov:
@dumega I made a fix after our offline conversation. Please take a look now.

vitaliyreznikov:
@dumega We need some changes on bolt side which allow you to test all flow and those changes are not deployed yet.
But please let me merge this one because:
1. It works on my side
2. shipping step error you see most likely not related to this PR
3. we have enough time before release to fix any issues in following PRs if needed

PR #1460, opened 2021-11-17
Bugfix: Mismatch shipping issue if the quote has a cart rule of 'Fixed amount discount for whole cart' and 'Apply to shipping amount'

johnmccombs1:
# Description
Currently, customers can't place an order if they apply a coupon of 'Fixed amount discount for whole cart' and 'Apply to shipping amount'. This is because that the mismatch shipping issue. This PR fixes the issue by correcting the shipping amount calculation in the Bolt modal. 

Fixes: https://app.asana.com/0/564264490825835/1201367629380173

#changelog Bugfix: Mismatch shipping issue if the quote has a cart rule of 'Fixed amount discount for whole cart' and 'Apply to shipping amount'

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Model/Api/CreateOrder.php
+++ b/Model/Api/CreateOrder.php
@@ -680,7 +680,7 @@ class CreateOrder implements CreateOrderInterface
         }
         if ($quote->getShippingAddress() && !$quote->isVirtual()) {
             $amount = $quote->getShippingAddress()->getShippingAmount();
-            if (! $this->isBackOfficeOrder($quote)) {
+            if (! $this->isBackOfficeOrder($quote) && !$this->cartHelper->ignoreAdjustingShippingAmount($quote)) {
                 $amount -= $quote->getShippingAddress()->getShippingDiscountAmount();
             }
             $storeCost = CurrencyUtils::toMinor($amount, $quote->getQuoteCurrencyCode());
--- a/Model/Api/Tax.php
+++ b/Model/Api/Tax.php
@@ -24,6 +24,7 @@ use Bolt\Boltpay\Api\Data\TaxResultInterfaceFactory;
 use Bolt\Boltpay\Api\Data\TaxResultInterface;
 use Bolt\Boltpay\Helper\Shared\CurrencyUtils;
 use Bolt\Boltpay\Helper\Discount as DiscountHelper;
+use Bolt\Boltpay\Helper\Cart as CartHelper;
 use Magento\Checkout\Api\TotalsInformationManagementInterface;
 use Magento\Checkout\Api\Data\TotalsInformationInterface;
 use Magento\Quote\Api\Data\TotalsInterface;
@@ -65,11 +66,16 @@ class Tax extends ShippingTax implements TaxInterface
      * @var TotalsInformationInterface
      */
     protected $addressInformation;
-    
+
     /**
      * @var PriceHelper
      */
     protected $priceHelper;
+
+    /**
+     * @var CartHelper
+     */
+    protected $cartHelper;
 
     /**
      * Assigns local references to global resources
@@ -81,6 +87,7 @@ class Tax extends ShippingTax implements TaxInterface
      * @param TotalsInformationManagementInterface $totalsInformationManagement
      * @param TotalsInformationInterface $addressInformation
      * @param PriceHelper $priceHelper
+     * @param CartHelper $cartHelper
      */
     public function __construct(
         ShippingTaxContext $shippingTaxContext,
@@ -88,14 +95,17 @@ class Tax extends ShippingTax implements TaxInterface
         TaxResultInterfaceFactory $taxResultFactory,
         TotalsInformationManagementInterface $totalsInformationManagement,
         TotalsInformationInterface $addressInformation,
-        PriceHelper $priceHelper
-    ) {
+        PriceHelper $priceHelper,
+        CartHelper $cartHelper
+    )
+    {
         parent::__construct($shippingTaxContext);
         $this->taxDataFactory = $taxDataFactory;
         $this->taxResultFactory = $taxResultFactory;
         $this->totalsInformationManagement = $totalsInformationManagement;
         $this->addressInformation = $addressInformation;
         $this->priceHelper = $priceHelper;
+        $this->cartHelper = $cartHelper;
     }
 
     /**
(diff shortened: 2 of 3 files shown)


danac-gs (approved):
LGTM

vitaliyreznikov:
@johnmccombs1 Could you please explain this a little, it looks hard to review now.
1) Steps to reproduce. Which discount do we need to create to reproduce this issue?
2) Is it regression or not?
3) When the issue happens do we see a mismatch with shipping only? Or with total as well
4) Looks like your solution does not include a shipping discount in the shipping price. Why does it help? Also, why do we change only the tax endpoint but not the shipping endpoint?

vitaliyreznikov (commented):
@johnmccombs1 Could you please explain this a little, it looks hard to review now.

1. Steps to reproduce. Which discount do we need to create to reproduce this issue?
2. Is it regression or not?
3. When the issue happens do we see a mismatch with shipping only? Or with total as well
4. Looks like your solution does not include a shipping discount in the shipping price. Why does it help? Also, why do we change only the tax endpoint but not the shipping endpoint?

johnmccombs1:
@vitaliyreznikov 
1. Steps to reproduce. Which discount do we need to create to reproduce this issue?
- Create a coupon with simple action is **Fixed amount discount for whole cart** and enable **Apply To Shipping** 
image
- Add product and applies the coupon in the cart page
- Open Bolt modal, estimate shipping, tax
- Click the "Pay" button in the Bolt modal.
- The error occurs

2. When the issue happens do we see a mismatch with shipping only? Or with total as well
=> That's mismatch shipping only. Bolt is calculating the shipping amount incorrectly. We don't need to include the shipping discount in this case.

3. Looks like your solution does not include a shipping discount in the shipping price. Why does it help? 
=> Yes. In this case, Magento calculates it correctly in the function  quote->getShippingAddress()->getShippingAmount(). We don't need to include the shipping discount

4.  Also, why do we change only the tax endpoint but not the shipping endpoint?
=> That is because in the tax endpoint we re-update the shipping amount https://github.com/BoltApp/bolt-magento2/blob/master/Model/Api/Tax.php#L167

Model/Api/Tax.php
```diff
@@ -215,7 +225,8 @@ protected function getShippingDiscount($totalsInformation, $currencyCode, $shipp
             $this->quote,
             $this->quote->getShippingAddress()
         );
-        if ($shippingDiscountAmount >= DiscountHelper::MIN_NONZERO_VALUE) {
+
+        if ($shippingDiscountAmount >= DiscountHelper::MIN_NONZERO_VALUE && !$this->cartHelper->ignoreAdjustingShippingAmount($this->quote)) {
```
vitaliyreznikov, line 229: Looks like this change is related to 3p party plugins support, Mirasvit credit, and Mirasvit rewards.
Is it relevant to cabinet's bug?

Model/Api/CreateOrder.php
```diff
@@ -680,7 +680,7 @@ public function validateShippingCost($quote, $transaction)
         }
         if ($quote->getShippingAddress() && !$quote->isVirtual()) {
             $amount = $quote->getShippingAddress()->getShippingAmount();
-            if (! $this->isBackOfficeOrder($quote)) {
+            if (! $this->isBackOfficeOrder($quote) && !$this->cartHelper->ignoreAdjustingShippingAmount($quote)) {
```
vitaliyreznikov, line 683: So you mean we do not deduct from shipping amount shipping discount if it has type DISCOUNT_ACTION_FIXED_AMOUNT_FOR_CART. But what about other types? In which cases do we still need to deduct?
I think we need to remove this adjustment completely or have an integration test that covers this adjustment.

vitaliyreznikov (commented):
I am not comfortable approving this PR right now, I think we need to further investigate the issue and the best way to fix it.

In general, if the total is OK and the problem is only with the shipping amount we should not break auth process. If we change this it solves the merchant issue as I understand.

Saying this I am OK to include this code in the merchant branch because of urgency.

Model/Api/CreateOrder.php
johnmccombs1: In which cases do we still need to deduct?
=> @vitaliyreznikov I tested and confirmed that we still need to deduct for cases DISCOUNT_ACTION_BY_PERCENT, DISCOUNT_ACTION_FIXED_AMOUNT. @dumega Can you confirm these since you added the deducting logic

Model/Api/Tax.php
johnmccombs1: It's not related to the Mirasvit credit, and Mirasvit rewards. I was able to replicate the issue a Magento (2.4.3) default store in my dev instance.

vitaliyreznikov: Oh, my bad, I misread the code

Model/Api/CreateOrder.php
dumega: @johnmccombs1 This logic is not added by me, and was added when creating this file for Bolt pre-auth (https://github.com/BoltApp/bolt-magento2/commit/8134b200197201a76a660663b84984baecad9e07#diff-208413ae52687204accddd79663f92e3a8277b5643fe0e935b16607e4ad9511e).  @danac-gs  any insight on this?

danac-gs: > any insight on this?
Not really. I think the case John is covering (apply discount to shipping) was deliberately skipped because it was assumed not needed / urgent at a time so it was left as is. I know from doing tests that that path was not working in all cases but seems like no merchants asked for it so far. Anyway, I believe John's approach is OK if it passes all test cases.

vitaliyreznikov:
@johnmccombs1 I have a question. You skipped all discounts if any of rules has type \Magento\SalesRule\Api\Data\RuleInterface::DISCOUNT_ACTION_FIXED_AMOUNT_FOR_CART.
Is this possible that we have two rules and actually need skipping only one?

johnmccombs1:
@vitaliyreznikov You mean in the case that the cart has 2 rules (\Magento\SalesRule\Api\Data\RuleInterface::DISCOUNT_ACTION_FIXED_AMOUNT_FOR_CART) and another rule which is needed to deduct discount ( ex: DISCOUNT_ACTION_BY_PERCENT or DISCOUNT_ACTION_FIXED_AMOUNT)
and my change didn't cover that case?

I tested and confirmed Magento doesn't allow customer applies 2 rules (Apply To Shipping is Yes) that have different types on the cart page. So we don't need to worry about that case.

vitaliyreznikov (approved):
LGTM. Please update the description with coupon types you tested your code so the QA team can test as well as part of the release qualification.

johnmccombs1:
@vitaliyreznikov Updated

PR #1466, opened 2021-11-18
Fix mismatch discount (rounding) issue by re-setting rounding delta

johnmccombs1:
# Description

Currently, if the cart has multiple products and discounts, there is a mismatch discount (rounding) issue occurring which causes the customers not to be able to place an order.

To correct the Bolt discounts, we need to reset the rounding delta before saving them to the checkout session. Otherwise, $this->_roundingDeltas 
(https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/SalesRule/Model/Utility.php#L200) would return the old values and cause invalid discounts

Fixes: https://app.asana.com/0/564264490825835/1201256841759895

#changelog Fix mismatch discount (rounding) issue by re-setting rounding delta

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Plugin/SalesRuleModelUtilityPlugin.php
+++ b/Plugin/SalesRuleModelUtilityPlugin.php
@@ -62,7 +62,7 @@ class SalesRuleModelUtilityPlugin
                                    $result, $discountData, $item, $qty)
     {
         $checkoutSession = $this->sessionHelper->getCheckoutSession();
-        $savedRuleId = $checkoutSession->getBoltNeedCollectSaleRuleDiscounts('');      
+        $savedRuleId = $checkoutSession->getBoltNeedCollectSaleRuleDiscounts('');
         if (!empty($savedRuleId)) {
             // If the sale rule has no coupon, its discount amount can not be retrieved directly,
             // so we store the discount amount in the checkout session with the rule id as key.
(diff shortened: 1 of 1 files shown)


vitaliyreznikov:
@dumega Could you please review?

dumega:
@johnmccombs1 This function Magento\SalesRule\Model\Utility::resetRoundingDeltas() is only called when trying to reset quote and address applied rules, and the related properties **_roundingDeltas** is set in the function Magento\SalesRule\Model\Utility::deltaRoundingFix which is called before Magento\SalesRule\Model\Utility::minFix (https://github.com/magento/magento2/blob/0af0fba365a4268c3f2d5c7168fcf17dbeab148c/app/code/Magento/SalesRule/Model/RulesApplier.php#L224), so why is there still old values?

dumega (commented):
@johnmccombs1 This function Magento\SalesRule\Model\Utility::resetRoundingDeltas() is only called when trying to reset quote and address applied rules, and the related properties _roundingDeltas is set in the function Magento\SalesRule\Model\Utility::deltaRoundingFix which is called before Magento\SalesRule\Model\Utility::minFix (https://github.com/magento/magento2/blob/0af0fba365a4268c3f2d5c7168fcf17dbeab148c/app/code/Magento/SalesRule/Model/RulesApplier.php#L224), so why is there still old values?

johnmccombs1:
@dumega @vitaliyreznikov

johnmccombs1:
@dumega  I debugged it further and I confirmed the issue isn't from our plugin. Closing the PR now.

PR #1468, opened 2021-11-24
Fix Controller/ReceivedUrlTrait.php for M2 API driven

vitaliyreznikov:
In M2 API driven flow we still need the intermediate redirect page to assign new order to user session on it.
Two differences with the previous flow:
- we send order id, not incremental order id
- we do not need dispatch PostCheckoutEvents because they were already dispatched

#changelog Fix Controller/ReceivedUrlTrait.php for M2 API driven


mikeshutov-bolt (approved):
LGTM

PR #1479, opened 2021-12-08
Reload cart if integration mode changed

vitaliyreznikov:
When we disable or enable new api driven flow we should try not to affect the users in the middle of checkout process.

Before this PR users on cart page are blocked until they update cart.

With this PR users on the cart page just need to reload the page to be unblocked.

Just in case we do cart invalidation only once per cart page reloading.

#changelog invalidate cart if plugin mode changed


mikeshutov-bolt (approved):
LGTM

PR #1481, opened 2021-12-17
Use existing payment fields rather than additional info

vitaliyreznikov:
Details:
https://app.asana.com/0/1201161314919732/1201531702752390

#changelog Use existing payment fields rather than additional info


mikeshutov-bolt (approved):
LGTM

PR #1488, opened 2022-01-01
Refactor get masked id endpoint

vitaliyreznikov:
Refactor get masked id endpoint in order to eventually have all bolt API related to carts/quotes under one interface

#changelog Refactor get masked id endpoint


mikeshutov-bolt (approved):
Makes sense to have it all under one interface. LGTM

PR #1491, opened 2022-01-02
Change cart update endpoint

vitaliyreznikov:
Update endpoint for cart changing to allow set cart isActive to false

#changelog Change cart update endpoint


mikeshutov-bolt (approved):
LGTM

PR #1501, opened 2022-01-16
Bugfix: Unable to checkout with whole cart shipping discount

dumega:
# Description

For the "Fixed amount discount for whole cart" sale rule, the plugin already sends full discount amount to Bolt server, so there is no need to add it to shipping discounts when calculating tax.

Fixes: https://app.asana.com/0/1124368832381302/1201663986351838/f

#changelog Bugfix: Unable to checkout with whole cart shipping discount

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:


danac-gs (approved):

Model/Api/Tax.php
```diff
@@ -226,6 +234,14 @@ protected function getShippingDiscount($totalsInformation, $currencyCode, $shipp
             $this->quote->getShippingAddress()
         );
 
+        // Exclude discount amount of "Fixed amount discount for whole cart" sale rule from shipping discounts.
+        // Cause the Bolt cart has full discount amount of such type of sale rule applied before collecting shipping&tax.
+        if ($cartRules = $this->cartRepository->get($this->quote->getId())->getCartFixedRules()) {
```
vitaliyreznikov, line 239: Why we need to get quote from DB again?

vitaliyreznikov (commented):
According to DD there is some undifened offset issue.
https://app.datadoghq.com/logs?query=%40merchant_name%3A%22Magento%202%20RC%22%20%40trace_id%3ARoot%3D1-61e17103-68b82eaf655f79c15067100e&cols=core_host%2Ccore_service&event=AQAAAX5YoXhr8hq4zAAAAABBWDVZb2wzY0FBQnRiX2N5cFpJeXpnQUE&index=&messageDisplay=inline&stream_sort=desc&from_ts=1641803369128&to_ts=1642408169128&live=true

Does this PR help with it?

johnmccombs1:
@dumega It looks like your fix doesn't resolve the issue in the Anisa task. Can you provide steps to replicate the issue? 

@vitaliyreznikov I added the fix for the "undefined offset" issue. The "undefined offset" issue just occurs if customer checkout with free shipping and whole cart shipping discount. 
See here: https://github.com/BoltApp/bolt-magento2/pull/1502

Helper/Cart.php
```diff
@@ -2863,11 +2875,18 @@ public function getSaleRuleDiscounts($quote)
...
+            $cartFixedRules = $address->getCartFixedRules();
             if ($extensionSaleRuleDiscounts && is_array($extensionSaleRuleDiscounts)) {
                 foreach ($extensionSaleRuleDiscounts as $value) {
                     /* @var \Magento\SalesRule\Api\Data\DiscountDataInterface $discountData */
                     $discountData = $value->getDiscountData();
-                    $saleRuleDiscountsDetails[$value->getRuleID()] = $discountData->getAmount();
+                    $salesRuleId = $value->getRuleID();
+                    if (!empty($cartFixedRules) && array_key_exists($salesRuleId, $cartFixedRules)) {
+                        $rule = $this->ruleRepository->getById($salesRuleId);
+                        $saleRuleDiscountsDetails[$salesRuleId] = $discountData->getAmount() + $rule->getDiscountAmount() - $cartFixedRules[$salesRuleId];
```
johnmccombs1, line 2886: Could you explain why you use this logic? That would be good if you can refer some Magento core logic here.  I'm having a little worried since the split discounts feature is causing a lot of bugs.

Model/Api/Tax.php
danac-gs: @dumega ---^

Helper/Cart.php
vitaliyreznikov: @dumega Could you please answer here?

Model/Api/Tax.php
vitaliyreznikov: @dumega Could you please explain?

Model/Api/DiscountCodeValidation.php
```diff
@@ -163,8 +163,17 @@ private function handleRequest()
...
+                    // If a customer applies a cart rule (fixed amount for whole cart and apply to shipping) via the cart pgae,
+                    // the function Magento\SalesRule\Helper\CartFixedDiscount::calculateShippingAmountWhenAppliedToShipping does not return correct value for tax calculation,
+                    // it is because the $address->getShippingAmount() still returns shipping amount of last selected shipping method.
+                    // So we need to correct the shipping amount.   
+                    $shippingCost = CurrencyUtils::toMajor($requestArray['cart']['shipments'][0]['cost']['amount'], $immutableQuote->getQuoteCurrencyCode());
+                    $immutableQuote->getShippingAddress()->setShippingAmount($shippingCost);
+            }
         }
-
+
```
vitaliyreznikov, line 176: nit: remove blanks

Model/Api/DiscountCodeValidation.php
```diff
@@ -163,8 +163,17 @@ private function handleRequest()
         // Set the shipment if request payload has that info.
         if (!empty($requestArray['cart']['shipments'][0]['reference'])) {
             $this->setShipment($requestArray['cart']['shipments'][0], $immutableQuote);
+            $this->setShippingAssignments($immutableQuote);
+            if (!$immutableQuote->isVirtual() && $this->cartHelper->checkIfQuoteHasCartFixedAmountAndApplyToShippingRule($immutableQuote)) {
+                    // If a customer applies a cart rule (fixed amount for whole cart and apply to shipping) via the cart pgae,
```
vitaliyreznikov, line 168: typo: cart pgae => cart page

Model/Api/UpdateCartCommon.php
```diff
@@ -398,6 +410,25 @@ public function createPayloadForVirtualQuote($quote, $requestData)
         ];
         return json_encode((object)$payload);
     }
+    
+    /**
+     * Create shipping assignment for shopping cart
+     *
+     * @param CartInterface $quote
+     */
+    public function setShippingAssignments($quote)
```
vitaliyreznikov, line 419: This function sets cart extensions as well. Is it better to rename it?

Helper/Cart.php
dumega: @johnmccombs1  @vitaliyreznikov  When there are multiple "Fixed amount discount for whole cart" sale rule applied to cart, M2 core has specific logic (https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/SalesRule/Model/Rule/Action/Discount/CartFixed.php#L116 and https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/SalesRule/Model/Rule/Action/Discount/CartFixed.php#L156) to calculate the discount applied to shipping amount, as a result, the actual discount amount of sale rule does not equal to its original discount amount.

eg.
There are two "Fixed amount discount for whole cart"  sale rules A and B. the original discount amount of A is $5.06 while it is $6.06 for B, they are both applied to cart. In some contexts, although the cart subtotal and shipping amount are both larger than $12.12, the actual discount total of cart is $12.06 other than $12.12 (5.06+6.06).
When looking into the M2 core, I found that discount from each sale rule has two parts, one is applied to line items and another one is applied to shipping, we can get the amount of first part by **$discountData->getAmount()**, the amount  of second part by **$rule->getDiscountAmount() - $cartFixedRules[$salesRuleId]**
image

Model/Api/Tax.php
dumega: @vitaliyreznikov  @danac-gs Cause only in this way, we can get the up-to-date info of rules applied to cart, that's also M2 does in core code when calculating totals https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Checkout/Model/TotalsInformationManagement.php#L48

Model/Api/DiscountCodeValidation.php
dumega: Fixed

Model/Api/DiscountCodeValidation.php
dumega: Fixed

Model/Api/UpdateCartCommon.php
dumega: This is actually a private function copied from M2 core https://github.com/magento/magento2/blob/3c5ff49c11eec2d4510b25571acf0d30b804cdf1/app/code/Magento/Persistent/Model/QuoteManager.php#L225, so I follow its name.

vitaliyreznikov: Ok, makes sense

vitaliyreznikov (approved):
LGTM

PR #1506, opened 2022-01-31
Allow using carts/mine/payment-information for admin users

vitaliyreznikov:
carts/mine/payment-information is magento endpoint which works for applying payment method to quote and create order.
This is a way how to create order for logged-in magento user.
This PR allows to use payment-information endpoint by admin integration users

#changelog Allow using carts/mine/payment-information for admin users


mikeshutov-bolt (approved):
LGTM clean.

PR #1510, opened 2022-02-03
Magento API: fix redirect for backoiffice orders

garmashdev:
For API-driven approach we need to set bolt checkout type in quote to backoffice.
It allows us to do proper redirect from admin page.

In the legacy approach we set this field for immutable quote in createImmutableQuote() method.

#changelog Magento API: fix redirect for backoiffice orders


vitaliyreznikov (approved):
Thank you for helping me create this PR Yaroslav

PR #1461, opened 2021-11-17
Use discount name instead of discount description for automatic coupons

johnmccombs1:
# Description
The automatic discount description is shown as part of a discount tag on the Bolt modal. That discount description isn't a required field. When it's empty, the discount tag will be very indescriptive and confuse the customers.

This PR resolves the issue by using the required discount name as part of the discount tag. 

Fixes: https://app.asana.com/0/0/1201256841759895/1201384293294758/f

#changelog Use discount name instead of discount description for automatic coupons

# Type of change

- [x] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?
Please validate that you have tested your change in at least one of the following areas:

- [x] Successfully tested locally (or docker image)
- [x] Successfully tested on a staging or sandbox server
- [ ] Successfully tested on a merchant's staging server

# For PR Reviewer 
- [ ] Reviewed unit tests to make sure we are using real components rather than mocks as much as possible?
- [ ] For any major change (observer, new Bolt feature, core Magento interaction) we must add a feature switch, did you verify this?

# Checklist:

- [ ] My code follows the style guidelines of this project.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] I have created or modified unit tests to sufficiently cover my changes.
- [ ] I have added my Jira ticket link and provided a changelog message.

PR #1516, opened 2022-02-10
Use rule name as coupon description only under feature switch

vitaliyreznikov:
Reverting https://github.com/BoltApp/bolt-magento2/pull/1461/ and use rule name as coupon description only under feature switch.
This fix affects merchants who use automated discounts with empty discount descriptions.
Before this fix we showed to shoppers **rule name** and after fix - word "`Discount`"

The reason under the fix: some merchants save internal information in the discount name field.

#changelog Use rule name as coupon description only under feature switch

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -2231,7 +2231,10 @@ class Cart extends AbstractHelper
                         break;
                     case RuleInterface::COUPON_TYPE_NO_COUPON:
                     default:
-                        $description = trim($rule->getDescription()) ? $rule->getDescription() : $rule->getName();
+                        $description = $rule->getDescription();
+                        if (!$description && $this->deciderHelper->isUseRuleNameIfDescriptionEmpty()) {
+                            $description = $rule->getName();
+                        }
                         $discounts[] = [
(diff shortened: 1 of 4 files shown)


mikeshutov-bolt (approved):
LGTM when tests pass.

Helper/Cart.php
```diff
@@ -2231,7 +2231,10 @@ public function collectDiscounts(
                         break;
                     case RuleInterface::COUPON_TYPE_NO_COUPON:
                     default:
-                        $description = trim($rule->getDescription()) ? $rule->getDescription() : $rule->getName();
+                        $description = $rule->getDescription();
+                        if (!$description && $this->deciderHelper->isUseRuleNameIfDescriptionEmpty()) {
```
johnmccombs1, line 2235: Missing function isUseRuleNameIfDescriptionEmpty in class deciderHelper ?

johnmccombs1 (changes requested):

Helper/Cart.php
johnmccombs1: LGTM

johnmccombs1 (approved):

PR #1520, opened 2022-02-15
Implement get order status endpoint

vitaliyreznikov:
Magento doesn't have API for get default order status per order state but we need to set status during webhook handling.
So need to have this simple GET endpoint.

#changelog Implement get order status endpoint


mikeshutov-bolt (approved):
LGTM

PR #1556, opened 2022-05-16
Discount calculation refactoring

garmashdev:
# Description
Refactored native discount calculation process.

Fixes: [(link ticket)](https://app.asana.com/0/1201931884901947/1202225720729002/f) . We have an issue in native code of discount calculation, which is described in the ticket abowe. When discount is more than subtotal we are sending to bolt data with not real discount applied but with rule full discount. In this case our discount amount is applying to shipment too which is incorrect. For example rule can have discount 5000$ and item price 20$, we are sending discount $5000 instead real discount applied 20$.  I've changed logic of this method(getSaleRuleDiscounts) to use default magento logic which is using in core collect totals method(Magento\SalesRule\Model\RulesApplier). In this case i'm able to have correct data of applied discounts per items and send those data to bolt by array: [ruleId => real discount applied].

#changelog Discount calculation refactoring

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:


vitaliyreznikov (commented):
@garmashdev Could you please add some explanation about your fix?

garmashdev:
@vitaliyreznikov PR description updated.

dumega (commented):
@garmashdev To fix this issue, you directly copy code from M2 core, right?
If this is for fixing specific bug of M2 core, I suggest we do it via plugin method to override [function getSaleRuleDiscounts](https://github.com/BoltApp/bolt-magento2/blob/2.25.1/Helper/Cart.php#L2880) and restrict the fix on specific M2 version only?
In this way, it does not effect M2 versions which already have fix and won't introduce new bug in our plugin.

garmashdev:
@dumega hi, it is not magento core bug, the bug in our magento https://github.com/BoltApp/bolt-magento2/blob/2.25.1/Helper/Cart.php#L2880 method.  On legacy Magento 2.4.3 we have the same issue which was confirmed from QA team. I added there magento core calculation logic(not a copy, as composition after investigation core methods) from core classes. thanks.

dumega:
> @dumega hi, it is not magento core bug, the bug in our magento https://github.com/BoltApp/bolt-magento2/blob/2.25.1/Helper/Cart.php#L2880 method. On legacy Magento 2.4.3 we have the same issue which was confirmed from QA team. I added there magento core calculation logic(not a copy, as composition after investigation core methods) from core classes. thanks.

But in the related ticket, it says this is caused by magento bug https://github.com/magento/magento2/issues/24422 , also I can not reproduce it in M2 2.4.3-p2 .

Do you have a screenshot of the settings of coupon?

garmashdev:
@dumega yes:
fixed

dumega (changes requested):
@garmashdev I run some test in my dev site with this PR, although it fixes the bug mentioned in the [ticket](https://app.asana.com/0/1201931884901947/1202225720729002/f), we face another issue happened before, your PR does not handle whole cart shipping discount properly. Please check the task https://app.asana.com/0/1124368832381302/1201663986351838/f for details.

Then could you dig into this issue again? Thanks

garmashdev:
@dumega hi, after discussion with @mikeshutov-bolt we decided to add a new option to add this pr calculation logic for special cases : "Magento Native Discount Calculation" with off default value.
Снимок экрана 2022-05-27 в 13 11 17

PR #1574, opened 2022-05-31
Bugfix: Unable to place guest order in admin

johnmccombs1:
# Description

**Issue:**
The administrator is unable to place a guest Bolt order in admin 

**Step to reproduce:**
1. Go to Magento Admin > Sales > Orders > Create New Order
2. In “Please select a customer” screen, choose “Create New Customer”
3. Fill the informations and checkout with Bolt

Expected result:
Order created successfully

Actual result:
Order creation failed
See here: https://prnt.sc/55bo4XXgVftt

Bolt is using Magento 2 API  endpoint (/rest/V1/guest-carts/:cartId/payment-information) to create orders for guest users. The problem is it needs a masked quote ID (see Magento core logic here: https://github.com/magento/magento2/blob/2.4.4/app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php#L165), which is only available for guest quotes in the frontend. In this case, the quote is created from the admin panel, so we will not have a masked quote id to use. 

So the solution is to set a Masked Id for the quote in case it is created from the backend

Fixes: https://app.asana.com/0/1201992942252703/1202286654991143

#changelog

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:


vitaliyreznikov (commented):
Code looks good, but could you please update tests rather then remove them?

johnmccombs1:
@vitaliyreznikov  I added the test for the case that quote doesn't mask quote id, we regenerate mask quote id.

johnmccombs1:
For the test cases I removed, I don't think it's it is necessary

vitaliyreznikov (approved):

PR #1580, opened 2022-06-07
Update user's quote during SSO login

vitaliyreznikov:
When we user logins with existing cart via bolt SSO the cart is not updated properly which leads to some issue.
In this PR we find quote_id in the session and update it

#changelog Update user's quote during SSO login

https://app.asana.com/0/1201931884901947/1202396691087893/f


mikeshutov-bolt (approved):
LGTM

danac-gs (approved):

PR #1583, opened 2022-06-08
Fix compatibility issues with PHP 8.1

dumega:
# Description
There are still some deprecated functionality errors in Bolt plugin once upgrading PHP to 8.1

`Deprecated Functionality: trim(): Passing null to parameter #1 ($string) of type string is deprecated`

Actually most of such errors are related to the functions in Bolt\Boltpay\Helper\Config, that the function is expected to return string but eventually null is returned. So in this PR we force to return empty string if any null value is detected.

Updates: using PHP 8 compatibility checker, we found some more compatibility issues and fix them in this PR as well

Fixes: https://app.asana.com/0/951157735838091/1202225151737588/f

#changelog Fix compatibility issues with PHP 8.1

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Config.php
+++ b/Helper/Config.php
@@ -1526,6 +1526,7 @@ class Config extends AbstractHelper
                 }
             }
         }
+        return '';
     }
 
     /**
--- a/Helper/Config.php
+++ b/Helper/Config.php
@@ -586,7 +586,7 @@ class Config extends AbstractHelper
         if ($this->isSandboxModeSet($storeId)) {
             return $this->getApiUrlFromAdditionalConfig($storeId) ?: $this->getCustomURLValueOrDefault(self::XML_PATH_CUSTOM_API, self::API_URL_SANDBOX);
         }
-        return self::API_URL_PRODUCTION ?: '';
+        return self::API_URL_PRODUCTION;
     }
 
     /**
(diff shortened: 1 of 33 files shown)


Helper/Config.php
```diff
@@ -754,7 +754,7 @@ public function getTitle($storeId = null)
      */
     public function getApiKey($storeId = null)
     {
-        return $this->getEncryptedKey(self::XML_PATH_API_KEY, $storeId);
+        return $this->getEncryptedKey(self::XML_PATH_API_KEY, $storeId) ?: '';
```
vitaliyreznikov, line 757: If API key is empty we can not proceed. Maybe check for empty and throw exception?

Helper/Config.php
```diff
@@ -586,7 +586,7 @@ public function getApiUrl($storeId = null)
         if ($this->isSandboxModeSet($storeId)) {
             return $this->getApiUrlFromAdditionalConfig($storeId) ?: $this->getCustomURLValueOrDefault(self::XML_PATH_CUSTOM_API, self::API_URL_SANDBOX);
         }
-        return self::API_URL_PRODUCTION;
+        return self::API_URL_PRODUCTION ?: '';
```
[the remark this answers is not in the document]
dumega: Make sense. Fixed

Helper/Config.php
dumega: This is helper function to return configuration value. I think only the functions that call it can decide whether to throw exception or not.

vitaliyreznikov (approved):
Approved, see a comment about another possible way

Helper/Config.php
dumega: @vitaliyreznikov  We faced similar issue with the function `explode` (Fixed in the [PR 1572](https://github.com/BoltApp/bolt-magento2/pull/1572)) so we would better fix it within the source rather than caller .

PR #1607, opened 2022-07-04
Add scope for Magento Commerce Gift Card plugin

vitaliyreznikov:
Module Magento Commerce Gift Card plugin introduces its own resource "Magento_GiftCardAccount::customer_giftcardaccount" 

#changelog Add scope for Magento Commerce Gift Card plugin


mikeshutov-bolt:
LGTM i think this is one of those changes that can be squeezed into the next release as well since theres nothing to test here but ill let you decide if its worth/not.

PR #1648, opened 2022-09-10
Limit the results of stores by shipping country; change the unit to mile

elielloyd:


vitaliyreznikov:
@dumega Could you please review?

dumega:
@minhdle did implement that feature on Bolt FE but it was only applied to American merchant, then he can update it to include Canadian as well. So I think we do not need to do anything in plugin side.

dumega (approved):

danac-gs (approved):

PR #1650, opened 2022-09-15
Skip cart discount total mismatch when needed

dumega:
# Description
We face cases that the discount total is different between Bolt cart and M2 cart when the native API is enabled, it could be caused by various reasons. For now we just add a new feature switch M2_SKIP_CART_DISCOUNT_TOTAL_MISMATCH to skip such mismatch when needed.

Fixes: https://app.asana.com/0/1201931884901947/1202975415590108/f

#changelog Skip cart discount total mismatch when needed

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -2325,12 +2325,9 @@ class Cart extends AbstractHelper
             if ($this->deciderHelper->isAPIDrivenIntegrationEnabled()
                 && $this->deciderHelper->isSkipCartDiscountTotalMismatch()
                 && $amount) {
-                usort($discounts, function ($a, $b) {
-                    return $b['amount'] <=> $a['amount'];
-                });
                 $discounts[0]['amount'] += $amount;
                 $totalAmount -= $amount;
-                $this->bugsnag->notifyError('Cart discount total mismatch', "Skip cart discount total mismatch. Actual Mismatch {$amount}");
+                $this->bugsnag->notifyError('Cart discount total mismatch', "Skip cart discount total mismatch. Actual Mismatch {$amount}. Discounts details: " . var_export($discounts, true));
             }
         }
         /////////////////////////////////////////////////////////////////////////////////
--- a/Helper/Cart.php
+++ b/Helper/Cart.php
@@ -2279,6 +2279,7 @@ class Cart extends AbstractHelper
         // check if getCouponCode is not null
         /////////////////////////////////////////////////////////////////////////////////
         if (($amount = abs((float)$address->getDiscountAmount())) || $quote->getCouponCode()) {
+            $amount = CurrencyUtils::toMinor($amount, $currencyCode);
             $ruleDiscountDetails = $this->getSaleRuleDiscounts($quote);     
             list($ruleDiscountDetails, $discounts, $totalAmount) = $this->eventsForThirdPartyModules->runFilter('filterQuoteDiscountDetails', [$ruleDiscountDetails, $discounts, $totalAmount], $quote);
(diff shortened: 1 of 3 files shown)


Helper/Cart.php
```diff
@@ -2318,8 +2319,19 @@ public function collectDiscounts(
 
                         break;
                 }
+                $amount -= $roundedAmount;
                 $totalAmount -= $roundedAmount;
             }
+            if ($this->deciderHelper->isAPIDrivenIntegrationEnabled()
+                && $this->deciderHelper->isSkipCartDiscountTotalMismatch()
+                && $amount) {
+                usort($discounts, function ($a, $b) {
```
vitaliyreznikov, line 2328: could we not change discount orders here?
It would be nice to have initial order.
It also affects on the order of discounts displaying in bolt modal

dumega: @vitaliyreznikov  I have removed the usort and add difference value to the amount of first element by default, actually we do not know which sale rule causes the difference, so it does not matter.

Helper/Cart.php
```diff
@@ -2318,8 +2319,19 @@ public function collectDiscounts(
...
+                    return $b['amount'] <=> $a['amount'];
+                });
+                $discounts[0]['amount'] += $amount;
+                $totalAmount -= $amount;
+                $this->bugsnag->notifyError('Cart discount total mismatch', "Skip cart discount total mismatch. Actual Mismatch {$amount}");
```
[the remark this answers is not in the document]
dumega: I add the discounts array to the log content.

Helper/Cart.php
vitaliyreznikov: In Andrew Christian case, the first discount can be $0.10 and the second $100
It's more logical to change the biggest discount

Helper/Cart.php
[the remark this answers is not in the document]
vitaliyreznikov: I think it's not enough for debugging. Could we have discount amount before rounding as well?

vitaliyreznikov (commented):
few more comments

Helper/Cart.php
dumega: Done

Helper/Cart.php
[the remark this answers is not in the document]
dumega: Done

Helper/Cart.php
```diff
@@ -2318,8 +2319,21 @@ public function collectDiscounts(
...
+                && $this->deciderHelper->isSkipCartDiscountTotalMismatch()
+                && $amount) {
+                uasort($discounts, function ($a, $b) {
+                    return $b['amount'] <=> $a['amount'];
+                });
+                $firstKey = array_key_first($discounts);
+                $discounts[$firstKey]['amount'] += $amount;
+                ksort($discounts);
+                $totalAmount -= $amount;
+                $this->bugsnag->notifyError('Cart discount total mismatch', "Skip cart discount total mismatch. Actual Mismatch {$amount}. Discounts details: (before rounding)" . var_export($ruleDiscountDetails, true) . " (after rounding) " .var_export($discounts, true));
```
vitaliyreznikov, line 2335: Have we tested that $ruleDiscountDetails is not deeply nested object and var_export($ruleDiscountDetails, true) works well?

dumega: $ruleDiscountDetails only contains sale rule id and its discount amount.

vitaliyreznikov (approved):

PR #1653, opened 2022-09-28
Created custom API endpoiont to enable or disable Route

KyerdnA:
# Description
Created custom endpoint for enabling or disabling Route Shipping insurance

Fixes: https://app.asana.com/0/1201931884901947/1202808401252801

#changelog Created custom API endpoiont to enable or disable Route

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:


Model/Api/RouteInsuranceManagement.php
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    private function setRouteIsInsuredToQuote($routeIsInsured, $cartId) {
+        $quote = $this->cartHelper->getQuoteById($cartId);
+        if (!$quote)
+        {
+            throw new WebapiException(__('Quote does not found by given ID'), 0, WebapiException::HTTP_NOT_FOUND);
+        }
+        $quote->collectTotals();
```
vitaliyreznikov, line 120: We actually don't change the quote, only save it. Could you explain the logic please?

Model/Api/RouteInsuranceManagement.php
```diff
@@ -0,0 +1,148 @@
...
+use Bolt\Boltpay\Helper\Bugsnag;
+use Exception;
+use Magento\Framework\Webapi\Exception as WebapiException;
+use Magento\Framework\Webapi\Rest\Response;
+use Magento\Framework\Module\Manager;
+use Magento\Framework\Serialize\SerializerInterface as Serialize;
+use Bolt\Boltpay\Helper\Cart as CartHelper;
+use Magento\Quote\Model\QuoteRepository;
+
+class RouteInsuranceManagement implements \Bolt\Boltpay\Api\RouteInsuranceManagementInterface
```
vitaliyreznikov, line 29: Another option: use the existing `interface CartManagementInterface` and have all operations under quotes inside of the interface

Plugin/WebapiRest/Route/Route/Helper/DataPlugin.php
```diff
@@ -0,0 +1,89 @@
...
+        \Bolt\Boltpay\Model\ThirdPartyModuleFactory $boltRouteQuoteClient
+    ){
+        $this->request = $request;
+        $this->quoteRepository = $quoteRepository;
+        $this->boltRouteQuoteClient = $boltRouteQuoteClient;
+    }
+
+    public function afterIsInsured(\Route\Route\Helper\Data $subject, $result): bool
+    {
+        $requestParams = $this->request->getParams();
```
vitaliyreznikov, line 54: Finding parameter in request is a little bit counterintuitive.
What do you think about having a set method in this class, like SetIsInsured, and call it directly from API class?
This is also a way to show dependency between API class and this plugin

Model/Api/RouteInsuranceManagement.php
```diff
@@ -0,0 +1,148 @@
...
+        } catch (WebapiException $e) {
+            $this->bugsnag->notifyException($e);
+            throw $e;
+        } catch (Exception $e) {
+            $this->bugsnag->notifyException($e);
+            throw new WebapiException(__($e->getMessage()), 0, WebapiException::HTTP_INTERNAL_ERROR);
+        }
+    }
+
+    private function setRouteIsInsuredToQuote($routeIsInsured, $cartId) {
```
vitaliyreznikov, line 114: the function name is a little bit misleading because the function itself doesn't set something to quote.
Also, there are two purposes for the function: trigger quotes total collection and generate API answer.

I suggest splitting it into two parts and call the first one somethings like collectTotalsAndSave quote.
It makes sense to mention in comments why we want quotes to be recollected.

Model/Api/RouteInsuranceManagement.php
```diff
@@ -0,0 +1,145 @@
...
+    public function execute($cartId, $routeIsInsured)
+    {
+        try {
+            if (!$this->isModuleEnabled()) {
+                $responseBody = [
+                    'message' => sprintf("%s is not installed on merchant's site", self::ROUTE_MODULE_NAME)
+                ];
+                $httpResponseCode = self::RESPONSE_FAIL_STATUS;
+            } else {
+                $responseBody = $this->collectTotalsAndSaveQuote($routeIsInsured, $cartId);
```
vitaliyreznikov, line 100: Is it possible insted of using plugins update the session before this line?

KyerdnA: To update the session we'll need to know session_id from FE customer, this data is available in browser-server flow using cookie
In theory we can save session data in db for further use, but anyway it'll require to have any identifier of user to access correct session data. 
I think we need to discuss and investigate possibilities of supporting sessions through magento API more deeply, as it's big customization

vitaliyreznikov: I mean even for API call we have session in place. And if we write into the session and initiate collect total route will read from this session.
Looks like simpler that approach with plugins

vitaliyreznikov (approved):

PR #1655, opened 2022-10-04
Amasty extra fee support: use default placeholder image

vitaliyreznikov:
In PR https://github.com/BoltApp/bolt-magento2/pull/1632 we fixed placeholder image URL but only in one place, for an add-on.
In this PR we fix the same URL in the other place, for a cart item.

https://app.asana.com/0/1200879031426307/1202780668772807/f

#changelog Amasty extra fee support: use default placeholder image


mikeshutov-bolt (approved):
LGTM, question is there another image we could use from Amasty or no?

johnmccombs1 (approved):

vitaliyreznikov:
Amasty extension could be used for different purposes, so I think the default image is the best we can do.

PR #1709, opened 2022-12-05
M2 API route support: set customer object to the session

vitaliyreznikov:
Because the route plugin doesn't support API order creation (it saves data in the checkout session), [we built a workaround in our plugin](https://github.com/BoltApp/bolt-magento2/blob/2d88e4f032ba2736af6ed85574a361c302a018a1/Plugin/WebapiRest/Magento/Checkout/Api/PaymentInformationManagementPlugin.php#L77), which emulates the session data.

But if between our code and actual order creation, any third party module read the quote from the session [this code cleaned it up](https://github.com/magento/magento2/blob/78dd0652658ced5b0e93cc6629e774bf1958c5e9/app/code/Magento/Checkout/Model/Session.php#L257) (introduced in Magento 2.3.2)

Solution: set in session customer object as well to prevent quote clean up.

Asana: https://app.asana.com/0/1201931884901947/1203441506208729/f

#changelog M2 API route support: set customer object to the session


KyerdnA (approved):
LGTM

PR #1731, opened 2022-12-22
SSO: fix safari log-in by Bolt Assistant

garmashdev:
# Description
Fixed 2 issues related to SSO
- safari: unable to add products to card(session expired) after log-in by "Bolt Shopper Assistant":

When we are logged-in by SSO without page redirect (by using Bolt Shopper Assistant for example) in this case Magento removing cookie by backend [here](https://github.com/magento/magento2/blob/5844ade68b2f9632e3888c81c946068eba6328bb/app/code/Magento/PageCache/Observer/FlushFormKey.php). But on safari without redirect the cookie is removed from browser (you can't see cookie on the devtool) but not in window.document object(we can see the cookie "form_key" there by checking object in browser console: "document.cookie") therefore all future requests is being called without "form_key" which is the reason of  "session expire" issue.
By normal flow Magento updates that cookie in files:
[2.3.0](https://github.com/magento/magento2/blob/44a7b6079bcac5ba92040b16f4f74024b4f34d09/app/code/Magento/PageCache/view/frontend/web/js/page-cache.js#L115)
[2.4.0](https://github.com/magento/magento2/blob/5844ade68b2f9632e3888c81c946068eba6328bb/app/code/Magento/PageCache/view/frontend/web/js/form-key-provider.js#L87) 
Magento js above sets new cookie only in case if cookie is not exist, but in our issue case Magento js file is able to read cookie from document object (because cookie was not removed by safari) and doesn't update browser cookie which is the root cause of "session expired error".

- Fixed guest cart merging issue for Magento 2.3.0

In Magento 2.3.0 the [code](https://github.com/magento/magento2/blob/44a7b6079bcac5ba92040b16f4f74024b4f34d09/app/code/Magento/Quote/Model/QuoteManagement.php#L297) doesn't have "merging" part as we have [here](https://github.com/magento/magento2/blob/5844ade68b2f9632e3888c81c946068eba6328bb/app/code/Magento/Quote/Model/QuoteManagement.php#L337) in Magento 2.4.0
Fixes: https://app.asana.com/0/1201931884901947/1203517869067339/f

Tested locally in chrome/safari with guest cards without/with products log-in by normal sso flow, bolt modal, bolt shopper assistant

#changelog fix sso safari flow by log-in by "Bolt Shopper Assistant"

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Model/Api/OAuthRedirect.php
+++ b/Model/Api/OAuthRedirect.php
@@ -45,6 +45,10 @@ use Magento\Store\Model\StoreManagerInterface;
 use Magento\Customer\Model\EmailNotification;
 use Magento\Quote\Api\CartManagementInterface;
 use Magento\Quote\Api\CartRepositoryInterface;
+use Magento\Framework\Session\Config\ConfigInterface as SessionConfigInterface;
+use Magento\Framework\Data\Form\FormKey;
+use Magento\Framework\App\PageCache\FormKey as CookieFormKey;
+use Magento\Framework\Exception\LocalizedException;
 
 class OAuthRedirect implements OAuthRedirectInterface
 {
@@ -158,6 +162,12 @@ class OAuthRedirect implements OAuthRedirectInterface
      */
     private $cartRepository;
 
+    private $sessionConfig;
+
+    private $formKey;
+
+    private $cookieFormKey;
+
     /**
      * @param Response $response
      * @param DeciderHelper $deciderHelper
@@ -200,7 +210,10 @@ class OAuthRedirect implements OAuthRedirectInterface
         Bugsnag $bugsnag,
         EmailNotification $emailNotification,
         CartManagementInterface $cartManagement,
-        CartRepositoryInterface $cartRepository
+        CartRepositoryInterface $cartRepository,
+        SessionConfigInterface $sessionConfig,
+        FormKey $formKey,
+        CookieFormKey $cookieFormKey
     ) {
         $this->response = $response;
         $this->deciderHelper = $deciderHelper;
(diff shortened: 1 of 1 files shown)


Model/Api/OAuthRedirect.php
```diff
@@ -391,6 +407,18 @@ public function login($code = '', $scope = '', $state = '', $reference = '')
                 if ($checkoutSession) {
                     $quote = $checkoutSession->getQuote();
                     if ($quote->getId()) {
+                        try {
+                            // we should merge quote here for magento 2.3.0 version, because
```
vitaliyreznikov, line 411: 2.3.x, correct?

Model/Api/OAuthRedirect.php
```diff
@@ -474,7 +502,47 @@ private function linkLoginAndRedirect($externalID, $customerID)
...
+        try {
+            // we should force to set new "form_key" cookie here to prevent safari "session expired" issue
+            // when we are logged-in by SSO without page redirect (by using Bolt Shopper Assistant for example)
+            // in this case magento removing cookie by backend here:
+            // https://github.com/magento/magento2/blob/5844ade68b2f9632e3888c81c946068eba6328bb/app/code/Magento/PageCache/Observer/FlushFormKey.php
+            // but on safari without redirect the cookie is removed from browser (you can't see cookie on the devtool) but not in window.document object
+            // (we can see the cookie "form_key" there by checking object in browser console: "document.cookie")
+            // therefore all future requests is being called without "form_key" which is the reason of  "session expire" issue.
+            // By normal flow Magento updates that cookie in file:
+            // 2.3.0 https://github.com/magento/magento2/blob/44a7b6079bcac5ba92040b16f4f74024b4f34d09/app/code/Magento/PageCache/view/frontend/web/js/page-cache.js#L115
```
vitaliyreznikov, line 514: The version below 2.4.0 affected, correct? By the way, what about 2.2.x, we did not drop it's support yet.

vitaliyreznikov (approved):

Model/Api/OAuthRedirect.php
garmashdev: updated comment, under 2.2.x the same code

Model/Api/OAuthRedirect.php
garmashdev: updated comment

PR #1739, opened 2023-01-19
Remove subscription frontend code which is going to be part of bolt backend

vitaliyreznikov:
Asana:
https://app.asana.com/0/1201931884901947/1203780995341070/f


Nitsorn (commented):
if we are removing this, where will this script be loaded, @vitaliyreznikov ?
We need to call
```
BoltCheckout.configureProductFrequencySelector(cart, persistedFrequencyValues, callbacks);
``` 
to load things inside the div.

Are you expecting us to do something like https://github.com/BoltApp/bolt-magento2/compare/vitaliy/add-to-cart-prototype?expand=1? We would love more context on what solution you propose if we remove this out. 

**Is this related to the script manager at all?**

Tell us how to help you better.

view/frontend/templates/subscription_freq_selector_product_page.phtml
```diff
@@ -21,151 +21,3 @@ if (!$block->isBoltProductPageForFrequencySelector()) { return;
...
-        };
-
-        var cart;
-        var setupProductPage = function () {
-            cart = {
-                items: getItemsData()
-            };
-
-            // if connect.js is not loaded postpone until it is
-            whenDefined(window, 'BoltCheckout', function(){
```
smahbod2014, line 157: Without this, who will be calling our `configureProductFrequencySelector()` function to mount to our `bolt-freq-selector` div?

PR #1753, opened 2023-02-23
Optimize code to avoid repeatable tax calls

KyerdnA:
# Description
Some third-party extensions which are using custom Tax collectors faced issue when some bolt api requests triggered multiple tax requests. 

Fixes:
https://app.asana.com/0/1201931884901947/1203931573764626

#changelog Optimize code to avoid repeatable tax calls

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Model/Api/OAuthRedirect.php
+++ b/Model/Api/OAuthRedirect.php
@@ -461,7 +461,11 @@ class OAuthRedirect implements OAuthRedirectInterface
             $immutableQuote->setData($data);
             $immutableQuote->setCustomer($customer);
             $immutableQuote->setCustomerIsGuest(false);
-            $this->cartHelper->saveQuote($immutableQuote);
+            // Magento quote repository save() method triggers collectTotals() method for quote.
+            // Which is triggering some third-party tax extensions to request & collect taxes.
+            // In order to avoid repeatable tax requests from third-party extensions
+            // we use resource model to save quote instead of repositoryFix
+            $immutableQuote->save();
         }
     }
 


Model/Api/OAuthRedirect.php
```diff
@@ -461,7 +461,11 @@ private function updateImmutableQuotes($parentQuote, $customer)
             $immutableQuote->setData($data);
             $immutableQuote->setCustomer($customer);
             $immutableQuote->setCustomerIsGuest(false);
-            $this->cartHelper->saveQuote($immutableQuote);
+            // Magento quote repository save() method triggers collectTotals() method for quote.
+            // Which is triggering some third-party tax extensions to request & collect taxes.
+            // In order to avoid repeatable tax requests from third-party extensions
+            // we use resource model to save quote instead of repositoryFix
+            $immutableQuote->save();
```
vitaliyreznikov, line 468: You are changing the function which we use when login via SSO, correct?
Do you want to change behavior for the SSO case only?

vitaliyreznikov: And if we need to do the same change in other places of our code - the change looks too big for me.
Can we instead do the fix specific for taxjar? I.e. create a plugin for their module which put the answer into cache?

If the global change is necessary please do it under the feature switch disabled by default.

danac-gs: @vitaliyreznikov  Here we only update (old) immutable quotes with the newly logged in customer data. No need to do / trigger anything else, especially not collecting totals for all of them, which may be quite an overhead, not only with TaxJar but with other parts of the system that are hooked to collecting totals. I thing the change above is OK.

vitaliyreznikov: I am still on the fence here. We change the quote significantly, made it logged-in from guest. Why don't recollect quotes? It may be necessary for some third-party modules.

@KyerdnA If you think that your change is necessary please do it under feature switch enabled by default.

danac-gs: @vitaliyreznikov you are basically right. The change from guest to logged-in may sometime have other implications and the totals could be changed based on the customer group. However, if that happens, the quote is not anymore in sync with what's on Bolt end and those orders would probably fail (as of right now, unless in the API approach we somehow trigger collect totals again, which would also work with the @KyerdnA's solution). So we'd just replace one issue with another. There may be another approach. If I remember correctly, the merchant reported over 60 immutable quotes generated for some parent quotes. I am wondering if all of them are different (meaning the customer added 60 items to cart, one by one) or something is wrong in our plugin with reusing (deleting?) the identical immutable quotes. Anyway, maybe it's time to completely remove immutable quotes from  the code base.

vitaliyreznikov: We don't use immutable quotes with API approach

KyerdnA: From my investigations, most tax calls are coming from shipping/methods endpoint in m2
We're using old plugin functionality there & calling collectTotals() function 7 times during this call & this is why taxjar being triggered that much
I've tried to make smth here, but because of the specifics of the code, we won't get rid of these repeatable collectTotals calls & merchant will need to keep their customization (the one that blocks repeatable tax request). So there is no good solution for this shipping/methods request for now, as our custom endpoint is simply based on this collectTotals calls to keep valid data inside the quote in all cases. 
So for now i've added the workaround for ssh endpoint which also was mentioned by merchant to use resource model instead of repository(even considering that it's not the best approach based on adobe code standarts) as it'll reduce amount of unneeded tax requests

PR #1756, opened 2023-02-27
Add PHPStan (static analysis tool) to circleCI pipeline

vitaliyreznikov:
- add PHPStan - a static analysis tool for PHP code - into the circle ci pipeline
- add PHPStan to the same job where we run PHP CodeSniffer
- add magento installation and code generation into the job so PHPStan can see magento classes, including fabric generated  classes
- add mocks for third party classes we use.

- fix few minor issues with code

Asana:
https://app.asana.com/0/1201931884901947/1204028793252855/f

#changelog Enable PHP lint


garmashdev (approved):

Test/scripts/phpcs_meqp2.sh
```diff
@@ -4,14 +4,34 @@ set -e
 set -u
 set -x
 
-sudo composer self-update -q
 composer show -i
 
-echo "{\"http-basic\":{\"repo.magento.com\":{\"username\":\"${MAGENTO_PUBLIC_KEY}\",\"password\":\"${MAGENTO_PRIVATE_KEY}\"}}}" > $HOME/.composer/auth.json
```
mikeshutov-bolt, line 10: Why are we removing this and the self update? Trying to understand this.

Test/scripts/phpcs_meqp2.sh
```diff
@@ -4,14 +4,34 @@ set -e
...
+
 composer create-project magento/magento-coding-standard --stability=dev magento-coding-standard
 
 cd magento-coding-standard
 
-vendor/bin/phpcs ../project --standard=Magento2 --colors --severity=10
+vendor/bin/phpcs ../project --standard=Magento2 --colors --severity=10 -p
+
+cd ../project
+export MAGENTO_VERSION="2.3.0"
```
mikeshutov-bolt, line 18: Why are we exporting this particular version?

Test/scripts/phpcs_meqp2.sh
vitaliyreznikov: We did not need it even before the change

.circleci/config.yml
```diff
@@ -227,7 +227,11 @@ jobs:
 
   phpcs:
     docker:
-      - image: cimg/php:7.2.34
+      - image: boltdev/m2-installed-plugin-ci-php72:2.3.0-v1
```
mikeshutov-bolt, line 230: Why are we changing to this installed image?

vitaliyreznikov: Before the change, we used this job only for PHP CodeSniffer. It doesn't require magento installed
PHPStan requires magento installed (it checked all magento classes)

I decided to use one job for both services because even together they finish before unit tests

mikeshutov-bolt: Just a question then. Doesnt this mean its only checking M2.3? or are we covered for newer versions regardless.

Test/scripts/phpcs_meqp2.sh
vitaliyreznikov: We already have script `Test/scripts/install_magento.sh`, which requires a specific magento version set before it runs. It doesn't matter which exact version we use.

.circleci/config.yml
mikeshutov-bolt: I guess its not super important actually. Just wanted to make sure we thought about all of these.

vitaliyreznikov: We only need Magento to check that we use the correct (existing) Magento classes in the plugin code.
So it's better to check with the oldest Magento version - this way, we can see which classes were created only in new Magento versions.

mikeshutov-bolt (approved):

dumega (approved):
LGTM, @vitaliyreznikov just a question, does it support PHP compatibility checks? If yes, do you enable it?

vitaliyreznikov:
@dumega Let's discuss in slack

PR #1768, opened 2023-03-08
Resolve issues with multiple calls to whenDefined

danac-gs:
# Description
We have a bug in [whenDefined](https://github.com/BoltApp/bolt-magento2/blob/master/view/frontend/web/js/utils/when-defined.js). When called multiple times it can return false positive result. The issue manifested on Headlights Depot PPC [page](https://shop.headlightsdepot.com/tail-light-smoked-factory-style-left-driver-fits-2016-2019-toyota-tacoma.html).

Fixes: ([link ticket](https://app.asana.com/0/0/1204134908074103/f))

#changelog Resolve issues with multiple calls to whenDefined

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

- [x] I have added my ticket link and provided a changelog message.

### DIFF
--- a/view/frontend/web/js/utils/when-defined.js
+++ b/view/frontend/web/js/utils/when-defined.js
@@ -17,9 +17,9 @@
 define([], function () {
     /**
      * Map of watched objects to maps of their respective watched properties to configured callbacks
-     * @type {Map<Object, Map<string, Map<string, Function>>>}
+     * @type {Map<Object, Map<string, Map<string, function>>>}
      */
-    var whenDefinedCallbacks = new Map([]);
+    var whenDefinedCallbacks = new Map();
 
     /**
      * Executes provided callback when a property gets defined on provided object.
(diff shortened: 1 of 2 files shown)


view/frontend/web/js/utils/when-defined.js
```diff
@@ -29,45 +29,47 @@ define([], function () {
...
-        if (object.hasOwnProperty(property)) {
+    function whenDefined (object, property, callback, key) {
+        if (object.hasOwnProperty(property) && typeof object[property] !== 'undefined') {
             callback();
         } else {
             var overloadedPropertyName = '_' + property;
             if (!whenDefinedCallbacks.has(object)) {
                 whenDefinedCallbacks.set(object, new Map([]));
             }
             if (!whenDefinedCallbacks.get(object).has(property)) {
```
vitaliyreznikov, line 42: Could we do an early return here? It looks like it may add readability
if (whenDefinedCallbacks.get(object).has(property)) {
 whenDefinedCallbacks.get(object).get(property).set(key, callback);
 return;
}
whenDefinedCallbacks.get(object).set(property, []).set(key, callback);

view/frontend/web/js/utils/when-defined.js
```diff
@@ -29,45 +29,47 @@ define([], function () {
      * @param {Object} object to check for property definition
      * @param {number|string} property that is expected to be defined on {@see object}
      * @param {Function} callback function to be called when {@see property} gets defined on {@see object}
-     * @param {null} key deprecated parameter used for setting multiple callbacks per property
+     * @param {string} key used to prevent setting duplicate callbacks 
      */
-    function whenDefined(object, property, callback, key = null) {
```
vitaliyreznikov, line 34: We need a key for cases when we want to add two callbacks to the same property, right?
I don't know JS well, but wonder if it's possible to check if we call the function with the same callback or with a different one?

view/frontend/web/js/utils/when-defined.js
```diff
@@ -29,45 +29,47 @@ define([], function () {
...
+                        this[overloadedPropertyName] = value;
+                        for (var propertyCallback of whenDefinedCallbacks.get(object).get(property).values()) {
+                            propertyCallback();
+                        }
                     }
-                }
-            });
+                });
+            }
+            if (typeof key == 'undefined') {
```
vitaliyreznikov, line 69: Alternative option: if there are few calls with undefined we may add only the first callback

vitaliyreznikov:
@danac-gs Thank you for fixing this! Could you please add link to the asana and explain the issue a little bit?

view/frontend/web/js/utils/when-defined.js
danac-gs: @vitaliyreznikov done

view/frontend/web/js/utils/when-defined.js
danac-gs: This component is already used in the most complex parts of the code, changing the behavior like this can be unnecessarily risky

view/frontend/web/js/utils/when-defined.js
danac-gs: I don't have a definite answer here. There are probably ways of doing it. Maybe if we always first define the function and pass the callback as a named functions instead of declaring an anonymous one inline, as we usually do. Seems out of scope of this particular problem and also not too practical.

vitaliyreznikov (approved):

PR #1780, opened 2023-03-27
Request magento cart from backend if it's not present

vitaliyreznikov:
Our frontend integration relies on Magento cart section data persisting and updating in local storage.
This works fine, but if the env is broken for some third-party integration Magento cart may not be retrieved and the bolt button doesn't work.
See an example in asana.

In this PR we catch this situation and requested magento cart from backend manually.

Asana:
https://app.asana.com/0/1200879031426307/1204081354756081/f

#changelog Request magento cart from backend if it's not present

### DIFF
--- a/view/frontend/templates/js/replacejs.phtml
+++ b/view/frontend/templates/js/replacejs.phtml
@@ -1253,6 +1253,14 @@ $isAjaxAddToCartSuccessTimeoutDisabled = $block->isAjaxAddToCartSuccessTimeoutDi
                 // check if magento section was initialized before our code execution
                 if (customerData.get('cart')().data_id !== undefined) {
                     magentoCartDataListener(customerData.get('cart')());
+                } else {
+                    // if magento cart is not present for some reason (wrong setup)
+                    // we need to get it ourself
+                    setTimeout(function () {
+                        if (BoltState.magentoCart === null) {
+                            customerData.invalidate(['cart']);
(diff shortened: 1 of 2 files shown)


view/frontend/templates/js/replacejs.phtml
```diff
@@ -1253,6 +1253,14 @@ $isAjaxAddToCartSuccessTimeoutDisabled = $block->isAjaxAddToCartSuccessTimeoutDi
...
                 if (customerData.get('cart')().data_id !== undefined) {
                     magentoCartDataListener(customerData.get('cart')());
+                } else {
+                    // if magento cart is not present for some reason (wrong setup)
+                    // we need to get it ourself
+                    setTimeout(function () {
+                        if (BoltState.magentoCart === null) {
+                            customerData.invalidate(['cart']);
+                        }
+                    }, 4000);
```
mikeshutov-bolt, line 1263: Any particular reason we chose 4000? Also can we turn the value into a const since were using it in 2 places so its easy to modify in the future.

vitaliyreznikov: 4 seconds is sort of a rule of thumb: the time is enough to have cart data if it set up correctly and not too long because shoppers eventually will wait until it resolves.
I'll add the constant.

garmashdev (approved):

PR #1808, opened 2023-05-05
Added addItemToQuote func call to apply addons to immutable quote

KyerdnA:
# Description
When we add addon and then request again shipping methods addon is removed from the cart. In order to solve it, we applying addons to immutable quote as well

Fixes: [(link ticket)](https://app.asana.com/0/1201931884901947/1203586216479563)

#changelog Added addItemToQuote func call to apply addons to immutable quote

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:

### DIFF
--- a/Model/Api/UpdateCart.php
+++ b/Model/Api/UpdateCart.php
@@ -269,6 +269,9 @@ class UpdateCart extends UpdateCartCommon implements UpdateCartInterface
                         return false;
                     }
                     $removeFromImmutableQuote = true;
+                    if (!$immutableQuote->getItemById($quoteItem['quote_item_id'])) {
+                        $removeFromImmutableQuote = false;
+                    }
                     if ($parentQuote->getId() == $immutableQuoteId) {
                         $removeFromImmutableQuote = false;
                     }
--- a/Model/Api/UpdateCart.php
+++ b/Model/Api/UpdateCart.php
@@ -209,7 +209,7 @@ class UpdateCart extends UpdateCartCommon implements UpdateCartInterface
                     if ($immutableQuote->getItems()) {
                         $addToImmutableQuote = false;
                     }
-                    if ($addToImmutableQuote) {
+                    if ($addToImmutableQuote && !$parentQuote->getId() == $immutableQuoteId) {
                         $result = $this->addItemToQuote($product, $immutableQuote, $addItem, $quoteItem);
                         if (!$result) {
                             // Already sent a response with error, so just return.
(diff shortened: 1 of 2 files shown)


Model/Api/UpdateCart.php
```diff
@@ -203,6 +203,13 @@ public function execute($cart, $add_items = null, $remove_items = null, $discoun
                         // Already sent a response with error, so just return.
                         return false;
                     }
+                    // Adding the same data into immutable quote
+                    // in order to avoid missing addons in parent quote
+                    $result = $this->addItemToQuote($product, $immutableQuote, $addItem, $quoteItem);
```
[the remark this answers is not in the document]
vitaliyreznikov: I think what we need to check here: is if parent quote and immutable quote have the same quote ID.
And add items only if they are different.

vitaliyreznikov (commented):
Please see my last comment

Model/Api/UpdateCart.php
```diff
@@ -203,6 +203,19 @@ public function execute($cart, $add_items = null, $remove_items = null, $discoun
                         // Already sent a response with error, so just return.
                         return false;
                     }
+                    // Adding the same data into immutable quote
+                    // in order to avoid missing addons in parent quote
+                    $addToImmutableQuote = true;
+                    if ($immutableQuote->getItems()) {
+                        $addToImmutableQuote = false;
```
vitaliyreznikov, line 210: Hmm, why do we check this? quote always has items, right?

KyerdnA: Yeah, you're right
I've replaced the old correct validation with this one in a hurry, will rollback this change in a moment

Model/Api/UpdateCart.php
```diff
@@ -203,6 +203,21 @@ public function execute($cart, $add_items = null, $remove_items = null, $discoun
                         // Already sent a response with error, so just return.
                         return false;
                     }
+                    // Adding the same data into immutable quote
+                    // in order to avoid missing addons in parent quote
+                    $addToImmutableQuote = true;
```
vitaliyreznikov, line 208: I just realized - we need to remove from immutable quote as well.
See line 259 of the same file

Model/Api/UpdateCart.php
```diff
@@ -203,6 +203,28 @@ public function execute($cart, $add_items = null, $remove_items = null, $discoun
...
+                    $immutableQuoteItems = $immutableQuote->getItems();
+                    if ($immutableQuoteItems) {
+                        foreach ($immutableQuoteItems as $item) {
+                            if ($item->getSku() == $product->getSku()) {
+                                $addToImmutableQuote = false;
```
vitaliyreznikov, line 213: Do you have any specific case in mind where we need this check or you added it just in case?
Is it possible that via add to cart we change quantity i.e. from 1 to 2, not add new item to cart?

vitaliyreznikov: Let's remove this check then

vitaliyreznikov (approved):

PR #1810, opened 2023-05-11
[subscriptions] Refactor the subscription frequency selector code

smahbod2014:
# Description

This PR refactors the subscription frequency selector that we previously worked on for the plugin. Before, we copied the Product Page Checkout (PPC) code and made a new phtml file and wrote data to local storage. This refactor addresses the concerns brought up in [this doc](https://docs.google.com/document/d/1sWv2sMwTjFhJ1DM0xPbV1Dx2vcs5yVQYk5KJVs246kg/edit#), namely:

* We removed all the code in the `subscription_freq_selector_product_page.phtml` file except for the `div` that will be mounted (this can be further addressed if we want to remove the file entirely, but the div is positioned in a specific way at the moment, so I would rather do that separately)
* No more writing to local storage. Instead we leverage the `ajax:addToCart` event to call a function we've written in storm to persist subscription settings for the product.

#changelog Sean/m2 subs1

# Type of change

# How Has This Been Tested?

# For PR Reviewer 

# Checklist:


Block/Js.php
```diff
@@ -91,6 +91,12 @@ public function __construct(
         $this->httpContext = $httpContext;
     }
 
+    public function getCurrentProduct() {
```
vitaliyreznikov, line 94: Could we add a check if we are on product page inside this function as well?

Block/Js.php
```diff
@@ -91,6 +91,12 @@ public function __construct(
...
+        $objectManager = \Magento\Framework\App\ObjectManager::getInstance();
+        $product = $objectManager->get('Magento\Framework\Registry')->registry('current_product');
```
vitaliyreznikov, line 96: Using objectManager is not a proper magento way. @garmashdev could you please recommend how we can take the current project here?

view/frontend/web/js/bolt-api-driven-checkout.js
```diff
@@ -1014,6 +1017,25 @@ define([
             );
         },
 
+        boltSetUpSubscriptionFrequencySelector: function() {
+            // deprecated, to be removed
```
vitaliyreznikov, line 1021: Could you please add asana task with explanation here?

view/frontend/web/js/bolt-api-driven-checkout.js
```diff
@@ -1077,9 +1102,39 @@ define([
...
+            return new Promise(resolve => {
+                setTimeout(() => { resolve('') }, millisec);
+            })
+        }
+
+        // When an item is added to an empty cart, the quoteMaskedId takes a moment to get populated, so wait for that
+        const maxAttempts = 3;
+        for (let i = 0; i < maxAttempts; i++) {
+            if (!BoltCheckoutApiDriven.quoteMaskedId) {
+                await sleep(1000);
```
vitaliyreznikov, line 1127: Could we remove this sleep code?
I still think that moving wait logic to storm is the best approach, so we can have two API calls:
`window.BoltCheckout.persistSubscriptionFrequencySelection(quoteMaskedId) where quoteMaskedId may be null
`and
`window.BoltCheckout.setCartToken`

but as a simple middleground we can set local flag
```
if (BoltCheckoutApiDriven.quoteMaskedId) {
            window.BoltCheckout.persistSubscriptionFrequencySelection(BoltCheckoutApiDriven.quoteMaskedId);
} else {
        this.shouldCallPersistSubscriptionFrequencySelection = true
}
```
and after line
`BoltCheckoutApiDriven.quoteMaskedId = magentoCart.quoteMaskedId;`
add
```
if this.shouldCallPersistSubscriptionFrequencySelection {
 BoltCheckoutApiDriven.quoteMaskedId = magentoCart.quoteMaskedId;
 this.shouldCallPersistSubscriptionFrequencySelection = false
}
```

view/frontend/web/js/bolt-api-driven-checkout.js
smahbod2014: I'm going to make a couple more changes to storm code, then I can remove this piece here.

vitaliyreznikov: Good! Let's add link to asana

smahbod2014:
Thank you for the review. We've had a sudden change in priorities, so I'm going to have to pause work on this, but I promise I'll re-open this as soon as we're able to resume work.

